PR #6, opened 2018-02-15
Added custom dataset schema

### LINKED ISSUE #1: Implement custom dataset Schema
In CKAN, the dataset schema lists the metadata fields present in [a dataset and resources](https://www.opendatani.gov.uk/api/action/package_show?id=notifiable-infectious-diseases-report-2016-week-17), as well as the validation rules attached to them. CKAN includes a [default schema](https://github.com/ckan/ckan/blob/master/ckan/logic/schema.py#L103) but most instances will customize it to suit their own needs.

This was traditionally done with the `IDatasetForm` interface, and it involved a fairly [tedious process](http://docs.ckan.org/en/latest/extensions/adding-custom-fields.html). Luckily now there is [ckanext-scheming](https://github.com/ckan/ckanext-scheming) available. ckanext-scheming allows to define one or more custom schemas declaratively using a JSON file like [this one](https://github.com/okfn/ckanext-unhcr/blob/master/ckanext/unhcr/schemas/dataset.json). The extension will then take care of creating any internal hooks on `IDatasetForm` and rendering the form depending on the definitions.

We might need to end up using `IDatasetForm` anyway if we need more flexibility but to start with let's try to stick with ckanext-scheming.

I already added the [schema definition file](https://github.com/okfn/ckanext-unhcr/blob/master/ckanext/unhcr/schemas/dataset.json) to this repo, which includes one of the custom fields (Data Processing Status), now we need to add the rest defined in the [Dataset Field Spec sheet](https://docs.google.com/spreadsheets/d/1Gg1lNXqZG_CBltv6hcRX6UkTmOTZdJ4MtwxmXQjY0Fk/edit#gid=90335564) and [File Field Spec sheet](https://docs.google.com/spreadsheets/d/1Gg1lNXqZG_CBltv6hcRX6UkTmOTZdJ4MtwxmXQjY0Fk/edit#gid=228295987) on the spreadsheet.

Many fields in the spec need to be clarified with the client. I added a couple of columns with comments about the status of each. We will update them as we get more details, but the ones marked `ok` and `done` are safe to add.

Tasks:

- [x] Edit the `dataset.json` file to add any existing core field and new fields needed
  * If custom validators are needed, register them with `IValidators` (example [here](https://github.com/frictionlessdata/ckanext-validation/blob/master/ckanext/validation/plugin.py#L231) and [here](https://github.com/frictionlessdata/ckanext-validation/blob/master/ckanext/validation/validators.py#L11))
  * If custom field types or template snippets are needed, create custom presets like [here](https://github.com/frictionlessdata/ckanext-validation/blob/master/ckanext/validation/examples/ckan_default_schema.json#L110) and [here](https://github.com/frictionlessdata/ckanext-validation/blob/master/ckanext/validation/presets.json)

Dataset:
- [x] Date range data collected -> #11
- [x] Date file created -> internal
- [x] Date modified -> internal
- [x] Linked datasets -> #7 
- [x] Keywords
- [x] Unit of measurement
- [x] Description
- [x] Operational purpose of data
- [x] Uploader -> internal
- [x] HXL-ated?
- [x] Raw Data Access Level
- [x] Dataset process status -> #8 
- [x] Project URL
- [x] Archive status
- [x] Identifiability -> #9 
- [x] Admin Notes -> #10 
- [x] File Attachments
- [x] Original ID
- [x] Data collection technique
- [x] License dataset

Resource:

- [x] Date range data collected -> #11 
- [x] Version
- [x] Date modified -> internal
- [x] File Type
- [x] Description
- [x] Created by -> internal
- [x] File format
- [x] HXL-ated?
- [x] File process status
- [x] Identifiability -> #9 
- [x] File Size -> internal

roll:
- fixes #1

### DIFF
--- a/ckanext/unhcr/schemas/dataset.json
+++ b/ckanext/unhcr/schemas/dataset.json
@@ -298,6 +298,22 @@
             "preset": "resource_url_upload"
         },
         {
+            "field_name": "data_sensitivity",
+            "help_text": "Apply to both Anonymized and  Personally identifiable data",
+            "label": "Data sensitivity",
+            "preset": "select",
+            "choices": [
+                {
+                    "label": "Yes",
+                    "value": "yes"
+                },
+                {
+                    "label": "No",
+                    "value": "no"
+                }
+            ]
+        },
+        {
             "field_name": "original_id",
             "help_text": "If the dataset already has an ID from the source org, DDI, etc...",
             "label": "Original ID"
--- a/ckanext/unhcr/schemas/dataset.json
+++ b/ckanext/unhcr/schemas/dataset.json
@@ -52,6 +52,19 @@
             "label": "Source organization(s)"
         },
         {
+            "field_name": "date_range_data_collected",
+            "help_text": "Date range on which the data was gathered",
+            "label": "Date range data collected"
+        },
+            "field_name": "metadata_created",
+            "help_text": "System generated field for date when this record was created in the RIDL database",
+            "label": "Date file created"
+        },
+            "field_name": "metadata_modified",
+            "help_text": "System generated field for date on which the RIDL record was last modified",
+            "label": "Date modified"
+        },
+        {
             "field_name": "process_status",
             "help_text": "Indicates the processing stage of the data. 'Raw' means that the data has not been cleaned since collection. 'In process' means that it is being cleaned. 'Final' means that the dataset is final and ready for use in analytical products",
             "label": "Dataset Process Status",
--- a/ckanext/unhcr/schemas/dataset.json
+++ b/ckanext/unhcr/schemas/dataset.json
@@ -67,7 +67,6 @@
             ]
         }
     ],
-    "dataset_type": "dataset",
     "resource_fields": [
         {
             "field_name": "url",
--- a/ckanext/unhcr/schemas/dataset.json
+++ b/ckanext/unhcr/schemas/dataset.json
@@ -23,12 +23,6 @@
             "preset": "tag_string_autocomplete"
         },
         {
-            "field_name": "license_id",
-            "form_snippet": "license.html",
-            "help_text": "License definitions and additional information can be found at http://opendefinition.org/",
-            "label": "License"
-        },
-        {
             "field_name": "owner_org",
             "label": "Organization",
             "preset": "dataset_organization"
@@ -415,6 +409,12 @@
                     "value": "other"
                 }
             ]
+        },
+        {
+            "field_name": "license_id",
+            "form_snippet": "license.html",
+            "help_text": "License definitions and additional information can be found at http://opendefinition.org/",
+            "label": "License"
         }
     ],
     "resource_fields": [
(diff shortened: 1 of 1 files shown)


ckanext/unhcr/schemas/dataset.json
```diff
@@ -46,11 +40,223 @@
...
             "label": "Source"
         },
+        {
+            "field_name": "source_organization",
+            "help_text": "Which organization owns / collected the data",
+            "label": "Source organization(s)",
+            "required": true
+        },
+        {
+            "field_name": "date_range_data_collected",
```
amercader: `date_range_data_collected`,  `metadata_modified`, `metadata_created` and `author` don't need to be present as these are handled by CKAN internally

ckanext/unhcr/schemas/dataset.json
```diff
@@ -65,9 +271,152 @@
...
+                    "value": "archived"
+                },
+                {
+                    "label": "Deleted",
+                    "value": "deleted"
+                }
+            ]
+        },
+        {
+            "field_name": "identifiability",
```
amercader: The `identifiability` at the dataset level will be automatically generated at the dataset level based on the values of all resources, so it should not be displayed in the dataset form, only in the resource form. Can we hide it from the form with `form_display: null`?

amercader: Use this [preset](https://github.com/frictionlessdata/ckanext-validation/blob/master/ckanext/validation/examples/ckan_default_schema.json#L120

ckanext/unhcr/schemas/dataset.json
```diff
@@ -46,11 +40,223 @@
             "form_placeholder": "http://example.com/dataset.json",
             "label": "Source"
         },
+        {
+            "field_name": "source_organization",
```
amercader: This will be a list, waiting for values

ckanext/unhcr/schemas/dataset.json
amercader: See #11

ckanext/unhcr/schemas/dataset.json
```diff
@@ -46,11 +40,223 @@
...
+            "form_snippet": null
+        },
+        {
+            "field_name": "metadata_modified",
+            "help_text": "System generated field for date on which the RIDL record was last modified",
+            "label": "Date modified",
+            "form_snippet": null
+        },
+        {
+            "field_name": "linked_datasets",
```
amercader: See #7

roll:
@amercader 
Please review. I've added list here - https://github.com/okfn/ckanext-unhcr/issues/1

I've tested dataset/resources creation - it works.

ckanext/unhcr/schemas/dataset.json
roll: It says preset `hidden_in_form` is not registered. I used `form_snippet: null`.

roll:
I will merge it now to start PRs for extracted issues. I think it should be OK because you have already reviewed it and we could update it anytime anyway.

PR #14, opened 2018-02-19
[WIP] Added dataset_process_status validator

### LINKED ISSUE #8: Dataset Process Status / File Process Status fields
Dataset Process Status is a system generated field, so it should not appear on the dataset form ( use this [preset](https://github.com/frictionlessdata/ckanext-validation/blob/master/ckanext/validation/examples/ckan_default_schema.json#L120))

Implementation

We need to write a validator that reads the values of all `file_process_status` fields and sets the `dataset_process_status` to the latest one (Raw < In progress < Final)

Validators can have two signatures:

```python

def validator_1(value, context):
   pass

def validator_2(key, data, errors, context):
   pass

```

We need the second one, as we need access to the rest of the fields submitted in the form

```python
def dataset_process_status(key, data, errors, context):

    for field in data:
        # check if it's file_process_status, if so store the value
       
   # compute the highest value (Raw < In progress < Final)

   # Set `dataset_process_status`
   data[key] = highest_resource_value

```

roll:
- fixes #8

### DIFF
--- a/ckanext/unhcr/plugin.py
+++ b/ckanext/unhcr/plugin.py
@@ -3,7 +3,7 @@ import ckan.plugins as plugins
 import ckan.plugins.toolkit as toolkit
 from ckan.lib.plugins import DefaultTranslation
 
-from ckanext.unhcr import helpers
+from ckanext.unhcr import helpers, validators
 
 
 class UnhcrPlugin(plugins.SingletonPlugin, DefaultTranslation):
@@ -12,6 +12,7 @@ class UnhcrPlugin(plugins.SingletonPlugin, DefaultTranslation):
     plugins.implements(plugins.IFacets)
     plugins.implements(plugins.ITemplateHelpers)
     plugins.implements(plugins.IPackageController, inherit=True)
+    plugins.implements(plugins.IValidators)
 
     # IConfigurer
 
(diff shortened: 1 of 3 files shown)


roll:
@amercader 
I have no luck here...

Actual (this error seems related):

```
ckan-dev_1    | ProgrammingError: (ProgrammingError) can't adapt type 'Missing' 'INSERT INTO package_extra (id, package_id, key, value, state, revision_id) VALUES (%(id)s, %(package_id)s, %(key)s, %(value)s, %(state)s, %(revision_id)s)' {'package_id': 'e1c39253-9cbc-436d-a170-78b2fdc09de5', 'value': <ckan.lib.navl.dictization_functions.Missing object at 0x7f35c6e33490>, 'state': 'active', 'key': u'dataset_process_status', 'revision_id': u'bc655e99-d003-42ae-bf59-d90486d34e53', 'id': u'52d2f80d-e4c8-40ba-bc9c-8e2b70364312'}
```

Expected:
an empty validator should be just no-op

Probably I'm missing something regarding validators but can't figure out from readmes and examples/

roll:
If I remove `validators` from `dataset.json:dataset_fields.` for `dataset_process_status` the extension works again.

roll:
CLOSED in favor of #17

PR #17, opened 2018-02-20
Compute dataset fields depending on file fields asynchronously

### LINKED ISSUE #8: Dataset Process Status / File Process Status fields

### LINKED ISSUE #9: Identifiability field
Dataset Identifiability is a system generated field, so it should not appear on the dataset form ( use this [preset](https://github.com/frictionlessdata/ckanext-validation/blob/master/ckanext/validation/examples/ckan_default_schema.json#L120))

Implementation

We need to write a validator that reads the values of all `identifiability` resource fields and sets the `identifiability` dataset to to the lowest ( Personally identifiable < Anonymized). Additionally, if the result is PI, we should ensure the `private` field is True.

Validators can have two signatures:

```python

def validator_1(value, context):
   pass

def validator_2(key, data, errors, context):
   pass

```

We need the second one, as we need access to the rest of the fields submitted in the form

```python
def identifiability_validator(key, data, errors, context):

    for field in data:
        # check if it's a resource identifiability field, if so store the value
       
   # compute the lowest value (Personally identifiable < Anonymized)

  # If PI, check that `private` in data is True, if not raise Invalid

   # Set `dataset_process_status`
   data[key] = lowest_resource_value

```

### LINKED ISSUE #11: Dataset Date Range Field
Dataset Date Range Field is a system generated field, so it should not appear on the dataset form ( use this [preset](https://github.com/frictionlessdata/ckanext-validation/blob/master/ckanext/validation/examples/ckan_default_schema.json#L120))

Implementation

We need to write a validator that reads the values of two resource fields (`date_range_start` and `date_range_end` and sets the same fields in the dataset using the lowest start and the highest end.

Validators can have two signatures:

```python

def validator_1(value, context):
   pass

def validator_2(key, data, errors, context):
   pass

```

We need the second one, check other issues for examples on how to use it.

### LINKED ISSUE #15: Compute dataset fields depending on file fields asynchronously
The approach described in #8, #9 and #11 won't work when creating the dataset via the UI (99.9% of the time), because we are creating the dataset in two stages, first the metadata and then the resource(s). When the resources are created we can not change dataset fields inside the validators.

Let's do something much more simpler, a [background job](http://docs.ckan.org/en/latest/maintaining/background-tasks.html) fired after dataset creation or update (creating or editing a resource calls `package_update`) that gets the dataset id, retrieves it and updates it with all the fields that depend on file values. There will be a delay to have these values filled but in most cases it should happen almost immediately.

Implementation:

* Write a job function that gets a dataset id:
  * Calls `package_show` to get the full details
  * Iterates over all resources (if any) and applies the different logic described in #8, #9 and #11 to calculate the values for `dataset_process_status`, `identifiability` and `date_range_start/end`.
  * Calls `package_update` to update the dataset
* Implement `IPackageController` [`after_create`](http://docs.ckan.org/en/latest/extensions/plugin-interfaces.html#ckan.plugins.interfaces.IPackageController.after_create) and `after_update` methods, and enqueue the above function passing the dataset id as parameter, eg

    ```python

    # IPackageController
         toolkit.enqueue_job(process_dataset_fields, [data_dict['id']])
  
```
* Hide these three fields from the dataset form, using [this preset](https://github.com/frictionlessdata/ckanext-validation/blob/master/ckanext/validation/examples/ckan_default_schema.json#L120). Right now the preset is defined in ckanext-validation, you can copy it to unhcr if you want.

### PR DESCRIPTION
**Pull request by roll**
- fixes #8 
- fixes #9 
- fixes #11 
- fixes #15

### CODE REVIEW

**Comment by roll** [comment]
@amercader 
WDYT of activating Travis? (#16)

I have some troubles running tests locally.

File: ckanext/unhcr/jobs.py [hunk 1]
```diff
@@ -0,0 +1,61 @@
...
+    DEFAULT = 'anonymized'
+    WEIGHTS = {'personally_identifiable' : 2, 'anonymized': 1}
+    for resource in package['resources']:
+        package_weight = WEIGHTS.get(package[KEY], 0)
+        resource_weight = WEIGHTS.get(resource[KEY], 0)
+        if resource_weight > package_weight:
+            package[KEY] = resource[KEY]
+    package[KEY] = package.get(KEY) or DEFAULT
+
+    # raw_access_data_level
```
roll: > If "Personally identifiable" is selected, the access level should default to "private."

Raw access level or there is also some kind of ckan access level to set as private?

ckanext/unhcr/jobs.py
```diff
@@ -0,0 +1,61 @@
...
+            package[KEY] = resource[KEY]
+    package[KEY] = package.get(KEY) or DEFAULT
+    return package
+
+
+def _modify_identifiability_and_raw_access_data_level(package):
+
+    # identifiability
+    KEY = 'identifiability'
+    DEFAULT = 'anonymized'
```
roll: Default value if there is no resources (not clear from requirements)

amercader: We will just use the `private` field in the dataset for this, which has a default of `True`, so use `Personally Identifiable` as the default value.

ckanext/unhcr/jobs.py
amercader: There is a boolean `private`field for this, if `raw_access_data_level` is `Private`, set it to `True`.

ckanext/unhcr/jobs.py
```diff
@@ -0,0 +1,75 @@
...
+    # Get package
+    package_show = toolkit.get_action('package_show')
+    package = package_show({}, {'id': package_id})
+
+    # Modify package
+    package = _modify_package(package)
+
+    # Update package
+    package_update = toolkit.get_action('package_update')
+    package_update({}, package)
```
roll: @amercader 
I expect this line to change the package but it does nothing

amercader: Try passing `{'ignore_auth': True}` as context

roll:
@amercader 
So because it works actually (:confused: :smiley:) - may be a quick look at the business rules aka review and we're good to go?

amercader:
I'll set up travis

amercader:
@roll I've setup Travis, can you push a little commit to trigger the tests?

amercader:
I've they pass this is good to go

roll:
@amercader 
I need to rebase when I'll be around my laptop

PR #23, opened 2018-02-22
Schema improvements

### LINKED ISSUE #22: Changes in dataset schema
- [x] Use the existing `url` field ("Source" field in the form) for "Project URL", don't create a new one. Just change the labels and help text to reflect the spec.
- [x] Having "Raw Data Access Level" and "Visibility" is redundant and confusing. Let's just use "Visibility" (which is the `private` field in the dataset) and rename the labels. So:
    * Remove the `raw_data_access_level` field
    * Rename the values in the "Visibility" field. To do so we need to extend [this template](https://github.com/ckan/ckanext-scheming/blob/master/ckanext/scheming/templates/scheming/form_snippets/organization.html) (probably just the `package_metadata_fields_visibility` block). To extend it create the file `ckanext/unhcr/templates/scheming/form_snippets/organization.html`. Not sure if this will be enough or you will need to copy the whole source:

    ```
    {% ckan_extends %}
    {% block package_metadata_fields_visibility %}
       # Change the values here

    {% endblock %}
    ```
    Values should change like this:
       * Visibility -> Raw Data Access Level
       * Public -> Internally Visible
   Also remember to change the background job to reflect this.

- [x] Archive field. Change to a boolean one, with label "Archived"
- [x] Linked datasets: remove for now
- [x]  File attachments: remove for now

roll:
- fixes #22


ckanext/unhcr/schemas/dataset.json
```diff
@@ -236,28 +230,19 @@
...
             "choices": [
                 {
-                    "label": "Active",
-                    "value": "active"
-                },
-                {
-                    "label": "Archived",
-                    "value": "archived"
+                    "label": "No",
+                    "value": "False"
```
roll: A weird question but I'm not getting how to create a boolean value using `chanext-sheeming`. I've tried to use true/false as values it it leads to a validation error.

roll:
@amercader 
Please take a look. Everything should be good except the boolean field question.

ckanext/unhcr/schemas/dataset.json
amercader: Let's leave it like that. There probably isn't a boolean preset yet.

PR #28, opened 2018-02-27
Updated data container schema

### LINKED ISSUE #25: Data container schema
Similar to the dataset one but using `schemas/data_container.json`.
Follow the spec in the spredsheet ("Data Container Fields Spec" sheet). There's still lots to clarify but we can start adding fields.

It requires the latest changes in the compose setup, and to rebuild the ckan image.

roll:
- fixes #25


roll:
Thanks

PR #29, opened 2018-02-28
Request data container

### LINKED ISSUE #19: Request data container creation
Administrators of an existing data container can request the creation of a children data container.

## Process overview

* Data Container Admins see a "Request Data Container" button in the Data Containers page
* This opens the same form as creating a new data container, but in the "Parent" field there are only available the data containers the user is an admin of. 
* When submitted, the data container is created, and the submitter is added as admin, but internally the data container is flagged as `approval_needed`, so no it's not visible in search results and datasets can not be assigned to them.
* Site administrators receive an email with a link to activate or delete the new data container
* In either case, the user that requested the data container receives an email with the outcome (approved or rejected)

## Implementation

- [x] Set `ckan.auth.user_create_organizations` to True
- [x] Override the core `organization_create` auth function, allowing it only to users that are admin of an organization (`IAuthFunctions`)
- [x] Override the core `organization_create` and `organization_update` actions via `IActions` with similar logic to [this plugin](https://github.com/ViderumGlobal/ckanext-iati/blob/master/ckanext/iati/logic/action.py#L60):
  - [x] Check if user is sysadmin, if so defer to the core action and create org normally
  - [x] Check if has access (`check_access`), if so override state to `approval_needed` before calling the core action.
  - [x] Notify sysadmins if necessary via email
  - [x] When the state is changed, notify original user via email
- [x] UI: Display a different button on data containers index template depending on whether a sysadmin or an org admin ("Request data container" if the latter)
- [x] UI: Display a notice in the org form if not a sysadmin ("This data container needs to be approved, etc"
- [x] Form: only display the organizations that the user has permission in the parent form (this might happen automatically)
- [x] Check that pending containers don't appear in searches, and that datasets can not be assigned to them

## Estimate

3-4 days

roll:
- fix #19 

----

- [x] finish/move tasks from #19 
- [x] add proper org url for emails
- [x] provide a way to approve pending data containers for administrators
- [x] fix `Approve/Request Data Container` button
- [x] check that pending containers don't appear in searches, and that datasets cannot be assigned to them
- [x] `ckanext-hierarchy` customization - https://github.com/okfn/ckanext-hierarchy/issues/1
- [x] improve Approve/Reject UI
- [x] send actual emails on the create/update org

---

- [x] write tests (see https://github.com/frictionlessdata/ckanext-validation/blob/master/ckanext/validation/tests/test_logic.py#L257)
- [x] revert the required flag removal

---

- [x] merge the `ckanext-scheming` quick fix (@amercader to do) - MOVED #34
- [x] merge the `ckanext-hierarchy` customization (@amercader to do)


ckanext/unhcr/schemas/data_container.json
```diff
@@ -24,16 +24,15 @@
...
-            "form_snippet": "org_hierarchy.html",
-            "validators": "ignore_missing"
+            "validators": "ignore_missing",
+            "form_snippet": "org_hierarchy.html"
         },
         {
             "field_name": "country",
             "label": "Country",
             "preset": "multiple_select",
-            "required": true,
```
roll: revert

ckanext/unhcr/schemas/data_container.json
```diff
@@ -46,7 +45,6 @@
             "label": "Geographic area referenced",
             "help_text": "The geographic areas covered by the datasets included within",
             "preset": "multiple_select",
-            "required": true,
```
roll: revert

ckanext/unhcr/schemas/data_container.json
amercader: why?

roll: Return `required: true` - just saving time on debugging while it's a WIP

amercader: cool

amercader:
Looking really good so far @roll. If you need inspiration for the auth tests, check these ones from [ckanext-validation](https://github.com/frictionlessdata/ckanext-validation/blob/master/ckanext/validation/tests/test_logic.py#L257) or let me know if you want me to write a few sample ones.

roll:
@amercader 
Something like this so far... I've added an extended task list to the first message.

ckanext/unhcr/actions.py
```diff
@@ -0,0 +1,92 @@
...
+        # Notify users
+        _send_activation_notification_email(context, new_org_dict)
+        h.flash_success('Data container activated, a notification email has been sent to its administrators.')
+
+    return new_org_dict
+
+
+def _send_new_organization_email(context, org_dict):
+
+    # TODO: get proper site url
```
amercader:
```python
toolkit.url_for('data-container_read', id=org_dict['name'], qualified=True)
```

ckanext/unhcr/actions.py
```diff
@@ -0,0 +1,90 @@
...
+    # created as pending, and sysadmins notified
+
+    notify_sysadmins = False
+    try:
+        toolkit.check_access('sysadmin', context, data_dict)
+    except toolkit.NotAuthorized:
+        # Not a sysadmin, create as pending and notify sysadmins (if all went
+        # well)
+        context['__unhcr_state_pending'] = True
+        data_dict['approval_status'] = 'approval_needed'
```
amercader: Use the existing `state` field, which is taken into account in searches etc, not a custom one. Sorry I should have wrote that in the spec.

amercader:
@roll For the actual approval process we need to set `state` (see [comment](https://github.com/okfn/ckanext-unhcr/pull/29/files#r172166826)) to `active` if approved or `deleted` otherwise.

CKAN is quite picky when updating individual fields. Just sending a request to `organization_update` with the state field will delete all existing extra fields. We have to either send the whole dict to `organization_update` or send the individual fields to [`organization_patch`](http://docs.ckan.org/en/latest/api/index.html#ckan.logic.action.patch.organization_patch).

What I think is best is to create two dedicated endpoints, accessible only to sysadmins:

* `/data-container/{id}/accept`: 
* `/data-container/{id}/reject`:

On each of them we:

   * Check auth (user sysadmin)
   * Check if data container exists
   * Check if data container has pending state
   * Call `organization_patch` with `{"state": "active"}` or `organization_purge` (`organization_delete` will just marked as deleted and people won't be able to reuse the name)
   * Send OK / Nope email to submitter
   * Add a Flash message: "Data Container approved / rejected" (with `h.flash_success()  / h.flash_error()`)
   * Redirect to data-container read / data-container index page

To actually call these endpoints we can add a banner in the edit page like you did in https://github.com/okfn/ckanext-unhcr/pull/29/commits/ed0201a53c76e4a361b6ae3ba6b0eed69bdd8b91.

I've started a stub registering the endpoints in 100644

ckanext/unhcr/actions.py
roll: Thanks. I'm trying to. But I'm curious than what's role of the `group.approval_status` - https://github.com/ckan/ckan/blob/master/ckan/logic/action/create.py#L812

amercader: I'm curious too :) I'm not sure either so let's try `state` first

amercader:
To send emails:

* We will use [`ckan.lib.mailer.mail_user()`](https://github.com/ckan/ckan/blob/master/ckan/lib/mailer.py#L114) to send the actual email. It just needs the recipient address, subject and body (and headers if needed).
* We can wrap it in another function that receives the organization data_dict and (use [this](https://github.com/ckan/ckan/blob/master/ckan/lib/mailer.py#L132) as example):
   * Gets the user email.:
     * To get the submitter email get the email from `organization['users'][{"capacity": "admin", "email": }]`
     * To get all sysadmins emails addresses use [this](https://github.com/ViderumGlobal/ckanext-iati/blob/master/ckanext/iati/logic/action.py#L230). (I don't think there's an API to just get sysadmin users)
   * Builds the subject and body. The body will be a jinja2 template as in the example/
* We call this function from the accept / reject endpoints

roll:
@amercader 
please review

roll:
@amercader 
I've made the changes

amercader:
Great stuff @roll!

roll:
@amercader 
Thanks. Merged

PR #55, opened 2018-04-27
File attachements

### LINKED ISSUE #21: File attachments in datasets
All work done so far in the `file-attachements` branch.

Requirements:
Basically we need a way of uploading files that are related to the data but not data files themselves, eg PDF , Word files, images with reports, viz etc. In practice this means that datasets need two kinds of resources, data files that have the custom schema implemented now (with fields like `process_status`, `hxl-ated` etc) and attachments, with the custom resource schema that CKAN uses

Implementation:
CKAN does not support different resource types (ie different resource schemas on the same dataset). The way I've implemented this in the backend is:

* Custom field `type` with two options: `data` and `attachment`
* New custom validator (`ignore_if_attachment`) that should be added at the beginning of all validators in all custom fields in the resource schema (so all fields except `type`, `format`, `name`, `description` and `url`)

What's left is the UI. What I would like to do is change the resource form:

screenshot-2018-4-20 edit - unnamed resource - test dataset europe - ridl

When creating a new resource Instead of showing the full resource form we first show a form with two buttons / options (we can polish the actual copy later, and @smth can help use with the design):
"What kind of file do you want to upload?"
"Data files: files containing the actual data. Common formats include CSV or Excel"
"File attachments: companion files related to the data. For instance PDFs or Microsoft Word reports, Images, etc"

When clicking option 1, the `type` field is set to `data` and all form fields are displayed.
When clicking option 2, the `type` field is set to `attachment` and only the default form fields are displayed.
All this implemented in the same page with some JS magic. There might be some issues with the buttons below (Hide "Save & add another" and "Finish" until the user has not choose a type)

On the Dataset display page we might want to separate somehow the data files and the attachments in two separate lists, but that can come later:

screenshot-2018-4-20 test dataset europe - ridl

Tasks:

- [x] Review that all custom fields have the `ignore_if_attachment` validator added
- [x] Find out what template to override to implement the custom UI. This might be tricky, as ckanext-scheming might be interfering with the resource form templates
- [x] Create an intermediate form where users choose the type of upload. Show different form fields depending on the option. Set the `type` field.
- [x] Hide / Show the buttons below as appropiate

roll:
- fixes #21

### DIFF
--- a/ckanext/unhcr/fanstatic/theme.js
+++ b/ckanext/unhcr/fanstatic/theme.js
@@ -60,7 +60,7 @@ this.ckan.module('resource-type', function ($) {
 
       // Add event listeners
       $('button.btn-data', this.field).click(this._onDataButtonClick.bind(this))
-      $('button.btn-attachement', this.field).click(this._onAttachementButtonClick.bind(this))
+      $('button.btn-attachment', this.field).click(this._onAttachmentButtonClick.bind(this))
       $('button:contains("Previous")').click(this._onPreviousButtonClick.bind(this))
 
       // Emit initialized
@@ -73,8 +73,8 @@ this.ckan.module('resource-type', function ($) {
     _onIntialized: function () {
       if (this.options.value === 'data') {
         this._onDataButtonClick()
-      } else if (this.options.value === 'attachement') {
-        this._onAttachementButtonClick()
+      } else if (this.options.value === 'attachment') {
+        this._onAttachmentButtonClick()
       } else {
         this.field.nextAll().hide()
         this.field.show()
@@ -88,12 +88,12 @@ this.ckan.module('resource-type', function ($) {
       this.input.val('data')
     },
 
-    _onAttachementButtonClick: function (ev) {
+    _onAttachmentButtonClick: function (ev) {
       if (ev) ev.preventDefault()
       this.field.hide()
       this.field.nextAll().show()
       $('#field-format').parents('.control-group').nextAll('.control-group').hide()
-      this.input.val('attachement')
+      this.input.val('attachment')
     },
 
     _onPreviousButtonClick: function (ev) {
--- a/ckanext/unhcr/fanstatic/theme.js
+++ b/ckanext/unhcr/fanstatic/theme.js
@@ -59,8 +59,9 @@ this.ckan.module('resource-type', function ($) {
       this.input = $('input', this.field)
 
       // Add event listeners
-      $('.btn-data', this.field).click(this._onDataButtonClick.bind(this))
-      $('.btn-attachement', this.field).click(this._onAttachementButtonClick.bind(this))
+      $('button.btn-data', this.field).click(this._onDataButtonClick.bind(this))
+      $('button.btn-attachement', this.field).click(this._onAttachementButtonClick.bind(this))
+      $('button:contains("Previous")').click(this._onPreviousButtonClick.bind(this))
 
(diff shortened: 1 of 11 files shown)


roll:
- [ ] prev button behaviour

amercader:
@roll for the different resource lists we need to override this block: https://github.com/ckan/ckan/blob/master/ckan/templates-bs2/package/read.html#L36

Just call the resources_list.html snippet two times, one with the data resources and one with the attachements (you might need a helper to get both lists, or maybe you can do it in jinja directly).

The only issue is that that snippet has the title hardcoded, so we might need to override it completely on our extension to be able to pass the title as a parameter:

ckanext/unhcr/fanstatic/theme.js
```diff
@@ -42,3 +42,58 @@ $( document ).ready(function() {
   });
 
 });
+
+this.ckan.module('resource-type', function ($) {
```
amercader: This can be deleted can't it? It's on its own file AFAICT

ckanext/unhcr/validators.py
```diff
@@ -20,6 +28,16 @@ def linked_datasets(value, context):
 
 # Internal
 
+def _is_attachement(index, data):
+    for field, value in data.iteritems():
+        if (field[0] == 'resources' and
+                field[1] == index and
+                field[2] == 'type' and
+                value == 'attachement'):
```
amercader: This is my fault as I misspelled it first, but can you replace all occurrences of `attachement` with `attachment`?

ckanext/unhcr/templates/scheming/form_snippets/resource_type.html
```diff
@@ -0,0 +1,14 @@
+<div id="field-resource-type" data-module="resource-type" data-module-value="{{ data[field.field_name] }}" style="display: none">
+  <h3>What kind of file do you want to upload?</h3>
+  <div>
+    <div class="col-md-6">
+      <button class="btn btn-info btn-data">Data file</button>
+      <p><small>Files containing the actual data. Common formats include CSV or Excel</small></p>
+    </div>
+    <div class="col-md-6">
+      <button class="btn btn-info btn-attachement">File attachment</button>
+      <p><small>File attachments: companion files related to the data. For instance PDFs or Microsoft Word reports, Images, etc</small></p>
```
amercader: Remove the first `File attachments: ` bit.

amercader:
@roll The upload button stopped working for me. Maybe some js error or file missing?

roll:
@amercader 
I'll check

ckanext/unhcr/templates/package/read.html
```diff
@@ -0,0 +1,10 @@
+{% ckan_extends %}
+  {% block package_resources %}
+    {% snippet "package/snippets/resources_list.html", pkg=pkg, resources=(pkg.resources|selectattr("type", "equalto", "data")) %}
+    {% snippet "package/snippets/resources_list.html", pkg=pkg, resources=(pkg.resources|selectattr("type", "equalto", "attachment")) %}
+    <script>
+      // It allows us to overcome the snippet limitation with the heading contents
```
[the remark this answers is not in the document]
amercader: I think the snippet is worth it, for instance to not show one of the sections of there are no resources of that type

roll:
- [ ] move the snippet to the plugin

roll:
@amercader 
I've added the snippet

PR #59, opened 2018-05-03
[WIP] Linked dataset back reference tests

roll:


roll:
@amercader 
I've added tests to check the business logic but unfortunately, I have just failed to make it work with ckan. 

At some point, it seems trying to run `linked_datasets` validator though it should be skipped while we have `job == True` in `context`.

Also, I have other strange problems with other tests locally having things like `IntegrityError: (IntegrityError) duplicate key value violates unique constraint "user_name_key"` and inability to cherry-pick tests (it says it can't find the context in the thread)  which makes it's very hard and slow to debug. 

I think I'm out of week hours already so can you please take a look? If it's too hard to make it unit-testable I would suggest just to test it via UI. I've made my own test using different edge-case and it works fine.

amercader:
Cool @roll, I'll have a look, don't worry

PR #76, opened 2019-02-15
Deposited dataset and deposited data container

### LINKED ISSUE #74: Create a separate schema for curated datasets
As curated datasets have a more flexible schema than the default one, it makes sense to handle those as a separate dataset type (`curated-dataset`), and create a proper one when the curation process is finished.

Using a custom type will also hide all curated datasets from the default search listings (but not on `/curated-dataset`, more on that later).

Curated datasets have the same fields as the default schema, but only a minimal subset is required (let's start with title and description for now). Additionally they have an extra field for the data container they should end up in if the curation process finishes (eg `owner_org_dest` or something similar). `owner_org` will always be the id of the Curation Data Container. This will be enforced by a custom validator.

User creates a curated dataset (minimal schema, proposed data container) -> Assigned to the Curation Data Container -> [ Curation process TBD ] -> A new dataset is created on the requested data container

One thing that might be useful is to create a new endpoint that validates the dataset metadata against the default schema but without creating a dataset: you upload the data_dict, validate it against the default schema and return any ValidationErrors found. This will be useful for data curators to know when their data is ready.

## Tasks
- [x]  Create a new `curated-dataset` type in the schemas folder. Same fields as the default but only title and description mandatory.
- [x] Add a new field that holds the suggested data container (we can validate if it's an existing org, there's probably an existing validator in core for that)
- [x] Add a validator on `owner_org` that forces the value to be the Curation Data Container one
- [x] Implement mechanism for creating a regular dataset from a curated one (that meets all the metadata requirements). Whatever works best of:
   1. Use the same dataset, just change the owner_org. This is cleaner but might give permission problems
   2. Create a new dataset in the destination data container (by an internal account) using the same metadata as the curated one, and delete the curated one afterwards.
- [x] Hide curated datasets from search to users that are not members of the curation container (this could be as easy as to create all datasets private, or we might need to use `IPermissionLabels`)

### LINKED ISSUE #75: Implement deposit data container permissions
The deposit data container is implemented as a normal data container, ie is just a regular organization under the hood, but it does not come up in the data container index. For now it can be manually created, and the curation team (ie this organization admins and editors) can be manually added.

All users should be able to add datasets to the curation data container, but only edit the ones they submitted. We will need to extend the `package_create` and `package_update` auth functions to accommodate this.

The editors and admins in this container should be able to edit all datasets in it (default behaviour). The rest of users should not be able to see what datasets belong to this data container.

## Tasks

- [x] Create Curation Data Container and add some editors and admins
- [x] Modify the authorization logic to implement the spec above
- [x] Restrict access to this data container to non-members (`organization_show` auth function) - MOVED #78

roll:
- fixes #74 
- fixes #75
- requires https://github.com/okfn/docker-ckan-unhcr/pull/8


roll:
@amercader 
Please take a look. It's a POC missing:
- hiding deposited datasets from search
- ability to approve only valid dataset (I'll make a try to show validation errors for curators)

ckanext/unhcr/controllers/deposited_dataset.py
```diff
@@ -0,0 +1,64 @@
...
+
+        # Change to regural dataset
+        pkg_dict['type'] = 'dataset'
+        pkg_dict['owner_org'] = pkg_dict['owner_org_dest']
+        del pkg_dict['owner_org_dest']
+
+        # Update package
+        # TODO: Validation errors are impossible here for normal flow
+        # We also set type in context to allow type switching by ckan patch
+        context = {'model': model, 'user': toolkit.c.user, 'type': 'dataset'}
```
amercader: We will review the whole flow at a later stage and maybe it won't be necessary to do it here but to validate the dataset we can just extract the [relevant bits from `package_update` ](https://github.com/ckan/ckan/blob/master/ckan/logic/action/update.py#L292:L301), using the default schema and the data_dict.

roll:
@amercader 
Please review. 

Could you please also take a look on failing tests on Travis. It works locally. Do we set correct test.ini for Travis?

amercader:
@roll for the failures on travis, didn't you patch CKAN core in the compose setup? are you doing the same on Travis?

ckanext/unhcr/controllers/deposited_dataset.py
```diff
@@ -0,0 +1,83 @@
...
+
+    # Check dataset exists and it's deposited
+    if pkg_dict.get('owner_org') != depo['id']:
+        message = 'Deposited dataset "%s" not found' % pkg_dict['id']
+        raise toolkit.ObjectNotFound(message)
+
+    # Check the user is sysadmin or data curator
+    granted = lib_helpers.check_access('sysadmin')
+    if not granted:
+        granted = lib_helpers.user_in_org_or_group(depo['id'])
```
amercader: Only users of the data deposit with role `admin` or `editor` should be able to approve / reject datasets. This will return True even if they are `member`.

ckanext/unhcr/helpers.py
```diff
@@ -1,17 +1,45 @@
...
 def render_tree(top_nodes=None):
     '''Returns HTML for a hierarchy of all data containers'''
     context = {'model': model, 'session': model.Session}
     if not top_nodes:
         top_nodes = toolkit.get_action('group_tree')(
             context,
             data_dict={'type': 'data-container'})
+
+    # Remove data deposit
+    # TODO: better way ot remove data deposit from search/tree?
```
amercader: See #78. We can do that separately

ckanext/unhcr/controllers/deposited_dataset.py
```diff
@@ -0,0 +1,83 @@
...
+        # Change to regural dataset
+        pkg_dict = helpers.convert_deposited_dataset_to_regular_dataset(pkg_dict)
+
+        # Update package
+        # We also set type in context to allow type switching by ckan patch
+        update_context = _get_no_authz_context(type=pkg_dict['type'])
+        try:
+            pkg_dict = update_core.package_update(update_context, pkg_dict)
+        except toolkit.ValidationError:
+            message = 'Deposited dataset "%s" is invalid' % pkg_dict['id']
```
amercader: Can we include the actual validation error as part of the exception message? Otherwise is hard to debug issues like the current failures in Travis.

ckanext/unhcr/controllers/deposited_dataset.py
```diff
@@ -0,0 +1,83 @@
...
+            raise toolkit.NotAuthorized(message)
+
+        # Send approval email
+        #  mail_data_container_update_to_user({}, pkg_dict, event='approval')
+
+        # Show flash message and redirect
+        toolkit.h.flash_success('Datasest "{}" is approved'.format(pkg_dict['title']))
```
amercader: 'Datasest "{}" approved and moved to the destination data container'

ckanext/unhcr/controllers/deposited_dataset.py
```diff
@@ -0,0 +1,83 @@
...
+        _raise_not_authz_or_not_deposited(pkg_dict)
+
+        # Send rejection email
+        #  mail_data_container_update_to_user({}, pkg_dict, event='rejection')
+
+        # Purge rejected dataset
+        delete_core.dataset_purge(context, {'id': id})
+
+        # Show flash message and redirect
+        toolkit.h.flash_error('Datasest "{}" is rejected'.format(pkg_dict['title']))
```
amercader: 'Datasest "{}" rejected

ckanext/unhcr/controllers/deposited_dataset.py
```diff
@@ -0,0 +1,83 @@
...
+    if pkg_dict.get('owner_org') != depo['id']:
+        message = 'Deposited dataset "%s" not found' % pkg_dict['id']
+        raise toolkit.ObjectNotFound(message)
+
+    # Check the user is sysadmin or data curator
+    granted = lib_helpers.check_access('sysadmin')
+    if not granted:
+        granted = lib_helpers.user_in_org_or_group(depo['id'])
+    if not granted:
+        message = 'Deposited dataset "%s" is forbidden' % pkg_dict['id']
```
amercader: I think 'Not authorized to perform this action' makes more sense here

ckanext/unhcr/helpers.py
```diff
@@ -1,17 +1,45 @@
...
+def get_data_container(id, context=None):
+    context = context or {'model': model}
+    return toolkit.get_action('organization_show')(context, {'id': id})
+
+
+def get_all_data_containers(exclude_ids=[]):
+    data_containers = []
+    context = {'model': model, 'ignore_auth': True}
+    orgs = toolkit.get_action('organization_list')(context,
+        {'type': 'data-container', 'all_fields': True})
```
amercader: Do we need `all_fields` here?

roll:
@amercader 
Fixed. Thanks

amercader:
One thing is worth adding to this PR or a separate one shortly after is the ability to deposit a dataset without being on a specific data container. For now this can just be a deposit dataset button in the dataset search page, and then the destination owner org field would need to be filled in the form

ckanext/unhcr/helpers.py
roll: Yes. It's used for form so we need titles etc

ckanext/unhcr/controllers/deposited_dataset.py
roll: Fixed. Now it's more consistent. I assumed that roles on data deposit actually doesn't mean (all are data curators). It's true for now, but may lead to incorrect things later.

roll:
@amercader 
Please take a look - I've added my changes.

PR #83, opened 2019-03-03
Deposited dataset state

### LINKED ISSUE #77: Deposited dataset states and associated permissions
We'll add a new field to track the state of deposited datasets. This will affect the permissions of the different roles and track on what stage of the curation process it is.

The overview of the whole process is as follows:

* The data depositor created a new deposited dataset (state = draft)
* They can do further changes, and submit it for curation (state = submitted)
* The curation admin(s) assign an editor to the dataset (state = in_progress)
* If they need the depositor to make changes to the dataset, they unlock it (state = draft)
* Once the depositor has made the changes, they re-submit it (state = in_progress, because there is already an assigned editor)

**Update:**

* Once the editor is happy and validation passes, the dataset is sent back to the depositor for final approval (state = approval_pending)
* The depositor does a final review and accepts the publication (state=published), or rejects it (state=in_progress)

## Tasks

- [x] Add a new custom field name `curation_state` (It's better not to use the default schema `state` as this is hardcoded in too many places). For now, the possible values for this will be :
   * `draft`
   * `submitted`
   * `in_progress`
   * `approval_pending`
   * `published`
  
- [x] Add a validator to ensure that unless the logged in user is member of the curation team, the value is always `draft` (ie default for new deposited datasets).
- [x] When the depositor is happy with the dataset, they can submit it for curation (state changes to `submitted` if there is no curation editor assigned, `in_progress` otherwise). Here's a sample of how it could be displayed:
screenshot_2019-02-18 testingtitle - u s department of education

- [x] When the curation admin assigns an editor to the dataset the state changes to `in_progress`
- [x] Curation admins and editors can change the state to `draft`, so the depositor can make further changes.
- [x] If the deposited dataset is in `draft` state, only the original depositor can edit it (we can match against the logged in user using `creator_user_id`)
- [x] If the deposited dataset is in `in_progress` state, only admins or editors of the Data Deposit container can edit it
- [x] When validation passes, Editors can request a final review from the original depositor. The depositor can still not edit the dataset, but will be able to publish the dataset or reject it, returning it to the editor
- [x] Datasets with a curation state of `published` have been migrated to the destination data container and changed their types to `dataset` so they inherit whatever normal permissions apply

roll:
- fixes #77


roll:
@amercader 
Please review

roll:
@amercader 
I've added:
- review state
- assignable admins

Makefile
```diff
@@ -9,6 +9,11 @@ assets:
 	npx grunt concat
 	npx grunt postcss
 
+# TODO:
+# Move to https://github.com/okfn/docker-ckan/blob/master/2.7/ckan-dev/setup/start_ckan_development.sh
+debug:
+	docker-compose -f ../../docker-compose.dev.yml exec ckan-dev /usr/bin/paster --plugin=ckan config-tool production.ini -s DEFAULT "debug = true"
```
amercader: Isn't this set on the image? https://github.com/okfn/docker-ckan/commit/44c99b20b03af548900a263544651a81dbe77104

ckanext/unhcr/auth.py
```diff
@@ -107,34 +111,38 @@ def organization_create(context, data_dict):
     return {'success': False, 'msg': 'Not allowed to create a data container'}
 
 
+# Package
+
+# TODO: shall we check that user is logged in?
```
amercader: No. Anon users are automatically refused before calling the auth function.

ckanext/unhcr/controllers/deposited_dataset.py
```diff
@@ -16,76 +12,234 @@ class DepositedDatasetController(toolkit.BaseController):
 
     def approve(self, id):
 
-        # Get access and check authz
-        context = _get_no_authz_context()
-        pkg_dict = get_core.package_show(context, {'id': id})
-        _raise_not_authz_or_not_deposited(pkg_dict)
+        # Get dataset
+        context = _get_context()
```
amercader: Maybe we can wrap these three calls in a helper function and return all vars in one go to avoid duplication?

```
context, dataset, curation = _get_curation_data(dataset_id, user_id)
```

ckanext/unhcr/controllers/deposited_dataset.py
```diff
@@ -16,76 +12,234 @@ class DepositedDatasetController(toolkit.BaseController):
...
+        dataset = _get_deposited_dataset(context, id)
+        curation = helpers.get_deposited_dataset_user_curation_status(
+            dataset, toolkit.c.userobj.id)
 
-        # Change to regural dataset
-        pkg_dict = helpers.convert_deposited_dataset_to_regular_dataset(pkg_dict)
+        # Authorize context
+        if 'approve' not in curation['actions']:
+            message = 'Not authorized to approve dataset "%s"'
+            raise toolkit.NotAuthorized(message % id)
```
amercader: This will cause a 500 Internal Server error because you are not catching the exception anywhere. Do this:

```python
return toolkit.abort(403, message)
```

ckanext/unhcr/controllers/deposited_dataset.py
amercader: Also as you are doing a `package_show` call unless you are a curator or the depositor you will get a `NotAuhtorized` exception if you access the endpoint as another user, so you should catch it:

```python
try:
     context, dataset, curation = _get_curation_data(dataset_id, user_id)
except toolkit.NotAuthorized as e:
     return toolkit.abort(403, str(e))
```

PR #91, opened 2019-03-14
Curation activities

### LINKED ISSUE #88: Curation activity streams
We need to keep track of all the activity around the different stages of the curation process a deposited dataset goes through. CKAN has a builtin method for storing these activities, activity streams, so it makes sense to reuse them to take advantage of the dataset models, actions and rendering snippets available. Sadly activity streams don't provide great interfaces for extending them from extensions (until https://github.com/ckan/ckan/pull/3972 gets merged) so we'll need to do some hacky stuff.

The idea is to store activities at the relevant curation points using the normal way (calling `activity_create`), but always storing a `curation_activity` property in the arbitrary `data` object (eg `dataset_deposited`, `dataset_submitted`, `curator_assigned` etc). We will then override the relevant activity list functions to filter out the curation activities (when requesting the public listing) or just including those (when requesting the private curation activities stream).

Tasks:

- [x] Custom `package_activity_list` action that includes or excludes curation specific activities (**Note**: this requires a patch in [CKAN core](https://github.com/CivicActions/docker-ckan-ed/blob/master/ckan/patches/ckan/package_activity_list_extensions.patch).

- [ ] Add `activity_create` calls at the relevant places. Most of these will be in the custom `package_create` and `package_update` action and in the controller. The list of things we need to record initially are:
    
    - [x] A dataset is first deposited
    - [x] A deposited dataset is submitted for curation (initially or after review)
    - [x] An Admin assigns a curator (include the user name in the `data` object)
    - [x] A Curator requests changes (unlocks) the deposited dataset (include the feedback in the `data` object)
    - [x] A Curator approves the deposited dataset
    - [x] A Curator requests a final review to the depositor (include the feedback in the `data` object)
    - [x] A Curator rejects the deposited dataset (include the feedback in the `data` object)
    - [x] A Depositor approves the deposited dataset
    - [x] A Depositor withdrawn the deposited dataset

  Here's an example of a call:
    ```python
          activity_context = {    
              'ignore_auth': True     
          }    
          data_dict = {    
              'user_id': toolkit.c.user,    # Logged in user
              'object_id': 'f1575503-6758-40a5-9157-5b06db1ec259',     # Dataset id
              'activity_type': 'changed package',    # Let's always use this type for now
              'data': {    
                  'curation_activity': 'changes_requested',    
                  'package': {    
                      'name': 'test-curation-5',    
                      'id': 'f1575503-6758-40a5-9157-5b06db1ec259',    
                  }
                  'feedback': 'Please add more detail on the notes field'
              }    
          }    
      
          toolkit.get_action('activity_create')(activity_context, data_dict)
    ```
   At this point curation activities should not be shown in the "Activity stream" tab of the dataset page.

- [ ] Add another endpoint / tab in the dataset page "Curation Activity", only visible to users that can edit the dataset. This should render the workflow activities only (calling `package_activity_list` with `get_workflow_activities=True` in the data_dict) **TODO**: Clarify who can see this tab once the dataset is published.

- [ ] Add a custom activity snippet renderer function so we can display the relevant messages. Sadly for this we need to monkey-patch CKAN core.

amercader:
Fixes #88 

Adds curation activities for the deposited datasets. 

Screenshot_2019-03-19 Some deposited dataset 2 - CKAN

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,4 +31,4 @@ shell:
 	docker-compose -f ../../docker-compose.dev.yml exec ckan-dev bash
 
 test:
-	docker-compose -f ../../docker-compose.dev.yml exec ckan-dev nosetests --ckan --nologcapture --reset-db -s -v --with-pylons=/srv/app/src_extensions/ckanext-unhcr/test.ini --with-coverage --cover-package=ckanext.unhcr --cover-inclusive --cover-erase --cover-tests /srv/app/src_extensions/ckanext-unhcr/ ${ARGS}
+	docker-compose -f ../../docker-compose.dev.yml exec ckan-dev nosetests --ckan --nologcapture -s -v --with-pylons=/srv/app/src_extensions/ckanext-unhcr/test.ini --with-coverage --cover-package=ckanext.unhcr --cover-inclusive --cover-erase --cover-tests /srv/app/src_extensions/ckanext-unhcr/ckanext/unhcr/tests/test_controllers.py:TestDepositedDatasetController ${ARGS}
--- a/ckanext/unhcr/controllers/deposited_dataset.py
+++ b/ckanext/unhcr/controllers/deposited_dataset.py
@@ -86,7 +86,7 @@ class DepositedDatasetController(toolkit.BaseController):
             context = _get_context(ignore_auth=True)
             curator = toolkit.get_action('user_show')(context, {'id': curator_id})
             helpers.create_curation_activity('curator_assigned', dataset['id'],
-                dataset['name'], user_id, curator_id=curator_id, curator_name=curator['name'])
+                dataset['name'], user_id, curator_name=curator['name'])
         else:
             helpers.create_curation_activity('curator_removed', dataset['id'],
                 dataset['name'], user_id)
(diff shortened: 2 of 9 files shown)


amercader:
@roll this is ready

Makefile
```diff
@@ -31,4 +31,4 @@ shell:
 	docker-compose -f ../../docker-compose.dev.yml exec ckan-dev bash
 
 test:
-	docker-compose -f ../../docker-compose.dev.yml exec ckan-dev nosetests --ckan --nologcapture --reset-db -s -v --with-pylons=/srv/app/src_extensions/ckanext-unhcr/test.ini --with-coverage --cover-package=ckanext.unhcr --cover-inclusive --cover-erase --cover-tests /srv/app/src_extensions/ckanext-unhcr/ ${ARGS}
+	docker-compose -f ../../docker-compose.dev.yml exec ckan-dev nosetests --ckan --nologcapture -s -v --with-pylons=/srv/app/src_extensions/ckanext-unhcr/test.ini --with-coverage --cover-package=ckanext.unhcr --cover-inclusive --cover-erase --cover-tests /srv/app/src_extensions/ckanext-unhcr/ckanext/unhcr/tests/test_controllers.py:TestDepositedDatasetController ${ARGS}
```
roll: Not sure I understand how it works now)

ckanext/unhcr/controllers/deposited_dataset.py
```diff
@@ -149,12 +157,15 @@ def request_review(self, dataset_id):
         dataset['curation_state'] = 'review'
         dataset = toolkit.get_action('package_update')(context, dataset)
 
-        # TODO: incorporate to activity/email
         message = toolkit.request.params.get('message')
-        log.debug('Action message: %s' % message)
 
         # Update activity stream
-        #
+        context = _get_context(ignore_auth=True)
```
roll: I think the local `context` var is already to same (with ignore_auth)

ckanext/unhcr/plugin.py
```diff
@@ -6,6 +7,9 @@
 from ckan.lib.plugins import DefaultTranslation
 from ckan.lib.plugins import DefaultPermissionLabels
 
+# 🙈
```
roll: I have to learn how to insert monkeys :sob:

Makefile
amercader: :scream: that wasn't mean to be commited

ckanext/unhcr/plugin.py
amercader: :speak_no_evil:

PR #92, opened 2019-03-17
Deposited dataset form UI

### LINKED ISSUE #87: Deposited dataset form UI improvements
- [x] Update the form's sidebar
- [x] We need a way to show Data Curators what fields are mandatory in a regular dataset while being in the deposited dataset form, so they don't have to go back an forth to the dataset read page. Using the red asterisks for required datasets would be confusing, as they can still save the datasets without filling these. One option is to use a special background for these fields that are required in a normal dataset, something like this (obviously looking better):
screenshot_2019-02-28 datasets - ckan
- [x] Data Depositors should not be thrown directly into the form. I think a previous page would be useful with an explanatory text describing the process, and where they would select the destination data container and the visibility setting. When they clicked a button they would be directed to the deposited dataset form. We'll need a custom endpoint and template for this, which is fine.
- [x] Can we show the actual validation errors for the resources on deposited datasets? Right now it just says "Resources invalid" without showing the missing fields.

estemate: 8h

roll:
- fixes #87


roll:
dataset
resource

roll:
@amercader 
Please review

roll:
@amercader 
Please re-review - I've updated.

Regarding wrongly highlighted fields - I was relaying on resource's names. It's not an id as for packages..

roll:
Merging for demo. I will use additional PR if other changes are needed

PR #94, opened 2019-03-21
Curation UI

### LINKED ISSUE #84: Data Deposit page and curation UI
We need an index page that lists all deposited datasets, accessible only to the curation team, where they can filter them by some specific fields.

I think it makes sense to have this separate from the user dashboard tabs in case the number of deposited datasets is very big, and to take advantage of all the faceting, pagination etc.

We have two options for implementing this page

1. The index page for an organization created in the new org type for the deposited datasets (eg `/data-deposit/deposit`)
2. Use the index page for the `deposited-dataset` dataset type, eg `/deposited-dataset`

I don't have a strong preference. It depends on which one is easier to lock down via authorization.

### Tasks

- [ ] Change permissions so only members of the curation team (ie members of the data deposit org) can access this page (related to #78)
- [ ] Point link in the header to this page. Link should be only visible to curators and admins. Move link to the right. Change icon
- [ ] Remove all facets, and include two new ones: Curation stage, and Curator assigned (if possible, if not show Assigned/Unassigned)
- [ ] Remove tabs (activity, etc)
- [ ] Display curation status in the dataset items in the listing, and also the original depositor and the editor assigned (if any)

---

From #81 

Overview of the Data Deposit and curation UI:

Issues prioritized in the milestone: https://github.com/okfn/ckanext-unhcr/milestone/1

* Data Deposit:  It is only accessible to sysadmins and curation team members. All deposited datasets are shown there, and there are filters for stage (draft, submitted, pending approval...) and for who is assigned to the datasets. (See #84).
* Custom deposited dataset page and form. (See #85 and #87).
* Tabs on the user dashboard (See #86)
   * For all users: My deposited datasets
   * For curator admins: Unassigned deposited datasets
   * For curator editors and admins: Deposited datasets assigned to me
* Curation activity stream (TBD)

----

From #81 

Keeping this here for reference:

If a curator approves a private dataset they will be redirected to a 404 page, as they don't have permissions to see private datasets that belong to another data container. After discussion with the client this is fine, we can sort it out at the app level for now, making editors sysadmins.

roll:
- fixes #84

### DIFF
--- a/ckanext/unhcr/plugin.py
+++ b/ckanext/unhcr/plugin.py
@@ -100,6 +100,7 @@ class UnhcrPlugin(plugins.SingletonPlugin, DefaultTranslation, DefaultPermission
             facets_dict.clear()
             facets_dict['curation_state'] = _('State')
             facets_dict['curator_name'] = _('Curator')
+            facets_dict['depositor_name'] = _('Depositor')
             return facets_dict
         else:
             return self._facets(facets_dict)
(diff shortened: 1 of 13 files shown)


roll:
@amercader 
Please review

roll:
Here is a list of all problems/improvements I found at the moment - https://github.com/okfn/ckanext-unhcr/issues/96

ckanext/unhcr/plugin.py
```diff
@@ -176,6 +186,28 @@ def before_index(self, pkg_dict):
...
+        # Index curator name for deposited dataset
+
+        if pkg_dict.get('type') == 'deposited-dataset':
+            # curator
+            curator_id = pkg_dict.get('curator_id')
+            if curator_id:
+                try:
+                    curator = toolkit.get_action('user_show')(
+                        {'ignore_auth': True}, {'id': curator_id})
+                    pkg_dict['curator_name'] = curator.get('name')
```
amercader: Should we use display_name here to make it more user friendly?

Same below

PR #97, opened 2019-03-27
Curation improvements

### LINKED ISSUE #96: Finalize curation implementation
# Overview

Here is a list of all outstanding issue aside main functionality (see main issues)

# Tasks

- [x] user gets 403 accessing the curation stream
- [x] hide not needed entries in curation stream
- [x] submit message is not in the curation stream
- [x] more prominent messages in curation stream (?)
- [x] ---
- [x] curation UI should show target container for datasets
- [x] if possible make `Data Deposit` header active for all/most curation pages
- [x] use display names for depositors/curators for contacs/filters/etc (?)
- [x] add links to user names in sidebar/dataset-item?
- [x] facet pills use `organization` url instead of data-container -> #99
- [x] Index user display names (depositor and curators), not `name`
- [x] user can't see curator email -> #98 
- [x] ---
- [x] data deposit org creation script
- [x] review `get_data_deposit` usage performance (caching?)
- [x] review/clean dashboard activities?
- [x] user testing of the whole workflow

roll:
- fixes #96

### DIFF
--- a/ckanext/unhcr/fanstatic/theme.js
+++ b/ckanext/unhcr/fanstatic/theme.js
@@ -10,6 +10,7 @@ $( document ).ready(function() {
   });
 });
 
+// Validation
 $(document).ready(function() {
 
   // Get fields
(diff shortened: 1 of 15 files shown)


ckanext/unhcr/fanstatic/theme.js
```diff
@@ -34,6 +35,15 @@ $(document).ready(function() {
 
 });
 
+// Header
```
amercader: This is duplicated in curation.js no?

roll: It's built from curation.js. We don't need this step actually

PR #107, opened 2019-04-14
Fixed dataset count on the homepage

### LINKED ISSUE #90: Datasets counter on homepage includes deposited datasets
# Ovierview

We use `IPermissionLabels` to hide deposited datasets from search but it somehow leaks to homepage.

roll:
- fix #90

### DIFF
--- a/ckanext/unhcr/helpers.py
+++ b/ckanext/unhcr/helpers.py
@@ -29,6 +29,11 @@ def get_all_data_containers(exclude_ids=[]):
     return data_containers
 
 
+def get_dataset_count():
+    return toolkit.get_action('package_search')(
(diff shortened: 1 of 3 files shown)


ckanext/unhcr/helpers.py
```diff
@@ -29,6 +29,11 @@ def get_all_data_containers(exclude_ids=[]):
     return data_containers
 
 
+def get_dataset_count():
+    return toolkit.get_action('package_search')(
+        {}, {'fq': 'dataset_type:dataset', 'rows': 1})['count']
```
amercader: FYI you can actually use `rows: 0` if you only want the count

roll: Thanks. I copied from core)

PR #110, opened 2019-04-16
Email messaging improvements

roll:
- add messages to emails
- send emails only to assigned curator on submit


roll:
@amercader 
done

PR #113, opened 2019-04-19
Pending requests page

### LINKED ISSUE #108: WP03: Centralized page for managing incoming requests
There should be a centralized page where admins can have an overview of all the pending actions they need to take on the site. Right now we have one type of request action: requests to sysadmins for the creation of new data containers, but in the future we might have more (like requests to access a restricted dataset), so the feature should be implemented in a generic way.

What I think makes more sense is a New tab on the dashboard page (Pending Requests) that lists all items (I'd actually default to this tab for sysadmins, as the News feed one is of less interest)

For a first stage, in terms of implementation I'm thinking

- [x] new `pending_requests_list` action that accepts a user id and returns the pending requests. For now this would only work for sysadmins and return a list of pending data containers
- [x] New tab in dashboard + controller + template
- [x] Helper to get the pending requests + snippets for each request type, ie for pending data containers display:
     * Title and Link to data container
     * Submitter
     * Approve / Reject buttons

@roll can you think of something curation-wise that would make sense to add to this page? Maybe submitted deposited datasets without curator, that would be shown to Curator Admins?

roll:
- fixes #108
- improved pending data containers UI (deposited datasets' approach)


roll:
@amercader 
Please take a look

What would you recommend to add other template vars to the controller's endpoint like core does:
```
    def _setup_template_variables(self, context, data_dict):
        c.is_sysadmin = authz.is_sysadmin(c.user)
        try:
            user_dict = get_action('user_show')(context, data_dict)
        except NotFound:
            h.flash_error(_('Not authorized to see this page'))
            h.redirect_to(controller='user', action='login')
        except NotAuthorized:
            abort(403, _('Not authorized to see this page'))

        c.user_dict = user_dict
        c.is_myself = user_dict['name'] == c.user
        c.about_formatted = h.render_markdown(user_dict['about'])
```

ckanext/unhcr/helpers.py
```diff
@@ -151,6 +151,34 @@ def get_linked_datasets_for_display(value, context=None):
     return datasets
 
 
+# Pending requests
+
+def get_pending_requests(all_fields=False, context=None):
```
amercader: It is a best practice to define this as an action (eg `pending_requests_list`) so it gets automatically exposed via the API. And then the helper just calls the action. This way the logic gets separated more clearly.

ckanext/unhcr/templates/user/dashboard_requests.html
```diff
@@ -0,0 +1,39 @@
+{% extends "user/dashboard.html" %}
+
+{% block dashboard_activity_stream_context %}
+{% endblock %}
+
+{% block primary_content_inner %}
+  <h2>{{ _('Data Containers') }}</h2>
```
amercader: Perhaps let's say "Requested Data Containers"

ckanext/unhcr/templates/user/dashboard_requests.html
```diff
@@ -0,0 +1,39 @@
...
+  {% if requests.containers %}
+    <hr>
+    <ul class="dataset-list unstyled">
+      {% for container in requests.containers %}
```
amercader: Can we show some more info here? eg created by X on Y. Y should be the orgs `created` field, and X the first admin if exists (as the user that created the pending org will be assigned as admin)

ckanext/unhcr/templates/organization/snippets/pending_sidebar.html
```diff
@@ -0,0 +1,50 @@
...
+
+  {# Actions #}
+  <section class="module module-narrow module-shallow curation-actions">
+    <h2 class="module-heading"><i class="fa fa-medium fa-bars"></i>Actions</h2>
+
+    <p class="module-content">
+      {% if h.check_access('sysadmin') %}
+        As an administrator you can approve or reject the data container.
+      {% else %}
+        As not an administrator you have no actions available.
```
amercader: Please change to "No actions available" or similar

ckanext/unhcr/templates/user/dashboard_requests.html
```diff
@@ -0,0 +1,39 @@
...
+        <li class="dataset-item">
+          <div class="dataset-content">
+            <h3 class="dataset-heading">
+              <a href="/data-container/{{ container.name }}">{{ container.display_name }}</a>
+            </h3>
+            {% if notes %}
```
amercader: This is not showing the pending data container description 
```suggestion
            {% if container.description %}
```

ckanext/unhcr/templates/user/dashboard_requests.html
amercader: And the parent data container if present in the metadata.

PR #125, opened 2019-04-27
Manage membership page

### LINKED ISSUE #124: Implement custom pages to manage data containers membership

roll:
- fixes #124
---
It requires some polishing and tests


amercader:
@roll Looks great, but I think with some changes it will work much better and be more flexible:

* I'm not sure the dashboard is the best place for this. Without looking at the code it took me a while where I could find it so I think it's going to be difficult for sysadmins too. Perhaps the best place is on the Data Containers page, add an Actions section like we have for the Data Deposit with a yellow Manage memberships button. The URL would then be `/data-container/memberships` (hopefully there will never a be a DC named like this :) )

* Even if it takes a bit to load I think it would be much more friendly if the user list was pre-populated (ie calling `user_list` in the controller). And add the option to search like in the Data Container field in the dataset form

* Can you load the rest of the UI when the user is selected, rather than when clicking "Manage membersip" (ie remove this button)?

* Same as users for the Data Container field, can we pre-populate them like we do in the dataset form? Also ideally can we select more than one in one go to assign the user to multiple DC (with the same role of course). IIRC we had fields like that in LA Counts. You'll need a small change in the controller to allow multiple orgs.

* Can you add the edit member button on the current DC list? Same as in the individual DC members page. URL is `/data-container/member_new/<org-id>?user=<user-id>`:

Screenshot_2019-05-09 Members - UK - Data Containers - CKAN

roll:
@amercader 
All are updated. Please review.

PR #137, opened 2019-05-12
Fixed "Request Data Container" button

### LINKED ISSUE #112: Use "Request Data Container" button instead of "Save Data Container"

roll:
- fixes #112

---

Whatever changes to `organization/snippet/organization_form.html` don't affect the button's text.


roll:
Probably we can do it alternatively by:
- moving `unhcr` ahead of `scheming` within the plugins list
- overring the block `save_text` in the `scheming/organization/group_form` template

amercader:
Yeah, this is not very elegant. I think moving `unhcr` ahead of `scheming` should be safe. If not the second option sounds good.

PR #141, opened 2019-05-12
Improved title/name approach for dataset copying

### LINKED ISSUE #127: Improvements in dataset name when making a copy
1. Check if the "{name}-copy" name already exists and add a suffix if it does (eg another "copy"). This is to avoid duplicate name errors
2. The JS widget for automatically updating the URL based on the title doesn't work when making a copy (perhaps because we are passing in a name?). Maybe we don't need to pass it to the form and trigger the widget on load so it computes the URL again.
Screenshot_2019-05-06 Create Dataset - RIDL

roll:
- fixes #127 

---

It fixes the problem with the title/slug widget (BTW I wasn't able to find any solution to fix the widget if we pass title to the form). 

Regarding naming.. I'm not sure what's the best solution. I think we have to check with the UN. Using things like `dataset (copy)`, `dataset (copy) (copy)` etc - I guess they don't want to have any of them in the database. So for now, I've just added a note and an ability to copy the dataset's name for further manual editing while the title field is empty by default.

If needed we can update it to have `generate auto title` link creating something like `dataset (copy-<time-or-something>)` but again I'm not sure it's what the RIDL is looking for.

Resources naming is also a question..


roll:
@amercader 
It's ready. An empty title field and removed `copy title` link:
copy

PR #144, opened 2019-05-13
Send confirmation email on depositing

### LINKED ISSUE #132: Send a confirmation email to the depositor when depositing a dataset
We generally don't send email notifications or actions that the user themselves are doing but this is an exception to have a record of the depositing.

roll:
- fixes #132


ckanext/unhcr/actions.py
```diff
@@ -2,16 +2,37 @@
 from ckan import model
 from urlparse import urljoin
 from ckan.plugins import toolkit
+from ckan.lib.helpers import url_for
```
amercader: Use the one from the toolkit please

PR #156, opened 2019-05-16
Add cache to get_data_deposit function

### LINKED ISSUE #103: Fix get_data_deposit function
# Overview

We have `get_data_deposit` function which called a lot especially for curation UI (datasets listing):

```
def get_data_deposit():
    try:
        context = {'model': model, 'ignore_auth': True}
        return toolkit.get_action('organization_show')(context, {'id': 'data-deposit'})
    except toolkit.ObjectNotFound:
        log.error('Data Deposit is not created')
        return {'id': 'data-deposit'}
```

A naive caching solution works fine but make tests failing:

```
deposit = None
def get_data_deposit():
    global deposit
    if not deposit:
        try:
            deposit =  toolkit.get_action('organization_show')(
                {'model': model, 'ignore_auth': True}, {'id': 'data-deposit'})
        except toolkit.ObjectNotFound:
            log.error('Data deposit is not created')
            return {'id': 'data-deposit'}
    return deposit
```
with an error:

```
ERROR: ckanext.unhcr.tests.test_controllers.TestDepositedDatasetController.test_activites_not_shown_on_normal_dataset('depositor',)
----------------------------------------------------------------------
  File "/usr/lib/python2.7/site-packages/nose/case.py", line 381, in setUp
    try_run(self.inst, ('setup', 'setUp'))
  File "/usr/lib/python2.7/site-packages/nose/util.py", line 471, in try_run
    return func()
  File "/srv/app/src_extensions/ckanext-unhcr/ckanext/unhcr/tests/test_controllers.py", line 62, in setup
    user=self.creator)
  File "/usr/lib/python2.7/site-packages/factory/base.py", line 80, in __call__
    return cls.create(**kwargs)
  File "/usr/lib/python2.7/site-packages/factory/base.py", line 453, in create
    return cls._generate(True, attrs)
  File "/usr/lib/python2.7/site-packages/factory/base.py", line 378, in _generate
    obj = cls._prepare(create, **attrs)
  File "/usr/lib/python2.7/site-packages/factory/base.py", line 358, in _prepare
    return cls._create(target_class, *args, **kwargs)
  File "/srv/app/src/ckan/ckan/tests/factories.py", line 336, in _create
    **kwargs)
  File "/srv/app/src/ckan/ckan/tests/helpers.py", line 109, in call_action
    return logic.get_action(action_name)(context=context, data_dict=kwargs)
  File "/srv/app/src/ckan/ckan/logic/__init__.py", line 457, in wrapped
  File "/srv/app/src/ckan/ckan/logic/action/create.py", line 176, in package_create
    raise ValidationError(errors)
ValidationError: {'owner_org': ['Invalid data deposit']}
```

PS.
Additional problem that first failed tests make all following tests fail with an error below. It complicates diagnostics. 

```
DetachedInstanceError: Instance <Revision at 0x7f229c0c8f10> is not bound to a Session; attribute refresh operation cannot proceed
```

roll:
- fixes #103


roll:
@amercader 
It seems I have finally made it working. But I really need help on:

- is it correct way to check for the test env (without disabling it for tests it seems a dead-end. tests just fail)?
- is it correct way to use `app_globals` (wasn't able to find docs)

 I think we need to fix this issue until it's forgotten because we already do a lot of calls to `get_data_deposit`.

ckanext/unhcr/helpers.py
```diff
@@ -7,6 +8,7 @@
 from ckan.plugins import toolkit
 import ckan.lib.helpers as core_helpers
 import ckan.lib.plugins as lib_plugins
+from ckan.lib.app_globals import app_globals
```
amercader: Use `toolkit.c`, which is essentially the same but more future proof

ckanext/unhcr/helpers.py
```diff
@@ -167,32 +169,24 @@ def get_pending_requests(all_fields=False, context=None):
...
 def get_data_deposit():
-    try:
-        context = {'model': model, 'ignore_auth': True}
-        return toolkit.get_action('organization_show')(context, {'id': 'data-deposit'})
-    except toolkit.ObjectNotFound:
-        log.error('Data Deposit is not created')
-        return {'id': 'data-deposit', 'name': 'data-deposit'}
+    """This function uses a cache so it's OK to call it multiple times
+    """
+    testing = os.environ.get('CKAN_TESTING')
```
amercader: This is a hack specific to this project. Can you try if `config['testing']` works?

ckanext/unhcr/helpers.py
roll: @amercader 
I always thought that `toolkit.c` is request-specific. Does `toolkit.c` have application-long life-time?

ckanext/unhcr/helpers.py
roll: Will do. I wasn't able to find this in Google)

amercader: Did config['testing'] work?

ckanext/unhcr/helpers.py
amercader: I'd need to check. Leave like this for now then

ckanext/unhcr/helpers.py
roll: Thanks. It works.

ckanext/unhcr/helpers.py
roll: Other option I think is just to use a global var:
```python
cached_deposit = None
def get_data_deposit():
    """This function uses a cache so it's OK to call it multiple times
    """
    deposit = None
    if not config.get('testing'):
        deposit = cached_deposit
    if deposit is None:
        try:
            context = {'model': model, 'ignore_auth': True}
            deposit = toolkit.get_action('organization_show')(
                context, {'id': 'data-deposit'})
            if not config.get('testing'):
                cached_deposit = deposit
        except toolkit.ObjectNotFound:
            log.error('Data Deposit is not created')
            deposit = {'id': 'data-deposit', 'name': 'data-deposit'}
    return deposit
```

amercader: Let's do this then and not mess with the global vars. This will change in future versions and will save us some headaches

PR #157, opened 2019-05-17
Implemented final review request option

### LINKED ISSUE #134: Indicate to curators if the dataset needs a final review
Right now when the deposited dataset is finally valid, curators need to decide whether to publish it directly or request a final review from the depositor.

This choice should be done by the depositor when depositing the dataset. In the popup we can add a checkbox saying "I want to do a final review of the dataset before it is published". This should be stored in the dataset as a hidden `curation_*` extra and used once the dataset is valid to show one button or the other ("Request final review" or "Approve dataset").

@roll can you think of a better way to implement this?

roll:
- fixes #134


roll:
@amercader 
Please review

PR #159, opened 2019-05-21
Implemented deposited drafts

### LINKED ISSUE #129: Draft datasets don't get created as deposited
Steps to reproduce:

1. Deposit a dataset, fill the metadata and click on "Next: Add Data"
2. Leave the page (don't add any data)
3. The dataset is not on the Data Deposit, but it appears as draft in the "My Datasets" dashboard tab

I'm not sure if the problem is that 

a. The dataset is not created as `deposited-dataset`
b. It is created as deposited dataset but the Data Deposit page doesn't list drafts (real drafts, not curation drafts)

From the client:
> #3: I had a problem (again) with the file upload during a deposit creation (the button is just not reacting, seems like the js is missing for me). It then timed out, and when I logged back in, I couldn’t find my draft deposit under the Data Deposit button. However, I have found it under My Datasets (please see below). Is that the expected behaviour?

roll:
- fixes #129


roll:
@amercader 
Please take a look. 
I'm not sure about terminology because we now kinda have two draft states.

ckanext/unhcr/templates/package/read.html
```diff
@@ -3,47 +3,54 @@
...
+      {# Not active #}
+      {% if not curation['active']  %}
 
-	  {% if curation['role'] == 'depositor' %}
-	    <p>{% trans %}This deposited dataset is in <strong>draft</strong> state. You can make further edits if necessary. When you want to submit it to the curation team, click on the "Submit Dataset" button on the left.{% endtrans %}</p>
-		<p>{% trans %}<strong>Note:</strong> You won't be able to edit the dataset once submitted. You can submit a dataset for curation even if it does not pass the validation requirements.{% endtrans %}</p>
-	  {% else %}
-		<p>{% trans %}This deposited dataset is in <strong>draft</strong> state. The depositor can still make further edits if necessary.{% endtrans %}</p>
-	  {% endif %}
+        <p>{% trans %}This dataset is not deposited yet.{% endtrans %}</p>
```
amercader: We should probably tell users what they need to do to deposit it, as it is not clear from the actions on the left. We can say something like:

```
<p>{% trans %}This dataset is not deposited yet. To deposit it finish editing the metadata and add some resources to it.{% endtrans %}</p>
```

PR #179, opened 2019-06-21
Microdata integration

roll:
Discussion document - https://docs.google.com/document/d/1fLImgYXX0JPAajC8avBB1tUBW55iF4b9Gj-GRzb5vaw/edit#heading=h.l8bu1s519p3f


amercader:
@roll Getting these errors on all cases:
First time I push to ML:
```
 Dataset "Tst final review3" publishing to the Microdata library is not completed for the following reason: "{u'title': u'The Title field is required.'}" 
```

Second time on same dataset:
```
 Dataset "Tst final review3" publishing to the Microdata library is not completed for the following reason: "IDNO already exists. 82" 
```

Which suggest that the survey is created but when creating the resources the error is raised. Looking at [the docs](https://microdata.unhcr.org/api-documentation/catalog-admin/#operation/addDatasetResources) it doesn't look like `title` is mandatory in resources but the errors seems to suggest otherwise. So if `resource['name']` is empty you should use some placeholder [here](https://github.com/okfn/ckanext-unhcr/pull/179/files#diff-8010d0abaa06260eb7ec9e627a990c38R628)

ckanext/unhcr/tests/test_actions.py
```diff
@@ -0,0 +1,109 @@
...
+
+class TestActions(base.FunctionalTestBase):
+
+    # General
+
+    def setup(self):
+        super(TestActions, self).setup()
+
+        # Config
+        config['ckanext.unhcr.microdata_api_key'] = 'API-KEY'
```
amercader: No need to change it but reference, it's generally cleaner to use the `@helpers.change_config('ckanext.unhcr.microdata_api_key', 'API-KEY')` decorator on each test, as it only changes the config in the scope of that test and cleans up after it, to not affect other tests elsewhere

ckanext/unhcr/controllers/extended_package.py
```diff
@@ -99,3 +103,34 @@ def resource_download(self, id, resource_id, filename=None):
...
+    def publish_microdata(self, id):
+        context = {'model': model, 'user': toolkit.c.user}
+        nation = toolkit.request.params.get('nation')
+        repoid = toolkit.request.params.get('repoid')
+
+        # Get dataset
+        try:
+            dataset = toolkit.get_action('package_show')(context, {'id': id})
+        except (toolkit.ObjectNotFound, toolkit.NotAuthorized):
+            message = 'Not authorized to to publish of dataset "%s"'
```
amercader: typo

roll:
> Looking at the docs it doesn't look like title is mandatory in resources but the errors seems to suggest otherwise.
And there are a lot of stuff like this..

Issue #148, opened 2019-05-14
WP02: Show private datasets in results


amercader:
### Implementation overview

CKAN core has public / private datasets. On RIDL we essentially need to ensure that all datasets are public, and on top of that add a custom auth check just on resource downloading, to check if the logged in user belongs to the dataset data container

| CKAN private field    | Custom RIDL visibility field  | RIDL label
| --------------------- | ----------------------------- | ----------
| private=True          | private                       | Private [*] 
| private=False         | restricted                    | Public (for logged-in users) with restricted downloads
| private=False         | public                        | Public (for logged-in users)

[*] Will be probably shown to sysadmins only

### Tasks

- [x] Add a validator for the default visibility field (`private`) to ensure that is always True (except if you are a sysadmin)

- [x] Add a new custom boolean field at the dataset level (`restricted`). If that field is True, it means that the resources can only be downloaded by members of the team. This field should replace the current visibilty field in the form, with the same current values and labels. 

- [x] Implement a custom `resource_download` auth function (register it with `IAuthFunctions`). If the dataset has `visibility=restricted` then the logged in user must belong to the dataset org to be able to download it. You can get the logged in user with `context['user']`, and the organizations they belong to using the [`get_organization_list_for_user`](https://docs.ckan.org/en/latest/api/index.html#ckan.logic.action.get.organization_list_for_user) action. Use `package_show` to get the full dataset details.

- [x] The resource download [endpoint](https://github.com/ckan/ckan/blob/2.8/ckan/controllers/package.py#L1151) checks access using `resource_show`, so that won't work for us. We need to override this whole function in our own controller (`IBlueprint` + custom blueprint in the extension), and customize so it calls `check_access` on the new `resource_download` auth function before any of the current code.

- [x] Limit access to the DataStore on restricted resources. Override the `datastore_search` and `datastore_search_sql` [auth functions](https://github.com/ckan/ckan/blob/master/ckanext/datastore/logic/auth.py#L49) to rely on on our own `resource_download` rather than `resource_show`.
- [x] Update the UI to reflect restricted datsets. We can follow the same convention the private datasets use with the small label with the lock, but apply it to datasets where `visibility=restricted`. The places where this needs to be updated are :

 * The dataset listings in the search, teams page etc: `package_item.html` snippet.

 * The dataset read page: `package/read.html` template

[...]

PR #187, opened 2019-07-04
Added private dataset tests

roll:
- connects #148


roll:
@amercader 
Could you please help.. I've into dreadful stuck with these tests:
- `test_access_visibility_private_pages_not_visible` - doesn't show 404 for a private dataset in tests (though works locally or on UAT)
- datastore tests just does't work after porting from the source. It seems we need some additional preparation steps to make it work
- and Travis still requires some tweaks to allow us to test with `datastore`

PR #194, opened 2019-07-15
Show "Please log in" on dataset if not logged in

### LINKED ISSUE #131: Redirect to login page if not authorized, not 404
When a user goes to a dataset or data container page and they are not logged in they are shown a 404 page (a custom UNHCR error).

Screenshot_2019-05-06 Page Not Found

This is standard CKAN behaviour and a something most content websites do (as to not disclose if a resource exist or not) but on this case it makes sense to redirect to the login page (ideally redirecting to the original linked page after logging in (`came_from` parameter).

This is related to #128 because logging out the user makes it surface much more often.

In terms of implementation I'm afraid there will be some controller overriding involved.

roll:
- fixes #131

### DIFF
--- a/ckanext/unhcr/controllers/extended_group.py
+++ b/ckanext/unhcr/controllers/extended_group.py
@@ -11,5 +11,5 @@ class ExtendedGroupController(GroupController):
 
     def read(self, id):
         if not toolkit.c.user:
-            return toolkit.render('page.html')
+            return toolkit.abort(403, toolkit.render('page.html'))
         return super(ExtendedGroupController, self).read(id)
--- a/ckanext/unhcr/controllers/extended_group.py
+++ b/ckanext/unhcr/controllers/extended_group.py
@@ -1,0 +1,15 @@
+import logging
+from ckan import model
+import ckan.plugins.toolkit as toolkit
+from ckan.controllers.group import GroupController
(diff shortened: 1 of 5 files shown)


roll:
@amercader 
For an anonymous user this page `http://ckan-dev:5000/dataset/` shows (with `403` return code):


I think it will be consistent if:
- http://ckan-dev:5000/dataset/name
- http://ckan-dev:5000/data-container/name (I need to add to this PR)

will show the same. But I'm not sure how to set `403` return code.

ckanext/unhcr/controllers/extended_group.py
```diff
@@ -0,0 +1,15 @@
...
+log = logging.getLogger(__name__)
+
+
+class ExtendedGroupController(GroupController):
+
+    # Read
+
+    def read(self, id):
+        if not toolkit.c.user:
+            return toolkit.render('page.html')
```
amercader: @roll, to show the please login page better to raise a 403 as you suggest:

```suggestion
            return toolkit.abort(403, 'Message here!')
```

roll:
@amercader 
It's updated. Please review. It seems works but please double check that it's safe to do this trick with the `OrganizationController` regarding scheming etc

PR #215, opened 2019-08-27
Fix deposited notifications timing

### LINKED ISSUE #210: Depositing notification email is sent before the dataset is saved
# Overview

We still need notification on a dataset creation but after it's moved to be not a draft.

roll:
- fix #210


roll:
@amercader 
does this solution make sense?

roll:
Thanks. Regarding API I think it's not important because probably it will use other business rules (e.g. mass-creation will lead to a lot of emails etc). So, for now, it should be fine

PR #216, opened 2019-08-27
Draft with one resource publishing

### LINKED ISSUE #206: Drafts with one file can't be completed without submitting another one
This is a core behaviour. We need to figure out the best way to patch it.

roll:
- fixes #206 

---

@amercader 
Please take a look and test if possible. It looks like this now:

draft


amercader:
Works great, added some tweaks in fe17d19

PR #217, opened 2019-08-27
Fixed deposited resources access

### LINKED ISSUE #213: Not authorized to submit a dataset error (403)
# Overview

See a screenshot - https://docs.google.com/document/d/1iWDzBx75BbOigY-ENeYyuBveWMmmwBH8J6NqNXu_lQw/edit#heading=h.apqooacy56dr

roll:
- fix #213 

---

@amercader 
I was able to reproduce the exact error that `wilkinso` had on the `sens-chad-2018-2019` dataset but based on this text from screenshot `Access was denied to this resource` I checked whether we were correct with resources access for deposited datasets. And it seems we weren't - a depositor wasn't able to see uploaded resources.

So here I'm adding a depositor to the white-list of cases to bypass the `visibility=restricted` check.

I think it should fix the original issue.

WDYT?


roll:
@amercader 
Take a look at the screenshot I mean the underlying exception is `Access was denied to this resource`. So I think it's a good chance that it's related.

Though I wasn't able to reproduce so we need more information it's true

amercader:
Maybe we are not talking about the same screenshot? I mean the one under point 25 in the RIDL bugs doc

roll:
No, actually you're right. I had a different understanding of what's happened

roll:
Ok. Thanks. Merging as not related to #213

PR #218, opened 2019-08-28
Data collector free text

### LINKED ISSUE #200: Make data collector a free text field (handle DDI import)
# Overview

From Mariann:

> it seems one of the livelihoods datasets (trying to figure out which) imported from DDI had two values for the data collector field: one was UNHCR, the other was the Jordan River Foundation, which does not exist in RIDL.
> It seems the DDI was imported successfully with only UNHCR as data collector. 

IIRC this was handled correctly but maybe there is a regression. Or maybe the data collector id is not one of the ones allowed

Example provided:

https://ridl.unhcr.org/dataset/unhcr-jor-2017-lis-1-1

# Tasks

- [ ] convert the field to a `free text`
- [ ] use autocomplete widget?
- [ ] handle multiple values 
- [ ] create a migration script

roll:
- fixes #200


scripts/migrate_data_collectors.py
```diff
@@ -0,0 +1,32 @@
...
+    datasets = ckan.action.package_list()
+    for dataset in datasets:
+        labels = []
+        dataset = ckan.action.package_show(id=dataset)
+        for value in helpers.normalize_list(dataset.get('data_collector')):
+            label = collectors.get(value, value)
+            labels.append(label)
+            dataset['data_collector'] = ','.join(labels)
+            # TODO: it doesn't work because of validation
+            ckan.action.package_update(**dataset)
```
roll: @amercader 
Can you suggest something - I need to suppress the validation. It just fails for every dataset, for now.

roll:
@amercader 
Please test and review

roll:
@amercader 
I think I've resolved all the problems. Please test.

ckanext/unhcr/fanstatic/theme.js
```diff
@@ -70,6 +70,61 @@ $( document ).ready(function() {
   $(".hierarchy-tree-top .highlighted").parents(".closed").removeClass("closed").addClass("open").children(".hierarchy-tree").removeClass("collapse");
 });
 
+$( document ).ready(function() {
```
amercader: Do we need this here if we have `data-collector.js`? Seems duplicated

scripts/migrate_data_collectors.json
```diff
@@ -0,0 +1,162 @@
+[
```
amercader: This file is no longer needed IIUC the script right?

scripts/migrate_data_collectors.py
```diff
@@ -0,0 +1,52 @@
...
+            engine.execute("update %s set value = '%s' where id = '%s' and key = 'data_collector'" % (table, value, row.id))
+            print('Updated dataset "%s" in table "%s"' % (row.package_id, table))
+
+    # Notify about finishing
+    print('Migration is finished')
+    print('Do not forget to rebuild the index: "paster --plugin=ckan search-index --config=production.ini rebuild"')
+
+
+if __name__ == '__main__':
+    url = sys.argv[1]
```
amercader: args not needed

roll:
@amercader
Have you run the migration script?

ckanext/unhcr/fanstatic/theme.js
roll: Yea that's our weird way to compile JS files) 

I think we need to drop this `src` folder and put the code directly to `fanstatic`

scripts/migrate_data_collectors.json
roll: This is needed to run the migration script

roll:
This PR requires:
- https://github.com/okfn/docker-ckan-unhcr/pull/14
- https://github.com/okfn/ckanext-ddi/pull/2

amercader:
This seems to work fine. There is only one thing that needs to be fixed. When migrating the old records we need to store the labels, not the existing values as these are the ids:

Screenshot_2019-08-30 Test old data collectors - RIDL

As you have the labels on the JSON object next to the migration script it should be easy to check for each existing old value if there is a label to be used, something like (untested):

```python
        for row in rows:
            old_value = row.value
            try:
                old_value = json.loads(old_value)
            except Exception:
                continue
            new_value = []
            for item in old_value:
                if item in collectors.keys():
                    new_value.append(collectors[item])
                else:
                    new_value.append(item)
            new_value = ','.join(json.loads(row.value))

```

PR #221, opened 2019-08-30
Added file type "other"

### LINKED ISSUE #220: Error during DDI import related to `file_type`
# Overview

Trying to import any DDI dataset but getting in the UI:

> http://ckan-dev:5000/dataset/import

```
Errors were found in the XML file:

import: Dataset import from XML failed: Could not import dataset test-ddi: Could not upload file: {u'file_type': [u'Missing value']}
```

roll:
- fix #220
- requires https://github.com/okfn/ckanext-ddi/pull/3


roll:
@amercader 
These 2 PRs fix the issue. 

I think we need to mention it to Mariann (ask if this is OK or we need to use some other type) but I don't see another option to solve it TBH.

PR #231, opened 2019-10-22
[WIP] Rebase on bootstrap3

roll:
# To get started
- it's better to delete docker volumes related to this project
- clone/checkout https://github.com/okfn/docker-ckan-unhcr/pull/18 (branch `bootstrap3`)
- `cd src`
- clone/checkout https://github.com/okfn/ckanext-unhcr/pull/231 (branch `bootstrap3`)
- `cd ckanext-unhcr`
- `make docker`
- `make start`
- open `http://ckan-dev:5000`
- login `http://ckan-dev:5000/user/login` (`ckan_admin:testpass`)


smth:
Hi

https://github.com/okfn/docker-ckan-unhcr/ 404s for me. Perhaps its private?

To clarify the objectives here, the task is to ensure that this extension works with a CKAN version that uses Bootstrap 3. Is that right?

I would think that this essentially means:

- Replacing any no longer supported utility classes that might be used.
- Correcting for any differences in reset / base styles.

Is this your expectation? Anything else you are expecting to need fixing?

roll:
Hi @smth!

I've sent you an invite.

Our high-level expectation is to make it look and work as it is on the `docker-ckan-unhcr@master` branch (basically your initial design). 

@amercader how can we give access to the staging site? I think it's easier to use as a reference but now with SAML2 I can't log in as an admin

roll:
@smth 
Access to the staging site - https://github.com/okfn/docker-ckan-unhcr/pull/18#issue-331018999

smth:
@roll would you like me to push changes directly to this branch? I don't appear to have access to do so.

roll:
@smth 
Yes, please push to this branch. I've sent you an invite to get write access

smth:
Thanks. Have just pushed some changes that I think should cover at least most of what is needed. Probably a good idea to check it on an instance with more data in it.

This was a little more complicated than I speculated above, because there are changes to the core theme, that go beyond swapping out classes. It's more difficult to predict what is broken than one might think.

A few points to note:

- I noticed `ddi-theme.css` is linked, but doesn't seem to exist. I don't know what this is.
- `/data-container/data-deposit` is 404ing on my instance, so haven't been able to look at that.
- CSS is not minified. I was under the impression that Fanstatic handled that, but it doesn't seem to be happening. Minification could easily be added to the CSS build step.
- There are some things that probably need to be fixed at the core theme level. For instance, I noticed that select2 inputs (e.g tags) look a bit broken, regardless of this theme.
- The error you get if you fail to log in should probably be fixed :p Much fun and games were had with the wrong log in credentials above.

amercader:
@smth thanks a lot for this, looking really good. Before addressing your points, some things I noticed:

* The width of the main container has increased a bit, not sure if it was a conscious decision or part of the upgrade to BS3. In wide screens like mine it feels a bit too stretched, but I don't have a strong opinion against it so if  you are happy with it let's leave it.

(New Bootstrap 3 on top, Boostrap 2 below in all cases)


Screenshot_2019-10-31 Welcome - RIDL

Screenshot_2019-10-31 Welcome - RIDL(1)

---

Screenshot_2019-10-31 Datasets - RIDL(1)

Screenshot_2019-10-31 Datasets - RIDL

---

Screenshot_2019-10-31 Test data collector - RIDL(1)

Screenshot_2019-10-31 Market Survey 2009 - RIDL

 * Can you fix a bit the styling in the Membership page (http://ckan-dev:5000/data-container/membership?username=ckan_admin) I think it's just missing a few classes (helper text, labels on the right, etc)

Bad:
Screenshot_2019-10-31 Manage Membership - RIDL(2)

Good:
Screenshot_2019-10-31 Manage Membership - RIDL(3)

To your points:
>    I noticed ddi-theme.css is linked, but doesn't seem to exist. I don't know what this is.
Thanks, fixed
>    /data-container/data-deposit is 404ing on my instance, so haven't been able to look at that.
You just new to create a new data container named "Data deposit"

>    CSS is not minified. I was under the impression that Fanstatic handled that, but it doesn't seem to be happening. Minification could easily be added to the CSS build step.
Feel free to add it but I don't think it's a big deal

>    There are some things that probably need to be fixed at the core theme level. For instance, I noticed that select2 inputs (e.g tags) look a bit broken, regardless of this theme.
Yes, these are a pain. I fixed them on this branch.

>    The error you get if you fail to log in should probably be fixed :p Much fun and games were had with the wrong log in credentials above.
Sorry about that. I've now pushed a fix (requires a rebuild of the image)

roll:
I think fanstatic will minify it for production

amercader:
@roll can you fix this merge conflict? it relates to your changes on the external access field

smth:
Yeah, the increased max width is a feature of Bootstrap 3. It could be constrained of course, but in general I think the extra width might be useful.

I have pushed some fixes for the Membership page, thanks for highlighting that.

Is that everything?

amercader:
@smth many thanks for this! This is done AFAIC

amercader:
@smth Sorry, one last thing. If the client wants to reduce the container width, can you paste here how would we do that? Would we change the values in `layout.css`?

Just covering some bases in advance :)

smth:
In that case I would suggest that you use the Bootstrap 3 "Medium device" container width as the maximum (it is 30px wider than the old maximum width). To do this you would need to:

Override the container width, by adding this somewhere (`layout.css` is a good suggestion)

```css
@media (--desktop-up) {
  .container {
    width: var(--container-landscape-tablet-up);
  }
}
```

Remove any styles (from all CSS files) that assume the wider width. Do that by removing all (other) occurrences of
```css
@media (--desktop-up) {
      ...
}
```

amercader:
Fantastic, thanks @smth!

PR #232, opened 2019-10-23
Always include sub-containers to container_read search

### LINKED ISSUE #229: Rename child data container checkbox, select by default
This one:

Screenshot_2019-10-18 Americas - Data Containers - RIDL

Rename label to "Include Sub-Containers" and also change the default to checked

roll:
- fix #229

----

sdc


roll:
@amercader 
I've ended up having this checkbox removed because if it's selected by default (hence it shows all datasets by default) there is no way to disable it unless we create the `Exclude sub-containers` checkbox or something like this. But it will require more additional work.

I'm going to ask Mariann if with no checkbox is OK. Search for sub-containers still is available via the containers hierarchy on the left and via the container facets selector on the left.

ckanext/unhcr/templates/snippets/search_form.html
```diff
@@ -10,9 +10,6 @@
       </button>
       {% endblock %}
     </div>
-     {% if include_children_option %}
```
amercader: Sorry if it wasn't clear but I think the requirement was to keep the checkbox, but checked by default.

roll: @amercader 
Take a look at my comment above

amercader: Sorry, missed that, makes sense.

PR #233, opened 2019-10-24
Rebase on external access level

### LINKED ISSUE #230: Changes to External Access field
This currently maps to the core license field, but needs to be replaced by a selector of the following values:

* Not available
* Direct access
* Public use
* Licensed use
* Data enclave
* Open access

As these are not proper licenses is best to create a new separate field `external_access` for this. 
Existing datasets should be updated to set `Not available` on this new field (most don't define a license currently https://ridl.unhcr.org/api/action/package_search?facet=true&facet.field=[%22license_id%22]&rows=0)

roll:
- fixes #230


ckanext/unhcr/templates/snippets/license.html
```diff
@@ -1,5 +1,6 @@
-{% ckan_extends %}
-
-{% block license_title %}
-  <h2 class="module-heading"><i class="fa fa-medium fa-lock"></i> {{ _('External access level') }}</h2>
-{% endblock %}
+<section class="module module-narrow module-shallow license">
```
amercader: Why do we need a custom snippet and helper here? Can't we use the standard `select` snippet?

roll: @amercader 
It's for `package_read` sidebar

roll:
@amercader 
It's ready for a review

ckanext/unhcr/templates/snippets/license.html
amercader: sorry for the noise, I need to pay more attention

amercader:
Given that `external_access_level` is a required field, we need to run a migration to provide a default value on all existing datasets. Otherwise if you edit the resources of an existing dataset you get an error, because CKAN internally calls package_update to add/edit resources and this fails with the following:

Screenshot_2019-10-28 Edit - Unnamed resource - Test data collector - RIDL

We just need to update all existing datasets once to have `externa_access_level=not_specified`. We can adapt the script in `scripts/update_private_datasets.py` for this.

roll:
@amercader 
Oh I see I'll add the script

roll:
@amercader 
I have added the script

roll:
Adria, please test the script if you'd like. I'm merging to simplify the `ckan-2.7` branch creation

PR #263, opened 2020-02-21
get the test suite passing

chris48s:
ebe612e is pretty straightforward: https://github.com/okfn/ckanext-unhcr/commit/3b249f6cf699ad51523dd690d0376c92e1a3a78a changed the key from `'upload'` to `'url'` but didn't update the test assertions.

Others are basically about resources having an associated upload.
Should be good for review

### DIFF
--- a/ckanext/unhcr/tests/test_actions.py
+++ b/ckanext/unhcr/tests/test_actions.py
@@ -380,7 +380,10 @@ class TestResourceUpload(base.FunctionalTestBase):
 
         resource = factories.Resource(
             package_id=dataset['id'],
-            upload=test_upload)
+            upload=test_upload,
+            url = "http://fakeurl/test.txt",
+            url_type='upload',
+        )
 
         assert_equals(
             resource['url'],
--- a/ckanext/unhcr/tests/test_actions.py
+++ b/ckanext/unhcr/tests/test_actions.py
@@ -401,10 +401,10 @@ class TestResourceUpload(base.FunctionalTestBase):
                 package_id=dataset['id'],
                 url='https://example.com/some.data.csv')
 
-        assert exc.exception.error_dict.keys() == ['upload']
+        assert exc.exception.error_dict.keys() == ['url']
 
         assert_equals(
-            exc.exception.error_dict['upload'],
+            exc.exception.error_dict['url'],
             ['All resources require an uploaded file'])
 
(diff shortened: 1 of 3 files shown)


ckanext/unhcr/tests/test_actions.py
```diff
@@ -380,7 +380,10 @@ def test_upload_present(self):
 
         resource = factories.Resource(
             package_id=dataset['id'],
-            upload=test_upload)
+            upload=test_upload,
+            url = "http://fakeurl/test.txt",
```
chris48s: Setting a fake value for `url` here get the test to pass. Without it, we'll thow something like

`AssertionError: 'http://test.ckan.net/dataset/ce36d426-9e5b-4bd6-9560-b1c595187c7b/resource/4be5611b-d535-497d-8a0a-988f095a79c3/download/link.to.some.data' != 'http://test.ckan.net/dataset/ce36d426-9e5b-4bd6-9560-b1c595187c7b/resource/4be5611b-d535-497d-8a0a-988f095a79c3/download/test.txt'
`

..but intuitively (not really knowing the internals), it seems like for a file upload, we should be picking this up from `filename` in `FakeFileStorage`. Does what I've done here seem right?

amercader: Are you sure this change is needed? This test passes fine for my locally without any change

chris48s: This is the test which was passing locally but failing in CI.

For me it also passes locally without this. In fact, you can delete `url_type='upload'` and it'll pass locally too.

On travis, if you set `url_type='upload'` but don't have this line `url="http://fakeurl/test.txt",` it fails like this:

https://travis-ci.org/chris48s/ckanext-unhcr/builds/655292086#L1209

amercader: Interesting. So on Travis, the file name from `FakeFileStorage` is ignored and the URL one is used (the URL coming from the default value in the [factory](https://github.com/ckan/ckan/blob/master/ckan/tests/factories.py#L146)).
Oh well, let's not waste more time on this. If we get more weird inconsistencies we'll investigate more. Let's merge this.

chris48s: Exactly. I just wish I could explain why the behaviour is different in CI then it is in dev

Issue #265, opened 2020-03-12
Create blueprint and template for metrics dashboard


amercader:
Create a new blueprint that renders a template for the metrics dashboard. 
The extension currently has Pylons controllers because we needed to extend the core dataset and organization ones, but this is a separate endpoint so let's use Flask based blueprints.

- [ ] Add a new `blueprints.py` module with a Blueprint that registers a `/metrics` (or `/site-dashboard`, anything but `/dashboard`) ([example](https://github.com/ckan/ckanext-dcat/blob/master/ckanext/dcat/blueprints.py))
- [ ] Create a new template and render it from the view function above. For the first version we can include easy stuff like:
* Number of datasets (`package_search`)
* Number of data containers (`organization_list`)
* Number of users on the site (`user_list`)
- [ ] Using `IBlueprint` register the new blueprint in `plugin.py`

chris48s:
Going to close this one as we've got the basics in place now. Lets treat further "add a table/graph for X" as new issues

Issue #266, opened 2020-03-12
Add a way to store time-based metrics


amercader:
Some of the indicators we want to show on the metrics dashboard require us to store a value over time, eg number of datasets over time, number of data containers, etc

For the datasets we could probably construct that with data from the search but it's better to have a generic way to cache a certain value regularly and get the full time series directly.

I'll leave the final implementation to you @chris48s but essentially we need a command that runs daily or weekly (depending on the granularity needed) and stores the necessary values in some kind of cache (db table, JSON file).

Let's start with something easy like number of datasets and data containers over time.

- [x] Create a command that stores what's needed. Sadly we can't use the new Click-based ones on 2.8 yet, here's an [example](https://github.com/ckan/ckanext-dcat/blob/master/ckanext/dcat/commands.py) of a Paster based one (plus how to [register](https://github.com/ckan/ckanext-dcat/blob/master/setup.py#L52:L53)).
- [x] Modify the Docker images to add a cron-job like process that runs the above regularly (maybe we can park this until we migrate to AWS)
- [ ] Once some timeline data is available, add a chart widget to the metrics dashboard template

amercader:
@chris48s some pointers on:

* Database migrations:
  1. Create model(s) describing your table plus functions to check if exist / create ([example](https://github.com/okfn/ckanext-collaborators/blob/master/ckanext/collaborators/model.py))
  2. Create command that creates the tables if they don't exist ([ex](https://github.com/okfn/ckanext-collaborators/blob/master/ckanext/collaborators/commands.py))
  3. Add an init script to our docker setup (example in the old one [here](https://github.com/okfn/docker-ckan-unhcr/blob/master/ckan/docker-entrypoint.d/03_setup_saml2.sh))

* Fanstatic resources
Docs are [here](https://docs.ckan.org/en/2.8/theming/fanstatic.html) and [here](https://docs.ckan.org/en/2.8/contributing/frontend/resources.html), but you basically add a `resource.config` file that defines the assets groups, like [this one](https://github.com/frictionlessdata/ckanext-validation/blob/master/ckanext/validation/fanstatic/resource.config) and then you include it in the templates like [here](https://github.com/frictionlessdata/ckanext-validation/blob/master/ckanext/validation/templates/validation/snippets/validation_report_dialog.html#L17).

* Cron replacement
I used [yacron](https://github.com/gjcarneiro/yacron) in the past with good results. Dockerfile looks like this:

```dockerfile
[...]

PR #268, opened 2020-03-13
Add metrics dashboard

chris48s:
Still more to do on this, but I think its probably useful to share a Draft PR with some comments at this stage. This will help inform where we go next.

Refs #265

### DIFF
--- a/ckanext/unhcr/helpers.py
+++ b/ckanext/unhcr/helpers.py
@@ -8,6 +8,7 @@ from ckan.logic import ValidationError
 from ckan.plugins import toolkit
 import ckan.lib.helpers as core_helpers
 import ckan.lib.plugins as lib_plugins
+from ckan.logic import NotFound
 from ckanext.hierarchy import helpers as hierarchy_helpers
 from ckanext.scheming.helpers import (
     scheming_get_dataset_schema, scheming_field_by_name
@@ -126,10 +127,13 @@ def get_came_from_param():
 def user_is_curator():
     user = toolkit.c.user
     group = get_data_deposit()
-    users = toolkit.get_action('member_list')(
-        { 'user': user },
-        { 'id': group['id'] }
-    )
+    try:
+        users = toolkit.get_action('member_list')(
+            { 'user': user },
+            { 'id': group['id'] }
+        )
+    except NotFound:
+        return False
     user_ids = [u[0] for u in users]
     user_id = toolkit.c.userobj.id
     return user_id in user_ids
--- a/ckanext/unhcr/helpers.py
+++ b/ckanext/unhcr/helpers.py
@@ -8,7 +8,6 @@ from ckan.logic import ValidationError
 from ckan.plugins import toolkit
 import ckan.lib.helpers as core_helpers
 import ckan.lib.plugins as lib_plugins
-from ckan.logic import NotFound
 from ckanext.hierarchy import helpers as hierarchy_helpers
 from ckanext.scheming.helpers import (
     scheming_get_dataset_schema, scheming_field_by_name
@@ -132,7 +131,7 @@ def user_is_curator():
             { 'user': user },
             { 'id': group['id'] }
         )
-    except NotFound:
+    except toolkit.ObjectNotFound:
         return False
     user_ids = [u[0] for u in users]
     user_id = toolkit.c.userobj.id
--- a/ckanext/unhcr/helpers.py
+++ b/ckanext/unhcr/helpers.py
@@ -123,6 +123,18 @@ def get_came_from_param():
     return toolkit.request.environ.get('CKAN_CURRENT_URL', '')
 
 
+def user_is_curator():
+    user = toolkit.c.user
+    group = get_data_deposit()
+    users = toolkit.get_action('member_list')(
+        { 'user': user },
+        { 'id': group['id'] }
+    )
+    user_ids = [u[0] for u in users]
+    user_id = toolkit.c.userobj.id
+    return user_id in user_ids
+
+
 # Linked datasets
 
 def get_linked_datasets_for_form(selected_ids=[], exclude_ids=[], context=None, user_id=None):
--- a/ckanext/unhcr/metrics.py
+++ b/ckanext/unhcr/metrics.py
@@ -7,10 +7,12 @@ import ckan.plugins.toolkit as toolkit
 
 def get_datasets_by_date(context):
     title = 'Total number of Datasets'
-    packages = toolkit.get_action('package_search')(
-        context,
-        { 'q': '*:*', 'rows': 0 }
-    )
+    data_dict = {
+        'q': '*:*',
+        'rows': 0,
+        'include_private': True,
+    }
+    packages = toolkit.get_action('package_search')(context, data_dict)
 
     return {
         'type': 'line_graph',
@@ -27,6 +29,7 @@ def get_containers(context):
         'rows': 0,
         'facet.field': ['organization'],
         'facet.limit': 10,
+        'include_private': True,
     }
     packages = toolkit.get_action('package_search')(context, data_dict)
 
@@ -59,6 +62,7 @@ def get_tags(context):
         'rows': 0,
         'facet.field': ['tags'],
         'facet.limit': 10,
+        'include_private': True,
     }
     packages = toolkit.get_action('package_search')(context, data_dict)
 
--- a/ckanext/unhcr/metrics.py
+++ b/ckanext/unhcr/metrics.py
@@ -97,7 +97,7 @@ def get_users(context):
         'title': title,
         'id': slugify(title),
         'total': len(users),
-        'headers': ['User', 'Datasets'],
+        'headers': ['User', 'Datasets Created'],
         'data': [
             {
                 'display_name': user['display_name'],
--- a/ckanext/unhcr/metrics.py
+++ b/ckanext/unhcr/metrics.py
@@ -84,10 +84,12 @@ def get_tags(context):
     }
 
 def get_users(context):
-    data_dict = {
-        'order_by': 'number_created_packages'
-    }
-    users = toolkit.get_action('user_list')(context, data_dict)
+    users = toolkit.get_action('user_list')(context, {})
+    users = sorted(
+        users,
+        key=itemgetter('number_created_packages'),
+        reverse=True,
+    )
 
     title = 'Users'
     return {
--- a/ckanext/unhcr/metrics.py
+++ b/ckanext/unhcr/metrics.py
@@ -9,6 +9,7 @@ def get_datasets_by_date(context):
     title = 'Total number of Datasets'
     data_dict = {
         'q': '*:*',
+        'fq': "-type:deposited-dataset",
         'rows': 0,
         'include_private': True,
     }
@@ -26,6 +27,7 @@ def get_datasets_by_date(context):
 def get_containers(context):
     data_dict = {
         'q': '*:*',
+        'fq': "-type:deposited-dataset",
         'rows': 0,
         'facet.field': ['organization'],
         'facet.limit': 10,
@@ -59,6 +61,7 @@ def get_containers(context):
 def get_tags(context):
     data_dict = {
         'q': '*:*',
+        'fq': "-type:deposited-dataset",
         'rows': 0,
         'facet.field': ['tags'],
         'facet.limit': 10,
--- a/ckanext/unhcr/metrics.py
+++ b/ckanext/unhcr/metrics.py
@@ -1,0 +1,118 @@
+# -*- coding: utf-8 -*-
+
+from operator import itemgetter
+from slugify import slugify
+import ckan.plugins.toolkit as toolkit
+
+
+def get_datasets_by_date(context):
+    title = 'Total number of Datasets'
+    packages = toolkit.get_action('package_search')(
+        context,
+        { 'q': '*:*', 'rows': 0 }
+    )
+
+    return {
+        'type': 'line_graph',
+        'short_title': 'Datasets',
+        'title': title,
+        'id': slugify(title),
+        'total': packages['count'],
+        'data': []
+    }
+
+def get_containers(context):
+    data_dict = {
+        'q': '*:*',
+        'rows': 0,
+        'facet.field': ['organization'],
+        'facet.limit': 10,
+    }
+    packages = toolkit.get_action('package_search')(context, data_dict)
+
+    organizations = toolkit.get_action('organization_list')(
+        context,
+        { 'type': 'data-container' }
+    )
+
+    data = sorted(
+        packages['search_facets']['organization']['items'],
+        key=itemgetter('count'),
+        reverse=True,
+    )
+    for row in data:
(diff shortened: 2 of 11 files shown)


ckanext/unhcr/metrics.py
```diff
@@ -0,0 +1,118 @@
...
+            }
+            for user in users[:10]
+        ]
+    }
+
+def metrics():
+    context = {'user': toolkit.c.user }
+
+    if not toolkit.c.userobj.sysadmin:
+        return toolkit.abort(403, 'Forbidden')
```
chris48s: The requirement is that this page should be available to sysadmins and the curation team. How would I check if a user is a curator?

ckanext/unhcr/templates/header.html
```diff
@@ -8,12 +8,24 @@
 
 {% block header_site_navigation_tabs %}
   {% if c.userobj %}
-    {{ h.build_nav_main(
-        ('data-container_index', _('Organizations')),
-        ('data-deposit', _('Data Deposit')),
-        ('search', _('Datasets')),
-       )
-    }}
+    {% if c.userobj.sysadmin %}
```
chris48s: Same comment here: How do I check if a user is a curator?

ckanext/unhcr/templates/metrics/snippets/line_graph.html
```diff
@@ -0,0 +1 @@
+[ TODO: graph ]
```
chris48s: TODO: issue #266

ckanext/unhcr/metrics.py
```diff
@@ -0,0 +1,118 @@
...
+        { 'q': '*:*', 'rows': 0 }
+    )
+
+    return {
+        'type': 'line_graph',
+        'short_title': 'Datasets',
+        'title': title,
+        'id': slugify(title),
+        'total': packages['count'],
+        'data': []
```
chris48s: TODO: issue #266

ckanext/unhcr/metrics.py
```diff
@@ -0,0 +1,118 @@
...
+                'display_name': user['display_name'],
+                'count': user['number_created_packages'],
+                'link': toolkit.url_for('user.read', id=user['id']),
+            }
+            for user in users[:10]
+        ]
+    }
+
+def metrics():
+    context = {'user': toolkit.c.user }
```
chris48s: I feel like it was slightly unclear earlier whether we actually want to bypass the auth here so that data curators (who might not have all access to all datasets) see the right totals for the entire site. Should this actually be `{ 'ignore_auth': True }`? Also, for similar reasons, would it make sense to be passing `'include_private': True,` in the `data_dict` when running action queries?

ckanext/unhcr/metrics.py
```diff
@@ -0,0 +1,118 @@
...
+
+    if not toolkit.c.userobj.sysadmin:
+        return toolkit.abort(403, 'Forbidden')
+
+    return toolkit.render('metrics/index.html', {
+        'metrics': [
+            get_datasets_by_date(context),
+            get_containers(context),
+            get_tags(context),
+            get_users(context),
```
chris48s: TODO: We could add more summary tables for "Classifications", "Data Collectors", "Operational Purposes", etc here, but maybe it is best to get _something_ in front of some real users first and ask what else they want to know before we spend a lot of time on this.

ckanext/unhcr/metrics.py
amercader: The curation team are users members of the Data Deposit, which is just a CKAN organization. There's a `get_data_curation_users()` helper in `helpers.py` that is used in the Data Deposit interface, but it probably does too much for what you need (you don't need individual users details and it does not return users with role `member` AFAICT).

So it's better if you roll your own little helper function that calls the [`member_list`](https://docs.ckan.org/en/latest/api/index.html#ckan.logic.action.get.member_list) action, using the id returned from the `get_data_deposit()` helper. With this you'll get a list of curation team members, and you just need to match their names against `toolkit.c.user`.

ckanext/unhcr/templates/header.html
amercader: See above. Register a new helper function (in `get_helpers()` from plugin.py) called `is_curation_team_member` or similar and you'll be able to call it from templates using `h.is_curation_team_member()`.

ckanext/unhcr/metrics.py
amercader: I agree

ckanext/unhcr/metrics.py
```diff
@@ -0,0 +1,118 @@
...
+        { 'type': 'data-container' }
+    )
+
+    data = sorted(
+        packages['search_facets']['organization']['items'],
+        key=itemgetter('count'),
+        reverse=True,
+    )
+    for row in data:
+        row['link'] = '/data-container/{}'.format(row['name'])
```
amercader: For custom dataset or org types, you can call url_for using :

```python
toolkit.url_for('data-container_read', id=row['name'])
```

ckanext/unhcr/metrics.py
amercader: We should `include_private: True` on `package_search` calls.
Sysadmis will be able to see anything. Curation team members won't, but I don't think setting `ignore_auth: True` will help with that. We'll have to think about the best way to handle this but for now don't set it.

ckanext/unhcr/metrics.py
```diff
@@ -0,0 +1,118 @@
...
+            {
+                'display_name': user['display_name'],
+                'count': user['number_created_packages'],
+                'link': toolkit.url_for('user.read', id=user['id']),
+            }
+            for user in users[:10]
+        ]
+    }
+
+def metrics():
```
amercader: Let's keep at least the `metrics` view function in the blueprint file, as this is the convention.

ckanext/unhcr/metrics.py
```diff
@@ -0,0 +1,118 @@
...
+    context = {'user': toolkit.c.user }
+
+    if not toolkit.c.userobj.sysadmin:
+        return toolkit.abort(403, 'Forbidden')
+
+    return toolkit.render('metrics/index.html', {
+        'metrics': [
```
amercader: I like this pattern of having a common interface for each "snippet" or widget. It will allow us to easily add new ones or modify them in the future

ckanext/unhcr/helpers.py
```diff
@@ -123,6 +124,21 @@ def get_came_from_param():
...
+def user_is_curator():
+    user = toolkit.c.user
+    group = get_data_deposit()
+    try:
+        users = toolkit.get_action('member_list')(
+            { 'user': user },
+            { 'id': group['id'] }
+        )
+    except NotFound:
+        return False
```
chris48s: The reason I've done this is because this function gets called in the base template. If its not here, any test which renders a template but doesn't create the `data-deposit` container fails. I'm not sure this is conceptually the ideal solution. The other option is we need to explicitly add `factories.DataContainer(users=[], name='data-deposit', id='data-deposit')` to the setup of every test class that renders a template.

chris48s:
Thanks for all the pointers on this. I think this is fairly close to complete :crossed_fingers:

ckanext/unhcr/tests/test_controllers.py
```diff
@@ -1101,3 +1101,35 @@ def test_membership_remove(self):
     def test_membership_remove_no_access(self):
         url = '/data-container/membership_remove?username=default_test&contname=container1'
         resp = self.get_request(url, user='user3', status=403)
+
+
+class TestMetricsController(base.FunctionalTestBase):
```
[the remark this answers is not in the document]
amercader: Sounds good

ckanext/unhcr/metrics.py
```diff
@@ -0,0 +1,107 @@
+# -*- coding: utf-8 -*-
+
+from operator import itemgetter
+from slugify import slugify
+import ckan.plugins.toolkit as toolkit
+
+
+def get_datasets_by_date(context):
+    title = 'Total number of Datasets'
```
amercader: No need to change it in our case but just so you know we tend to make all user-facing strings translatable using `toolkit._('Total number of Datasets')`.
I doubt RIDL will ever be translated but just to keep in mind for other projects.

ckanext/unhcr/metrics.py
```diff
@@ -0,0 +1,107 @@
...
+    users = toolkit.get_action('user_list')(context, data_dict)
+
+    title = 'Users'
+    return {
+        'type': 'freq_table',
+        'title': title,
+        'id': slugify(title),
+        'total': len(users),
+        'headers': ['User', 'Datasets'],
+        'data': [
```
amercader: This should be sorted by `number_created_packages` as currently it is sorted by user name

ckanext/unhcr/metrics.py
```diff
@@ -0,0 +1,107 @@
...
+    }
+    users = toolkit.get_action('user_list')(context, data_dict)
+
+    title = 'Users'
+    return {
+        'type': 'freq_table',
+        'title': title,
+        'id': slugify(title),
+        'total': len(users),
+        'headers': ['User', 'Datasets'],
```
amercader: Let's use "Datasets created" here

ckanext/unhcr/templates/header.html
```diff
@@ -8,12 +8,24 @@
...
-        ('search', _('Datasets')),
-       )
-    }}
+    {% if c.userobj.sysadmin or h.user_is_curator() %}
+      {{
+        h.build_nav_main(
+          ('data-container_index', _('Organizations')),
+          ('data-deposit', _('Data Deposit')),
+          ('search', _('Datasets')),
+          ('unhcr_metrics.metrics', _('Metrics')),
```
amercader: To change the icon check `fonts.css` and `header.css`. This project uses PostCSS
To compile it I use:

    node_modules/grunt/bin/grunt postcss

ckanext/unhcr/templates/metrics/index.html
```diff
@@ -0,0 +1,44 @@
+{% extends "page.html" %}
+
```
amercader: Add `{%- block subtitle %}Metrics{% endblock -%}` here to update the title in the browser

chris48s: Sure.

Can you explain the difference between using

`{%- block subtitle %}Metrics{% endblock -%}` (with the dashes)

and

`{% block subtitle %}Metrics{% endblock %}` (without)?

They seem to do the exact same thing.

ckanext/unhcr/templates/header.html
chris48s: Thanks. I did have a quick look at this. For some reason, this project is using 2 different icon sets: In some places its fontawesome but in the top menus, the icons come from a really cut down version of icomoon which doesn't have a graph/chart icon in it. I've chosen the most appropriate thing I can find in the icomoon set we've got. I'm sure there is a way to regenerate the webfont to include a graph icon from the same icon set as the others, but I'm not going to spend loads of time on it right now.

ckanext/unhcr/metrics.py
chris48s: Hmmm.. that was certainly the intent of:

https://github.com/okfn/ckanext-unhcr/blob/a1684c0d52f547fe1c5021eae976214017976c51/ckanext/unhcr/metrics.py#L83-L86
I'd expect that to sort by `number_created_packages`: https://docs.ckan.org/en/2.8/api/index.html#ckan.logic.action.get.user_list

but you are totally right that they were being returned sorted by username. In fact, whatever field I put in there, I'm getting the user list sorted by username. For the moment I've switched it to just sort them as a python list after to get it working but am I missing something there about how to use `order_by`, or am I hitting a bug in core?

amercader: It's totally a core bug :man_facepalming: 
The `order_by` param is totally ignored [here](https://github.com/ckan/ckan/blob/master/ckan/logic/action/get.py#L750:L757).

If you want to submit a patch this seemed to make it work for me (I have not tested it thoroughly):

```diff
diff --git a/ckan/logic/action/get.py b/ckan/logic/action/get.py
index ce28f4878..d22ce4a53 100644
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -749,17 +749,29 @@ def user_list(context, data_dict):
 
     if order_by == 'edits':
         raise ValidationError('order_by=edits is no longer supported')
-    query = query.order_by(
-        _case([(
-            _or_(model.User.fullname == None,
-                    model.User.fullname == ''),
-            model.User.name)],
-            else_=model.User.fullname))
+
+    order_by_field = None
+    if order_by == 'number_created_packages':
+        order_by_field = order_by
+    elif order_by != 'name':
+        try:
+            order_by_field = getattr(model.User, order_by)
+        except AttributeError:
+            pass
+    if order_by == 'name' or order_by_field is None:
+        query = query.order_by(
+            _case([(
+                _or_(model.User.fullname == None,
+                        model.User.fullname == ''),
+                model.User.name)],
+                else_=model.User.fullname))
+    else:
+        query = query.order_by(order_by_field)
 
     # Filter deleted users
     query = query.filter(model.User.state != model.State.DELETED)

```

For now sorting in Python sounds great in our extension

ckanext/unhcr/templates/header.html
amercader: Sounds good

ckanext/unhcr/templates/metrics/index.html
amercader: They prevent whitespace from appearing, see https://stackoverflow.com/questions/11813422/what-does-this-in-jinja2-template-engine-do

ckanext/unhcr/metrics.py
```diff
@@ -0,0 +1,112 @@
...
+    return {
+        'type': 'freq_table',
+        'title': title,
+        'id': slugify(title),
+        'total': len(users),
+        'headers': ['User', 'Datasets Created'],
+        'data': [
+            {
+                'display_name': user['display_name'],
+                'count': user['number_created_packages'],
```
chris48s, line 107: Reporting `number_created_packages` on the users table is still going to include data deposits in those figures (I've filtered them out from the other tables in 59ab3de ).
I don't think there is a way to do that filter via the `user_list` action, so we probably need to construct a custom query. I'm going to suggest we ignore that for now just for the purposes of getting something on UAT. Lets log it as an issue and come back to it later.

ckanext/unhcr/templates/metrics/index.html
chris48s: :mortar_board:  that's a useful feature :)

ckanext/unhcr/metrics.py
chris48s: For now I will log it as an issue on the CKAN repo so we don't lose sight of it, but I will have a look at getting the tests running locally etc to work on this at some point. Seems like a good first issue.

amercader: It is! I'll help you with the tests if you work on it

ckanext/unhcr/metrics.py
amercader: Sounds good

PR #270, opened 2020-03-20
add DB table and paster commands for capturing snapshots in time

chris48s:
refs #266

Next bits of this jigsaw are:

* Submit a PR on https://github.com/okfn/docker-ckan-unhcr to deal with scheduling
* Look at using this data to make a graph

..but I think this PR can be reviewed as it stands

### DIFF
--- a/ckanext/unhcr/commands.py
+++ b/ckanext/unhcr/commands.py
@@ -56,6 +56,7 @@ class Unhcr(toolkit.CkanCommand):
 
         packages = toolkit.get_action('package_search')(context, {
             'q': '*:*',
+            'fq': "-type:deposited-dataset",
             'rows': 0,
             'include_private': True,
         })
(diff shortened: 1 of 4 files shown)


ckanext/unhcr/commands.py
```diff
@@ -0,0 +1,74 @@
...
+    def init_db(self):
+        if tables_exist():
+            print(u'UNHCR tables already exist')
+            sys.exit(0)
+
+        create_tables()
+        print(u'UNHCR tables created')
+
+    def snapshot_metrics(self):
+        context = { 'user': toolkit.c.user }
```
chris48s: If we run this from a paster command, `toolkit.c.user == 'default'` here which should mean we collect everything, I think

amercader: Commands are always run by a privileged user, and outside the context of a web request so we shouldn't be using `toolkit.c`.
Can you try to pass `{'ignore_auth': True}` as context and see if we get what we need? If not there's a safer way of getting the "default" user:

```python
site_user = toolkit.get_action('get_site_user')({'ignore_auth': True}}
// use site_user['name']
```

chris48s: Just calling `package_search` directly with `{ 'ignore_auth': True }` seems to get the job done :+1:

Issue #267, opened 2020-03-12
Implement weekly summary emails


amercader:
An email sent weekly to curation team members and sysadmins that summarizes the activity on the portal for the previous week.

![image2](https://user-images.githubusercontent.com/200230/76514194-d2cffd80-6457-11ea-90ba-aa80d26f18d8.png)

- [x] Create the mailing function. We already have the infrastructure and templates in place for sending emails, so it's just a matter of extending and combining these on a function that:

  * Gets the recipients of the email (curation team members, `organization_show` on the `data-deposit` org + `h.get_sysadmins()`)
  * Gets the data needed. Let's stick to New datasets, New deposited datasets, Datasets awaiting review. These can all be retrieved with `package_search`, I'll add some examples to get you started
  * Builds the content of the email. There are templates on `templates/email` that can be extended.
  * Sends it. Check the `mailer.py` module.
- [x] Wrap this function in a command that can be called externally (eg `paster --plugin=ckanext-unhcr ridl send_weekly_email`)
- [x] Like in #266 we will need a cron-like task that runs every week and calls the above command to send the email

PR #273, opened 2020-04-03
Weekly summary emails

chris48s:
Refs #267

This PR adds the command to build and send summary emails.
I'll do a separate PR on the docker repo for the scheduling.

Example email:

Screenshot at 2020-04-03 14-05-00

### DIFF
--- a/ckanext/unhcr/mailer.py
+++ b/ckanext/unhcr/mailer.py
@@ -183,8 +183,11 @@ def get_summary_email_recipients():
     curator_ids = [c[0] for c in curators]
 
     all_users = toolkit.get_action('user_list')({ 'ignore_auth': True }, {})
+    default_user = toolkit.get_action('get_site_user')({ 'ignore_auth': True })
 
     for user in all_users:
+        if user['name'] == default_user['name']:
+            continue
         if user['sysadmin'] or user['id'] in curator_ids:
             recipients.append(user)
 
--- a/ckanext/unhcr/mailer.py
+++ b/ckanext/unhcr/mailer.py
@@ -1,10 +1,10 @@
+from datetime import datetime, timedelta
 import logging
 from ckan import model
 from ckan.common import config
 from ckan.plugins import toolkit
 from ckan.lib import mailer as core_mailer
 from ckan.lib.base import render_jinja2
-import ckan.logic.action.get as get_core
 from ckanext.unhcr import helpers
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


ckanext/unhcr/mailer.py
```diff
@@ -79,3 +79,94 @@ def compose_membership_email_body(container, user_dict, event):
...
+# Weekly Summary
+
+def _get_new_packages(context, start_time):
+    data_dict = {
+        'q': '*:*',
+        'fq': (
+            '-type:deposited-dataset AND ' +\
+            'metadata_created:[{} TO NOW]'.format(start_time)
+        ),
+        'include_private': True,
```
chris48s, line 93: I think we want to `include_private` in the summary emails - same as on the dashboard, right?

ckanext/unhcr/mailer.py
```diff
@@ -79,3 +79,94 @@ def compose_membership_email_body(container, user_dict, event):
...
+    start_time = datetime.now() - timedelta(days=7)
+    context['start_date'] = start_time.strftime('%A %B %e %Y')
+    query_start_time = start_time.strftime('%Y-%m-%dT%H:%M:%SZ')
+
+    context['recipient'] = user_dict.get('fullname') or user_dict.get('name')
+    context['site_title'] = config.get('ckan.site_title')
+    context['site_url'] = config.get('ckan.site_url')
+
+    context['datasets_url'] = toolkit.url_for('search', qualified=True)
+    context['deposits_url'] = toolkit.url_for('data-deposit', qualified=True)
```
chris48s, line 139: It would be nice here if we could link to the datasets and deposits pages filtered by date range based on `metadata_created`, or sorted by `metadata_created` (so at least all the stuff we're notifying them about is at the top). Is there a nice way we can do this?

ckanext/unhcr/mailer.py
```diff
@@ -79,3 +79,94 @@ def compose_membership_email_body(container, user_dict, event):
...
+
+    action_context = { 'user': user_dict['name'] }
+    context['new_datasets'] = _get_new_packages(action_context, query_start_time)
+    context['new_deposits'] = _get_new_deposits(action_context, query_start_time)
```
chris48s, line 143: I got a bit confused about `context` objects here. Are we supposed to pass actions the same `context` we pass the template renderer, or are action context and template context two different things? I've kept them separate for now.

ckanext/unhcr/mailer.py
```diff
@@ -79,3 +79,94 @@ def compose_membership_email_body(container, user_dict, event):
...
+    context['awaiting_review'] = _get_deposits_awaiting_review(
+        action_context,
+        query_start_time
+    )
+
+    context['h'] = toolkit.h
```
chris48s, line 149: Maybe this is fine, but it feels like a bit of a hack. Trying to use `toolkit.render()` was throwing `TypeError: No object (name: session) has been registered for this thread`. I think in this case it is nicer to have access to `h.link_to` and `h.url_for` in the template if we can. Is there a cleaner way of doing this?

ckanext/unhcr/tests/test_mailer.py
```diff
@@ -0,0 +1,115 @@
...
+        recipients = get_summary_email_recipients()
+        recipient_ids = [r['name'] for r in recipients]
+
+        assert len(recipient_ids) == 3
+        assert curator['name'] in recipient_ids
+        assert self.sysadmin['name'] in recipient_ids
+        assert (
+            'default_test' in recipient_ids or  # local
+            'test.ckan.net' in recipient_ids  # travis
+        )
```
chris48s, line 114: I don't know why, but for some reason the default user is called `'default_test'` if I run the tests locally but `'test.ckan.net'` if I run them in CI. Not sure why this is. I can't find anywhere we explicitly set the value. I've done this to get the tests to pass, but maybe there's a better solution.

It looks like we've hit this issue before and just ignored it. e.g:

https://github.com/okfn/ckanext-unhcr/blob/f2a31ce7c9dde4bcfaa48bf2b2f6c03046d34623/ckanext/unhcr/tests/test_controllers.py#L1058
https://github.com/okfn/ckanext-unhcr/blob/f2a31ce7c9dde4bcfaa48bf2b2f6c03046d34623/ckanext/unhcr/tests/test_controllers.py#L1072
https://github.com/okfn/ckanext-unhcr/blob/f2a31ce7c9dde4bcfaa48bf2b2f6c03046d34623/ckanext/unhcr/tests/test_controllers.py#L1077

Is there a reliable to get the username of the default user, or can we explicitly specify it somewhere in the setup so its the same locally and in CI?

ckanext/unhcr/mailer.py
amercader: Yes, that's right

ckanext/unhcr/mailer.py
amercader: The URL we want is something like

https://ridl-uat.unhcr.org/dataset?q=metadata_created%3A%5B2020-01-01T00%3A00%3A00.000Z+TO+NOW%5D&sort=metadata_created+desc

So you want something like:

```python
     context['datasets_url'] = toolkit.url_for('search', q='metadata_created:[{} TO NOW]'.format(date_formatted), sort='metadata_created desc, qualified=True) 
# date format should be YYYY-MM-DDTHH:MM:SSZ
```

ckanext/unhcr/mailer.py
amercader: They are separate things. It was unfortunate naming.

ckanext/unhcr/tests/test_mailer.py
amercader: There is an internal sysadmin user created with the name of the `ckan.site_id` setting. In our local setup we use [this one](https://github.com/okfn/docker-ckan-unhcr/blob/master/.dev.env#L53) (which don't come up in grep / ag searches because it's a hidden file :)). In Travis we use a different config file, which uses a different site_id

In any case this user (which you can retrieve by calling `get_action('get_site_user')()`), should not receive any emails so you can filter it out.

ckanext/unhcr/mailer.py
amercader: That's annoying. Ideally we should be using toolkit method to be able to use the `h` functions in the template as you suggest but there's an issue here with requests contexts. Let's go with your approach for now and we'll revisit later.

Issue #269, opened 2020-03-19
Store number of downloads


chris48s:
One of the next requirements for metrics will be around reporting the number of downloads:

- Most downloaded datasets (datasets ranked by downloads)
- Most active users (users ranked by downloads)

Currently we don't track this so the first step is to work out the best way to log this information before we can report it.

amercader:
Aside from the metrics needed by the dashboard there is also a requirements for admins to be able to track who downloaded individual resources so I'd like to try an use activities to track downloads as well.

Be warned, it's quite hacky, but activities on CKAN 2.8 are messy :)

In summary we need to create a custom activity at the time a user requests a file to download. That is the `/dataset/<id>/resource/<resource_id>/download` endpoint, which in our case we are already overriding ([route definition](https://github.com/okfn/ckanext-unhcr/blob/master/ckanext/unhcr/plugin.py#L117), [controller](https://github.com/okfn/ckanext-unhcr/blob/master/ckanext/unhcr/controllers/extended_package.py#L105). Once the activities are created we can query the relevant type to get the counts needed for the metrics dashboard, or implement custom activity streams for authorized users (downloads should not be mixed with the public activities).

A comment on our custom download endpoint. Right now it only handles the default local storage backend for file uploads, but once we move to AWS we are going to switch to ckanext-cloudstorage, which has its own overriden endpoint. So we'll need to defer to their one, but let's cross that bridge when we get there.

We'll focus first on storing the activities and retrieving any counts you need for the metrics.

* Call `activity_create` on the download endpoint using the logged in user id, the dataset id and the UTC timestamp, with a custom `activity_type` like `download resource`.
* Filter out `download resource` activities from the default returned activities by `*_activity_list` functions. We are already doing this for curation related activities, so it's just a matter to add a further filter. See functions starting [here](https://github.com/okfn/ckanext-unhcr/blob/master/ckanext/unhcr/actions.py#L254)
* At this point you should have activities stored in the DB, so you can retrieve what you need either via actions (`activity_list`) or more likely via a bespoke DB query (it will be more performant).

I'll share existing code that should help with this.

chris48s:
- [x] Log resource downloads using an activity when using `ckanext.unhcr.controllers.extended_package:ExtendedPackageController`: https://github.com/okfn/ckanext-unhcr/pull/280
- [x] Use custom activity to produce tables on the metrics dashboard: https://github.com/okfn/ckanext-unhcr/pull/281
[...]

PR #280, opened 2020-04-16
log resource downloads

chris48s:
refs #269

This PR adds only the code to store a custom `'download resource'` activity.
To keep the PRs easier to review, I'll submit a second PR adding outputs to the metrics dashboard.

### DIFF
--- a/ckanext/unhcr/actions.py
+++ b/ckanext/unhcr/actions.py
@@ -317,20 +317,53 @@ def package_activity_list_html(context, data_dict):
         context, activity_stream, extra_vars)
 
 
-# Without this action the `*_activity_list` is not overriden (ckan bug?)
+@toolkit.side_effect_free
 def dashboard_activity_list_html(context, data_dict):
-    activity_stream = dashboard_activity_list(context, data_dict)
-    model = context['model']
+    '''Override core ckan dashboard_activity_list_html action so download resource
+    activities are not rendered in the HTML.
+    '''
+    activity_stream = toolkit.get_action('dashboard_activity_list')(
+        context, data_dict)
+
+    filtered_activity_stream = [
+        a for a in activity_stream if a['activity_type'] != 'download resource'
+    ]
+
     user_id = context['user']
-    offset = data_dict.get('offset', 0)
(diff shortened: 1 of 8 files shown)


ckanext/unhcr/actions.py
```diff
@@ -317,20 +317,53 @@ def package_activity_list_html(context, data_dict):
...
-    activity_stream = dashboard_activity_list(context, data_dict)
-    model = context['model']
+    '''Override core ckan dashboard_activity_list_html action so download resource
+    activities are not rendered in the HTML.
+    '''
+    activity_stream = toolkit.get_action('dashboard_activity_list')(
+        context, data_dict)
+
+    filtered_activity_stream = [
+        a for a in activity_stream if a['activity_type'] != 'download resource'
```
amercader, line 329: We should make this filtering at the `dashboard_activity_list` level, otherwise the download activities will leak at the API level

ckanext/unhcr/actions.py
```diff
@@ -317,20 +317,53 @@ def package_activity_list_html(context, data_dict):
...
+@toolkit.side_effect_free
+def user_activity_list_html(context, data_dict):
+    '''Override core ckan user_activity_list_html action so download resource
+    activities are not rendered in the HTML.
+    '''
+    activity_stream = toolkit.get_action('user_activity_list')(
+        context, data_dict)
+
+    filtered_activity_stream = [
+        a for a in activity_stream if a['activity_type'] != 'download resource'
```
amercader, line 354: Same as above with `user_activity_list`

ckanext/unhcr/actions.py
chris48s: No probs. I've moved the check up a level.
Just so you know, I copied the approach of applying the filter in the `_html` functions straight out of that GDX example you sent me so you might want to make the same change on that project too? (unless the requirements are different there)

chris48s:
For some reason travis has not reported the build status on the commit, but it passed: https://travis-ci.org/github/okfn/ckanext-unhcr/builds/678252734

PR #281, opened 2020-04-16
show dataset downloads on metrics dashboard

chris48s:
refs #269

This PR builds on #280 and uses the `'download resource'` activities to show some extra tables on the dashboard

### DIFF
--- a/ckanext/unhcr/metrics.py
+++ b/ckanext/unhcr/metrics.py
@@ -129,7 +129,7 @@ def get_containers_by_date(context):
         'type': 'timeseries_graph',
         'short_title': 'Containers',
         'title': title,
-        'total': dates.values()[-1],
+        'total': dates.values()[-1] if len(dates.values()) > 0 else None,
         'id': slugify(title),
         'data': [
             ['x'] + [str(date) for date in dates.keys()],
--- a/ckanext/unhcr/metrics.py
+++ b/ckanext/unhcr/metrics.py
@@ -2,7 +2,7 @@
 
 from operator import itemgetter
 from slugify import slugify
-from sqlalchemy import select, func
+from sqlalchemy import select, func, desc
 import ckan.model as model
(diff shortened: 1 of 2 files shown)


ckanext/unhcr/metrics.py
```diff
@@ -149,3 +190,41 @@ def get_users(context):
...
+    ).order_by(
+        desc('count')
+    ).limit(10)
+
+    result = model.Session.execute(sql).fetchall()
+
+    data = []
+    for row in result:
+        data.append({
+            'display_name': row['fullname'] if row['fullname'] else row['name'],
```
chris48s, line 218: Could I have written the above query in a way that hydrates the User object with the `fullname` property, or do I only get that if I query users via an action?

amercader: `display_name` is a property of the `User` model. Can you do `row.display_user`? Otherwise the property just does what you did here.

chris48s: No. I'm confused about why we can't access this property. I tried a few other query styles and I get the same result:

```py
user_table = model.meta.metadata.tables['user']
user = user_table.select().execute().fetchone()

print(user.display_name)     # AttributeError: Could not locate column in row for column 'display_name'
print(user['display_name'])  # NoSuchColumnError: "Could not locate column in row for column 'display_name'"
```

```py
user_table = model.meta.metadata.tables['user']
user = model.Session.execute(model.Session.query(model.User)).fetchone()

print(user.display_name)     # AttributeError: Could not locate column in row for column 'display_name'
print(user['display_name'])  # NoSuchColumnError: "Could not locate column in row for column 'display_name'"
```

(one of the things I'm struggling to get to grips with in SQLAlchemy is it seems to give you many subtly different ways of doing the same thing and I'm not really sure when to use each one)

I must be missing something. I'm just going to merge this and not going to get too bogged down in it for now, but I'd like to try and understand it :thinking:

PR #285, opened 2020-04-24
Enable ckanext-collaborators, allow users to request access to a dataset

### LINKED ISSUE #276: Enable ckanext-collaborators (Dataset level permissions)
Install [ckanext-collaborators](https://github.com/okfn/ckanext-collaborators) and enable it so we can test it works as expected. It just requires creating some tables on startup.

This extension allows Data Container admins to add "Collaborators" to datasets. These can have two roles:
* Members (can download private datasets)
* Editors (can edit datasets)

The UI is in the dataset page > Manage > Collaborators

I think all should work as expected as we have used it in another project with similar authorization requirements but worth playing around.

### LINKED ISSUE #277: Request access to a dataset
Goal: Allow users that can't download a dataset to notify the Data Container Admins so they can add them as collaborators (or not)
TODO: Can users request access to edit datasets?

* Users that are not allowed to download a dataset resources should see a "Request access" button (in the action buttons section, where the "Manage" appears)
* Clicking opens a popup that allows the user to provide a message, like the ones in the curation interface:
Screenshot_2020-04-15 Urban Profiling Thessaloniki Greece - RIDL

* This submits the request to a custom endpoint that receives (via URL or form fields): dataset id, user id and message.
* This endpoint calls a function that constructs an email with the relevant details (user name, link to dataset etc) and sends it to the Admins of the Data Container the dataset belongs to.

For a first iteration we could just put a link to the dataset collaborators page on the email so Admins can add the user if the choose to.
Eventually though it would be nice to list all pending requests in the dedicated Dashboard page for pending requests (in a new "Access to Datasets" section):

Screenshot_2020-04-15 Manage - - Users - RIDL(1)

Approving would directly create a collaborator using the API.

To do this though we would need to persist the requests somehow in the database so we can retrieve them on this page. Given that we also need to implement requesting access to data containers (TODO issue #) it might make sense to create a dedicated table where we can store all requests plus an action to retrieve them.

We can include a link

chris48s:
Closes #276
Closes #277
Refs https://github.com/okfn/docker-ckan-unhcr/pull/23
Refs https://github.com/okfn/docker-ckan-unhcr/pull/25

I'll mark up the diff with a few more specific comments/questions inline, but the main thing I haven't addressed is preventing users who are collaborators on a dataset with write access to the metadata from changing the container. It would be useful to get some guidance on how to go about this.

As the PR stands at the moment, I've just left it so that the default behaviour is that if you edit dataset metadata as a collaborator with edit access you will automatically try and move the dataset out of its current container which is not a safe behaviour.


ckanext/unhcr/controllers/extended_package.py
```diff
@@ -180,3 +181,49 @@ def publish_microdata(self, id):
...
+            return toolkit.abort(403, 'Not Authorized')
+
+        if toolkit.h.can_download(dataset):
+            toolkit.h.flash_notice(
+                'You already have access to download resources from {}'.format(
+                    dataset['title']
+                )
+            )
+            return toolkit.redirect_to('dataset_read', id=dataset['id'])
+
```
chris48s, line 208: TODO: For the next iteration of this feature, we will want to store the request in a table here. I haven't started looking at this yet. I think its better left to another PR as this one is already quite large.

ckanext/unhcr/controllers/extended_package.py
```diff
@@ -180,3 +181,49 @@ def publish_microdata(self, id):
...
+        org_admins = mailer.get_request_access_email_recipients(dataset)
+        if not org_admins:
+            pass
```
chris48s, line 211: I'm not really sure what we should do if we end up here. Should we throw an error? Fall back to emailing sysadmins?

ckanext/unhcr/mailer.py
```diff
@@ -199,3 +199,42 @@ def get_summary_email_recipients():
...
+    context['dataset'] = package_dict
+    context['requesting_user'] = requesting_user_dict
+    context['message'] = message
+    context['collaborators_url'] = toolkit.url_for(
+        'collaborators.new',
+        dataset_id=package_dict['id'],
+        qualified=True,
+    )
+    context['h'] = toolkit.h
+    return render_jinja2('emails/collaboration/request_access.html', context)
```
chris48s, line 240: As well as emailing the container admins when a user wants access, there's also an inverse process of emailing the user to let them know it has been granted (or refused).

Emailing about a refused request will have to wait until we're recording the access requests in a DB, so lets ignore that for now.

It looks like the collaborators plugin already has email templates for notifying the user when they are added to a dataset as a collaborator: https://github.com/okfn/ckanext-collaborators/tree/master/ckanext/collaborators/templates/emails
Do we need to customise those to have the same visual identity as the other RIDL emails (and if so, how) or do we just leave those as they are?

amercader: > Do we need to customise those to have the same visual identity as the other RIDL emails (and if so, how) or do we just leave those as they are?

Yes, I think that's worth doing. You can override the collaborators emails just by adding the same templates on ckanext-unhcr and load it first on `ckan.plugins`. TBH the collaborators ones should be namespaced (eg in `/ckanext/collaborators/templates/collaborators/emails`) to avoid confusion, so you can send a PR there.

ckanext/unhcr/helpers.py
```diff
@@ -778,3 +747,11 @@ def get_sysadmins():
 
 def get_ridl_version():
     return __VERSION__
+
+
+_paragraph_re = re.compile(r'(?:\r\n|\r(?!\n)|\n){2,}')
+
+def nl_to_br(text):
```
amercader, line 754: Can we not use `h.render_markdown()` instead of this? It will have the same effect unless I'm missing something, I will save us this helper and regex

ckanext/unhcr/controllers/extended_package.py
```diff
@@ -180,3 +181,49 @@ def publish_microdata(self, id):
             message = 'Dataset "%s" publishing to the Microdata library is not completed for the following reason: "%s"'
             toolkit.h.flash_error(message % (dataset['title'], error))
         toolkit.redirect_to('dataset_edit', id=dataset['name'])
+
+    def request_access(self, id):
+        if toolkit.request.method != 'POST':
```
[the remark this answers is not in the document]
chris48s: bbc252d

ckanext/unhcr/controllers/extended_package.py
chris48s: 475ad81

ckanext/unhcr/mailer.py
chris48s: For the moment, I have just put the overridden templates in the locations used by `ckanext-collaborators@0.0.4` in 392fe78 but I will have a look at:
- Submitting a PR to the https://github.com/okfn/ckanext-collaborators repo to move the templates
- We can review that and tag a release
- Update ckanext-collaborators in the https://github.com/okfn/docker-ckan-unhcr repo
- Move the overridden templates here

We can either leave this open while I do it, or I can do it as a follow-up.

ckanext/unhcr/helpers.py
chris48s: I did try it with `render_markdown()`, but that doesn't retain single line breaks. For example, it squashes the input text

```
Hello
Can I have access to this dataset so I can do things and stuff with it
Regards,
Chris
```

into

```
<p>Hello Can I have access to this dataset so I can do things and stuff with it Regards, Chris</p>
```

That is not the end of the world, but also I don't think we're losing much by adding a template helper here either.

chris48s:
> I don't think it's safe to deploy it as it is now

Agreed.

> One quick way to implement it would be to customize the owner_org form field so if the logged in user is not a member of the org (ie it's a collaborator), the owner_org field is hidden and with the current value set. What do you think?
Later on we can implement a validator to ensure this can not be bypassed via the API.

I did try this, but if I replace the `<select>` with `<input id="field-organizations" name="owner_org" type="hidden" value="{{ data.organization.id }}" />` and submit the form, it throws:

```
Data Container: You cannot add a dataset to this data container
```

That error is getting thrown in `owner_org_validator()`.

It seems like to have a collaborator with edit access, we need to deal with this on the validation side too, so I've had a go at writing a custom `owner_org_validator()` which allows a dataset collaborator to save a dataset back to its existing owner org, but not to move it to another org. I think this is OK in this situation, but I'm not sure where else `owner_org_validator()` is called so its difficult to know the full impact.

This isn't finished: I've got some failing tests and I haven't written any tests for this implementation, but given I'm a bit uncertain about it, do you want to have a look at 33840ea . If you think I'm completely on the wrong track with this, I won't put more time into it. If you think its heading in the right direction I will work out how to finish it.

amercader:
This sounds like a great approach @chris48s. `owner_org_validator` is only called when creating or updating a dataset and that's exactly the place where we should handle this kind of logic.

ckanext/unhcr/helpers.py
amercader: OK, let's add it then

ckanext/unhcr/controllers/extended_package.py
chris48s: I've moved the stuff from https://github.com/okfn/ckanext-unhcr/issues/277 that isn't done in this PR to https://github.com/okfn/ckanext-unhcr/issues/290 so this PR can just close #277

chris48s:
I think this one is good  for a final check now

ckanext/unhcr/mailer.py
chris48s: I opened some PRs on the ckanext-collaborators repo
https://github.com/okfn/ckanext-collaborators/pulls/chris48s

It may be useful to think about incorporating some of the changes we've made there and in 33840ea / 9bc2b55 / 057dd1c in this PR to https://github.com/ckan/ckan/pull/5346

ckanext/unhcr/templates/scheming/form_snippets/owner_org.html
```diff
@@ -19,12 +19,21 @@
...
   {% set org_required=not h.check_config_permission('create_unowned_dataset') or h.scheming_field_required(field) %}
 
-  {% snippet "scheming/form_snippets/_organization_select.html",
-    field=field,
-    data=data,
-    errors=errors,
-    organizations_available=organizations_available,
-    org_required=org_required,
-    organization_option_tag=organization_option_tag %}
+  {% if data.original_dataset or h.user_in_org_or_group(data.organization.id) %}
```
amercader, line 22: This will crash on the new dataset form because `data.organization` is None. We shouldn't use `data.organization` anyway as this is added in Solr for convenience. The actual org id field is `data.owner_org` (the one this snippet is about).
It would be great to have a small test that checks that the new dataset form renders fine to catch these issues early.
Also if you are creating a new dataset you want to see the orgs select and the check above won't allow it. So you need something like:

```diff
diff --git a/ckanext/unhcr/templates/scheming/form_snippets/owner_org.html b/ckanext/unhcr/templates/scheming/form_snippets/owner_org.html
index db90515..df33f98 100644
--- a/ckanext/unhcr/templates/scheming/form_snippets/owner_org.html
+++ b/ckanext/unhcr/templates/scheming/form_snippets/owner_org.html
@@ -19,7 +19,7 @@
   {% set organizations_available=h.organizations_available('create_dataset') | rejectattr('id', 'equalto', deposit.id) %}
   {% set org_required=not h.check_config_permission('create_unowned_dataset') or h.scheming_field_required(field) %}
 
-  {% if data.original_dataset or h.user_in_org_or_group(data.organization.id) %}
+  {% if data.original_dataset or not data.id or (data.owner_org and h.user_in_org_or_group(data.owner_org)) %}
     {% snippet "scheming/form_snippets/_organization_select.html",
       field=field,
       data=data,
@@ -32,7 +32,7 @@
       id="field-organizations"
       name="owner_org"
       type="hidden"
-      value="{{ data.organization.id }}"
+      value="{{ data.owner_org }}"
     />
   {% endif %}

```
Why do we need the `if data.original_dataset` bit?

ckanext/unhcr/validators.py
```diff
@@ -212,3 +212,19 @@ def activity_type_exists(activity_type):
...
         return activity_type
     return validators.activity_type_exists(activity_type)
+
+
+# Collaborators
+
+def owner_org_validator(key, data, errors, context):
+    user = context.get('user')
+    package = context.get('package')
+    action = toolkit.get_action('dataset_collaborator_list_for_user')
```
amercader, line 222: This will raise an exception if the action is not found (ie collaborators plugin not loaded). If you want to check this you will need to catch ImportError when importing `ckanext.collaborators`.

ckanext/unhcr/validators.py
```diff
@@ -212,3 +212,19 @@ def activity_type_exists(activity_type):
...
+    if user and action and package:
+        datasets = action(context, {'id': user})
+        if package.id in [d['dataset_id'] for d in datasets]:
+            if data.get(key) == package.owner_org:
+                return
```
amercader, line 227: We still want the value to get validated using the core logic

```suggestion
                return validators.owner_org_validator(key, data, errors, context)
```

ckanext/unhcr/validators.py
```diff
@@ -212,3 +212,19 @@ def activity_type_exists(activity_type):
...
     return validators.activity_type_exists(activity_type)
+
+
+# Collaborators
+
+def owner_org_validator(key, data, errors, context):
+    user = context.get('user')
+    package = context.get('package')
+    action = toolkit.get_action('dataset_collaborator_list_for_user')
+    if user and action and package:
```
amercader, line 223: You can add a sysadmin as a collaborator on a dataset. I know it doesn't make much sense, but if it happens they would not be able to change orgs. So let's skip this block if the user is a sysadmin. You'll need `model.User.get(user).sysadmin` for this

amercader: And I guess that technically if someone is an Org Editor in both the current Data Container and another one, and is also added as collaborator on the dataset they won't be able to move it but this is perhaps a bit too far fetched. I guess that we could allow moving it if you are an admin / editor of the current organization.

Let's do the sysadmin case only for now.

ckanext/unhcr/mailer.py
```diff
@@ -199,3 +199,52 @@ def get_summary_email_recipients():
...
+        recipients = [
+            user for user in org["users"]
+            if user["capacity"] == "admin" and user["name"] != default_user["name"]
+        ]
+    except toolkit.ObjectNotFound:
+        recipients = []
+
+    # if we couldn't find any org admins, fall back to sysadmins
+    if not recipients:
+        all_users = toolkit.get_action("user_list")(context, {})
```
amercader, line 222: Use the `get_sysadmins()` helper from this extension, it's much more efficient

ckanext/unhcr/templates/emails/collaboration/request_access.html
```diff
@@ -0,0 +1,33 @@
...
+  {% call email.paragraph() %}
+    {{ requesting_user.fullname or requesting_user.name }} supplied the following message:
+  {% endcall %}
+
+  {% call email.paragraph() %}
+    {{ h.nl_to_br(message) }}
+  {% endcall %}
+
+  {% call email.paragraph() %}
+    To add the user as a collaborator, visit the following page (logged in as a site administrator):
```
amercader, line 26: Data Container admins also can add collaborators (in fact they are the main target of this email) so "site administrator" might be confusing. What about "logged in with your administrator account"

ckanext/unhcr/templates/package/snippets/collaboration_modals.html
```diff
@@ -0,0 +1,33 @@
+<div class="collaboration collaboration_modals-modals">
+
+{% if pkg.resources|length > 0 and not h.can_download(pkg) %}
```
amercader, line 3: The message input looks wonky for me:
Screenshot_2020-05-07 Test dataset 2 - RIDL

This is how the curation ones look:
Screenshot_2020-05-07 YOO DEpo - RIDL

Also can we say "Please explain" :)

chris48s: ea74002

ckanext/unhcr/templates/emails/collaboration/request_access.html
chris48s: adf04bd

ckanext/unhcr/mailer.py
chris48s: e0ee79c

ckanext/unhcr/validators.py
chris48s: Done in 327e32b

We cna guard from it here anyway, but given it doesn't really make sense to explicitly add a user to a dataset as a collaborator if they already have access to it (via being a sysadmin or group membership), do you think it makes sense to think about preventing this via validation in the collaborators plugin/feature?

ckanext/unhcr/validators.py
chris48s: If the user is a collaborator on a dataset and we fall back to `validators.owner_org_validator()` here, we will then check if the user is in the org and throw `Invalid(_('You cannot add a dataset to this organization'))`. One of the reasons for implementing a custom validator here is to bypass that check. I've proposed a solution in 739cdde although I'm not mad keen on relying on the exception message string.

ckanext/unhcr/validators.py
chris48s: It will.. but why is that a problem? This is a plugin that essentially has one deployment environment and we're adding collaborators in https://github.com/okfn/docker-ckan-unhcr/pull/25 Do we need to handle that case? Shouldn't it be safe within this codebase to assume that everything in https://github.com/okfn/docker-ckan-unhcr/blob/8f437044f638ba862fffeebb323dd22885d9bc9a/ckan/Dockerfile#L17-L35 is available to us? For example, elsewhere in this module we're just calling `from ckanext.scheming import ...` on the assumption that `ckanext-scheming` is installed and if its not, we'll throw an unhandled `ImportError`.

ckanext/unhcr/templates/scheming/form_snippets/owner_org.html
chris48s: Done in d4460b9

> Why do we need the if `data.original_dataset` bit?

`data.original_dataset` is there because this form snippet also gets used on the "copy dataset" form. If we're copying a dataset, we want to show the standard `organization_select` form (i.e: the user can copy a dataset to any organization they are a member of). `data.original_dataset` if we are copying.

ckanext/unhcr/validators.py
amercader: Yes, you can assume collaborators is loaded if using unhcr, I just thought you were checking if it was loaded as you were doing `if action and ...`.

ckanext/unhcr/validators.py
amercader: Probably, let's improve it in the future

chris48s:
Thanks for reviewing. Its good to get this one merged - it turned into a bit of a monster

Issue #254, opened 2019-12-12
Deploy the project to AWS (phase 1)


roll:
# Overview

An issue to track the progress of the AWS migration

# Tasks

## To deploy staging

- [x] decide on WAF usage - will be handle by UNHCR
- [x] extract items from the meeting
- [x] use `npm-check-updates` for dependency upgrades
- [x] set the retention period for log groups
- [x] add explicit no public access to storage (S3)
- [x] add explicit no public access to database (RDS)
- [x] review the S3 buckets permissions (set not public)
- [x] review the RDS setup (see - https://github.com/okfn/ckanext-unhcr/issues/245)
- [x] review how we create the `datastore_ro` user in `docker-entrypoints` 
- [x] analyze costs and reduce it!
- [x] add notifications on a failing build
- [x] move approval notifications to SNS? - just re-use the email list
- [x] review the SSH security
- [x] new resource page is broken after cloudstorage controller update / unhcr@1.0.7 rebase
- [x] downloading/previewing resources doesn't work (permission denied)
- [x] get a review from Adria (grep for `TODO` in the codebase)
- [x] get access to the staging accounts (blocked by UNHCR)
- [x] update CIDR block for VPC to `10.0.0.0/16`
- [x] deploy staging

## To deploy production

- [x] setup http -> https redirect (fix 2 load balancers situations)
- [x] decide on single/separate accounts for staging/production
- [x] finish and review the DNS configuration (we need a certificate for `ridl/ridl-uat.unhcr.org`) - waiting for UNHCR
- [x] refresh secrets
- [x] we might need to use CloudFront to deal with permissions for ckanext-cloudstorage
- [x] review usage of the local storage (rebase on the `ckanext-cloudstorage` fork to support orgs)
- [x] review and improve the documentation
- [x] deploy production

amercader:
@roll this looks like a great plan to me!

roll:
@chris48s 
I would move all these from for the future part to a separate issue to make the thing less massive a little bit

chris48s:
- [ ] Continue to log downloads when we migrate to cloudstorage (see #269 )
- [ ] Copy existing database content from existing Prod/UAT sites to AWS

chris48s:
Cheated and moved a few bits to phase 2 :see_no_evil:

PR #289, opened 2020-04-28
log download activity when using cloudstorage

chris48s:
Refs #269
Refs #254

This _should_ work, but I've not actually tested it using the cloudstorage backend yet

### DIFF
--- a/bin/travis-build.bash
+++ b/bin/travis-build.bash
@@ -68,6 +68,11 @@ git checkout 0.0.5
 python setup.py develop
 cd -
 
+git clone https://github.com/TkTech/ckanext-cloudstorage
+cd ckanext-cloudstorage
+python setup.py develop
+cd -
+
 echo "Installing ckanext-unhcr and its requirements..."
 python setup.py develop
 pip install --upgrade -r requirements.txt
--- a/ckanext/unhcr/plugin.py
+++ b/ckanext/unhcr/plugin.py
@@ -3,6 +3,7 @@ import os
 import json
 import logging
 
+from ckan.common import config
 import ckan.plugins as plugins
 import ckan.plugins.toolkit as toolkit
 from ckan.lib.plugins import DefaultTranslation
@@ -113,7 +114,7 @@ class UnhcrPlugin(
         _map.connect('/dataset/{id}/resource_copy/{resource_id}', controller=controller, action='resource_copy')
         _map.connect('/dataset/{id}/publish_microdata', controller=controller, action='publish_microdata')
         _map.connect('/dataset/{id}/request_access', controller=controller, action='request_access', conditions={'method': ['POST']})
-        if 'cloudstorage' not in os.environ.get('CKAN__PLUGINS', ''):
+        if 'cloudstorage' not in config['ckan.plugins']:
             _map.connect('/dataset/{id}/resource/{resource_id}/download', controller=controller, action='resource_download')
             _map.connect('/dataset/{id}/resource/{resource_id}/download/{filename}', controller=controller, action='resource_download')
         else:
--- a/ckanext/unhcr/plugin.py
+++ b/ckanext/unhcr/plugin.py
@@ -117,7 +117,7 @@ class UnhcrPlugin(
             _map.connect('/dataset/{id}/resource/{resource_id}/download', controller=controller, action='resource_download')
             _map.connect('/dataset/{id}/resource/{resource_id}/download/{filename}', controller=controller, action='resource_download')
         else:
-            controller='ckanext.cloudstorage.controller:StorageController'
+            controller='ckanext.unhcr.controllers.extended_storage:ExtendedStorageController'
             _map.connect('/dataset/{id}/resource/{resource_id}/download', controller=controller, action='resource_download')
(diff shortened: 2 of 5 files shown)


chris48s:
I tried deploying this to UAT to test it with S3 and it.. doesn't work :angry:  The issue is:

With the file storage, `resource_download` returns a response, so we can do:

```py
resp = super(BaseController, self).resource_download(
    id, resource_id, filename=filename
)
self._log_download_activity(context, resource_id)
return resp
```

With the cloudstorage controller, `resource_download` issues a redirect, so if we try to do

```py
resp = super(BaseController, self).resource_download(
    id, resource_id, filename=filename
)
self._log_download_activity(context, resource_id)
return resp
```

`resource_download` issues a redirect and we never execute `self._log_download_activity(context, resource_id)`. We've already issued a response.

The things I can think of to do here are:

1. Log the download first when we're using cloudstorage, which is simpler but will probably generate some false-positives (as you correctly pointed out in https://github.com/okfn/ckanext-unhcr/pull/280#discussion_r412041624 )
2. Completely override `resource_download`, copy and paste the entire `resource_download` method from https://github.com/TkTech/ckanext-cloudstorage/blob/master/ckanext/cloudstorage/controller.py into our plugin and then we can call our `log_download_activity()` before `h.redirect_to()`.
3. Fork (or see if we can contribute to) `cloudstorage` to make this easier

@amercader - what are your thoughts on which way to go on this, or do you have another smarter idea?

chris48s:
I've not tried it yet, but over the weekend I thought of another option, which is that we could probably use `mock.patch()` to inject our own `h.redirect_to()` and log the download activity like that.

amercader:
@chris48s I think I'd go for the simple approach and choose option 1. By the point we log the download we'd have checked if the resource exists and the authorization which would be the most common user errors. Any further exceptions would come from cloudstorage itself, and we hopefully will pick these and fix them so I'd say the danger of false positives is low.

CKAN 2.9 might introduce signals (https://github.com/ckan/ckan/pull/5359) which will allow a much cleaner implementation of downloads tracking

bin/travis-build.bash
```diff
@@ -62,6 +62,11 @@ cd ckanext-hierarchy
 python setup.py develop
 cd -
 
+git clone https://github.com/TkTech/ckanext-cloudstorage
+cd ckanext-cloudstorage
+python setup.py develop
+cd -
```
chris48s, line 68: This is a bit annoying. The problem I'm trying to solve here is:

If we try to call `from ckanext.cloudstorage.controller import StorageController` and `ckanext-cloudstorage` isn't installed, we'll throw `ImportError`. In general, this is useful behaviour. If I'm trying to extend `ckanext.cloudstorage.controller.StorageController` and `ckanext-cloudstorage` isn't installed then that's exactly how we want to be told about it - we shouldn't try to fall back to some other behaviour.

This is fine, except that when nose tries to read in everything to measure statement coverage at the end of the test run, trying to parse `extended_storage.py` throws an import error. I think installing `ckanext-cloudstorage` to silence the error but not registering `cloudstorage` in `CKAN__PLUGINS` (so we still use the local storage backend under test - that's what we look at when we set up the routes in `plugin.py`) is probably a reasonable workaround for this.

Another approach would be to explicitly ignore that file for test coverage, but I think it is useful to retain the information that we don't have tests covering this behaviour.

Open to another suggestion though.. I think whatever we do will be a bit of a fiddle

chris48s:
This is ready for review now.

bin/travis-build.bash
amercader: Installing and not enabling sounds fine

ckanext/unhcr/plugin.py
```diff
@@ -116,7 +116,7 @@ def before_map(self, _map):
             _map.connect('/dataset/{id}/resource/{resource_id}/download', controller=controller, action='resource_download')
```
amercader, line 116: Can we combine the logic for local and cloudstorage downloads in a single endpoint? We can have a single `resource_download` function and then do the check to see if cloudstorage is loaded there. 
The existing check (`if 'cloudstorage' not in os.environ.get('CKAN__PLUGINS', '')` is not reliable, is best to do `'cloudstorage' in config['ckan.plugins']`

chris48s: I've made the change you've suggested and submitted it as a different PR in #297

Having done that implementation too, I actually think the approach in this PR is slightly more "correct" due to https://github.com/okfn/ckanext-unhcr/pull/297#discussion_r424989752 . What do you think?

amercader: It's not a big deal and both approaches have drawbacks so let's go with the PR approach.
I was thinking something along these lines for the other approach:

```python

class ExtendedStorage(CorePackageController):

    def resource_download():

         if cloudstorage:
               return CloudStorageController().resource_download()

        return super(ExtendedStorage).resource_download()

```

But let's keep the two files like in this PR

chris48s:
OK, cool. I've rebased this to resolve the merge conflicts. I will have a look at getting this deployed to the UAT instance on AWS today or Monday so we start logging the downloads there.
I'll also rebase #294 so that can be reviewed.

Issue #284, opened 2020-04-23
New admin activity stream, showing curation activities and downloads


amercader:
Datasets currently have a separated tab called "Curation activity" that displays activities related to the curation process. This is only shown to users who can edit the dataset. (ie org editors and admins).

![Screenshot_2020-04-23 MDM Metadata - RIDL](https://user-images.githubusercontent.com/200230/80084810-a976b700-8557-11ea-9f70-e30c5393d065.png)
![Screenshot_2020-04-23 Urban Profiling Thessaloniki Greece - RIDL](https://user-images.githubusercontent.com/200230/80084818-ada2d480-8557-11ea-937f-66533b7058c6.png)

The goal is to add new internal only activities like downloads into this tab to avoid adding more. We can rename it to "Internal activities" for now.

* The tab is added in the [`read_base.html`](https://github.com/okfn/ckanext-unhcr/blob/fdebd7d2dfeed6836daa20fe4716c7ca466f6292/ckanext/unhcr/templates/package/read_base.html#L17) template
* The link calls a custom action in the [`deposited_dataset.py`](https://github.com/okfn/ckanext-unhcr/blob/fdebd7d2dfeed6836daa20fe4716c7ca466f6292/ckanext/unhcr/controllers/deposited_dataset.py#L337) controller. We should probably move it to the `extended_package.py` one as activities won't be curation specific from now on.
* This calls our custom `package_activity_list_html` action with the `get_curation_activities` param in the context set to True. All `*_activity_*` actions accept this parameter to return either the normal activities or the internal ones. We'll need to update them all to also include download ones.
* TODO code will be slightly different depending on who can see download activities:

  - If editors can (same as current curation activities) we just rename the param and include the download activities to the current curation ones.
 - If only admins can see downloads then we will need to filter the internal activities returned depending on whether the user is an admin or not of the data container (curation only if it's an editor, curation+ download if it's an admin). I'd probably determine this at the controller level and pass  `get_curation_activities` and / or `get_admin_activities` depending on the user role, and then the actions would act accordingly.
* In any case we will need to add a custom render function for the `resource download` activity. This is done via a lovely monkey-patch in [here](https://github.com/okfn/ckanext-unhcr/blob/fdebd7d2dfeed6836daa20fe4716c7ca466f6292/ckanext/unhcr/plugin.py#L46)

chris48s:
[...]

PR #294, opened 2020-05-12
log 'download resource' activity with dataset_id not resource_id

chris48s:
refs #284

As discussed in https://github.com/okfn/ckanext-unhcr/issues/284#issuecomment-626718358 this switches us to logging the 'download resource' activity against the `dataset_id` instead of the `resource_id` so it is easier to query.

The downside of this is that we've already logged a bunch of download resource' activities in UAT against `resource_id`s and all that data will essentially become 'orphaned'. I haven't included any data migration in this PR to fix or delete those records. I think this is probably fine to just lose that data as that code is only deployed in UAT, but let me know if I need to do something else there.

### DIFF
--- a/ckanext/unhcr/actions.py
+++ b/ckanext/unhcr/actions.py
@@ -256,6 +256,7 @@ def package_activity_list(context, data_dict):
     get_curation_activities = toolkit.asbool(
         data_dict.get('get_curation_activities'))
     full_list = get_core.package_activity_list(context, data_dict)
+    full_list = [a for a in full_list if a["activity_type"] != "download resource"]
     curation_activities = [
         a for a in full_list if 'curation_activity' in a.get('data', {})]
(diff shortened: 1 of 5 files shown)


ckanext/unhcr/actions.py
```diff
@@ -256,6 +256,7 @@ def package_activity_list(context, data_dict):
     get_curation_activities = toolkit.asbool(
         data_dict.get('get_curation_activities'))
     full_list = get_core.package_activity_list(context, data_dict)
+    full_list = [a for a in full_list if a["activity_type"] != "download resource"]
```
chris48s, line 259: For the moment, I'm just filtering these out completely. Once we've got a bit more clarify from the client about how/where to present these I'll come back to this when we continue #284

chris48s:
ready for review

PR #295, opened 2020-05-13
Add metrics for data deposits

### LINKED ISSUE #293: Add metrics for data deposits
Same graph we have for datasets, but for deposits

chris48s:
closes #293

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@ dist: trusty
 language: python
 python:
     - "2.7"
+addons:
+    postgresql: '9.6'
 env:
     - CKANVERSION=2.8
 install:
--- a/ckanext/unhcr/models.py
+++ b/ckanext/unhcr/models.py
@@ -7,6 +7,7 @@ from sqlalchemy import Column, DateTime, Integer
 from sqlalchemy.ext.declarative import declarative_base
 
 from ckan.model.meta import metadata
+import ckan.model as model
 
 
 log = logging.getLogger(__name__)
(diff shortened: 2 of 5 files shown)


ckanext/unhcr/models.py
```diff
@@ -18,13 +19,28 @@ class TimeSeriesMetric(Base):
     __tablename__ = u'time_series_metrics'
 
     timestamp = Column(DateTime, primary_key=True, default=datetime.datetime.utcnow)
-    datasets_count = Column(Integer, nullable=False)
-    containers_count = Column(Integer, nullable=False)
+    datasets_count = Column(Integer)
+    deposits_count = Column(Integer)
+    containers_count = Column(Integer)
```
chris48s, line 24: When I initially added this table I thought we probably didn't want to allow `NULL`.
Now I'm adding a new column I realise that we do need `NULL` for the new column because there's a bunch of rows that already exist where we don't have a value for the new column. I figure now that we need `NULL` as a 'no data' value, we might as well be consistent.

.travis.yml
```diff
@@ -3,6 +3,8 @@ dist: trusty
 language: python
 python:
     - "2.7"
+addons:
+    postgresql: '9.6'
```
chris48s, line 7: The reason I've upgraded the CI environment here is because I've used `ADD COLUMN IF NOT EXISTS` which exists in Postgres 9.6+ and we were using an older version in CI. In the current(old) production/UAT environment we are using Postgres 9.6, so I'll pin to that for the moment. The RDS instances we've provisioned in AWS are running Postgres 11 (woo :tada: ).
Once we're running on Postgres 11 in production, we should look at bumping our dev/test environments so we're in step with the live environment.

amercader: Sounds great

PR #297, opened 2020-05-14
log download activity when using cloudstorage (2)

chris48s:
Alternative approach to #289

### DIFF
--- a/ckanext/unhcr/controllers/extended_storage.py
+++ b/ckanext/unhcr/controllers/extended_storage.py
@@ -1,0 +1,86 @@
+import os
+from ckan import model
+from ckan.common import config
+import ckan.plugins.toolkit as toolkit
+
+
+def get_base_controller():
+    """
+    if we're using ckanext-cloudstorage we want to extend
+    ckanext.cloudstorage.controller:StorageController
+    otherwise, we want to extend
+    ckanext.unhcr.controllers.extended_package:ExtendedPackageController
+    """
+
+    if 'cloudstorage' in config['ckan.plugins']:
+        from ckanext.cloudstorage.controller import StorageController
+        return StorageController
+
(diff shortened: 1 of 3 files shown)


ckanext/unhcr/controllers/extended_storage.py
```diff
@@ -0,0 +1,86 @@
...
+def get_base_controller():
+    """
+    if we're using ckanext-cloudstorage we want to extend
+    ckanext.cloudstorage.controller:StorageController
+    otherwise, we want to extend
+    ckanext.unhcr.controllers.extended_package:ExtendedPackageController
+    """
+
+    if 'cloudstorage' in config['ckan.plugins']:
+        from ckanext.cloudstorage.controller import StorageController
```
chris48s, line 16: One advantage of this approach is that the coverage reports don't break if `ckanext-cloudstorage` is not installed because we don't need to try the import to read the file.

ckanext/unhcr/controllers/extended_storage.py
```diff
@@ -0,0 +1,86 @@
...
+        # Check resource_download access
+        try:
+            toolkit.check_access(u'resource_download', context, {u'id': resource_id})
+        except toolkit.ObjectNotFound:
+            return toolkit.abort(404, toolkit._(u'Resource not found'))
+        except toolkit.NotAuthorized:
+            return toolkit.abort(403, toolkit._(u'Not Authorized to download the resource'))
+
+
+        if 'ExtendedPackageController' in [b.__name__ for b in self.__class__.__bases__]:
```
chris48s, line 70: Having tried this implementation, I actually think this is the wrong way to go. If we need different behaviour depending on which base class we're extending, this is a pretty strong indication that we really want two different classes here.

Issue #286, opened 2020-04-27
Request access to a Data Container


chris48s:
Ability for normal users to request access to a Data Container

This will look quite similar to the work in #277 / #285

It may make sense to look at this with phase 2 of #277 which is storing the requests for access in a table and presenting a front-end to manage them (instead of just triggering emails)

chris48s:
now #308 is merged, I'll get this done next

PR #308, opened 2020-06-02
interface for listing/approving/rejecting access requests

### LINKED ISSUE #290: Improve dataset access requests
#285 implements the MVP for this

Next job:

It would be nice to list all pending requests in the dedicated Dashboard page for pending requests (in a new "Access to Datasets" section):

79341164-a95d3280-7f2b-11ea-8c85-d9607d500b39

Approving would directly create a collaborator using the API.

To do this though we would need to persist the requests somehow in the database so we can retrieve them on this page. Given that we also need to implement requesting access to data containers (TODO issue #) it might make sense to create a dedicated table where we can store all requests plus an action to retrieve them.

chris48s:
Closes #290
Refs #286

I've done the basics relating to #290 and part of #286 in this PR, but there are some tasks I've ignored, either because I'm not sure about the best way forward, or just because this PR is already really long and we can get away with deferring them to another PR. I think there is enough here that it is sensible to review it as-is.

## Stuff I have done

* Requesting access to a dataset stores an `AccessRequest` object in the DB.
* An `AccessRequest` is sufficiently generic that it can be a request for access to either a dataset or a container.
* Container admins have a front-end to view/approve/reject access requests from users.

## Stuff I haven't done

* The backend changes add the ability for us to store/list/approve/reject requests to access a container as well as a dataset, but I haven't added any of the front-end/email templates for a user to request access to a container. This isn't a huge job, but this diff is already too long, so I think I'm going to defer adding that to a second PR.
* As things stand, if we approve an access request, the user gets an email saying they've been added to the container or added to the dataset as a collaborator. If we reject the request, the requesting user doesn't get an email. We could add a some code to send them an email saying "your request was rejected". To be really useful, we maybe also need to give the admin a textbox to include a message explaining _why_ the request was rejected. For the moment I haven't done either. What do you think?
* At the moment, the access requests page only displays requests which are in the 'requested' state. Once they are approved or rejected, they basically just disappear from the front-end (although the DB record still exists). I think its ok to "launch" the feature like this, but maybe we need to think about a history/audit requirement here. Perhaps one approach would be to log an activity when a request is approved/rejected that could be surfaced in an activity stream? I haven't done anything about it either way.

As usual, I will also mark up the diff with some inline comments about specific issues.


ckanext/unhcr/models.py
```diff
@@ -24,13 +26,30 @@ class TimeSeriesMetric(Base):
...
+        Enum('member', 'editor', 'admin', name='access_request_role_enum'),
+        nullable=False,
+    )
+    status = Column(
+        Enum('requested', 'approved', 'rejected', name='access_request_status_enum'),
+        default='requested',
+        nullable=False,
+    )
+    object_type = Column(
+        Enum('dataset', 'container', name='access_request_object_type_enum'),
```
chris48s, line 46: I've used 'dataset' and 'container' here as that's the user-facing terminology but I wonder if I should have used 'package' and 'organization' to match the CKAN internals?

ckanext/unhcr/controllers/extended_package.py
```diff
@@ -178,6 +179,16 @@ def request_access(self, id):
             )
             return toolkit.redirect_to('dataset_read', id=dataset['id'])
 
+        rec = AccessRequest(
+            user_id=toolkit.c.userobj.id,
+            object_id=dataset['id'],
+            object_type='dataset',
+            message=message,
+            role='member',
```
chris48s, line 187: I've hard-coded `role='member'` here, but I've made it flexible enought that we can store other access levels if we want.

ckanext/unhcr/auth.py
```diff
@@ -253,3 +253,9 @@ def unhcr_datastore_search_sql(context, data_dict):
 
 def datasets_validation_report(context, data_dict):
     return {'success': False}
+
+
+# Access Requests
+
+def access_request_list_for_user(context, data_dict):
+    return {'success': True}
```
chris48s, line 261: Do we want to restrict this at all? I've taken a cue from https://github.com/ckan/ckan/blob/ckan-2.8.4/ckan/logic/auth/get.py#L78-L79 here as it seems conceptually similar.

ckanext/unhcr/plugin.py
```diff
@@ -366,12 +366,15 @@ def get_auth_functions(self):
         functions['unhcr_datastore_search'] = auth.unhcr_datastore_search
         functions['unhcr_datastore_search_sql'] = auth.unhcr_datastore_search_sql
         functions['datasets_validation_report'] = actions.datasets_validation_report
```
chris48s, line 368: This isn't strictly related to this PR except that I noticed it while I was working on it.
Is this supposed to be `auth.datasets_validation_report`, or is it `actions.datasets_validation_report` for _good reasons_ ?

chris48s:
This is good for review, but lower priority than all the other stuff that's open

screenshot with a bunch of requests to approve/reject:

Screenshot_2020-06-09 Manage - - Users - RIDL

---

with nothing to approve/reject:

Screenshot at 2020-06-09 16-33-49

ckanext/unhcr/plugin.py
chris48s: Github marked this as "outdated" after I rebased to resolve a bunch of merge conflicts, but its still a relevant question

ckanext/unhcr/models.py
amercader: I like to keep terms inside each layer (db in this case) consistent so +1 for package / org. Also if we ever want to abstract this in a separate extension it will be less ridl specific

ckanext/unhcr/actions.py
```diff
@@ -452,3 +454,151 @@ def datasets_validation_report(context, data_dict):
...
+        context, {"id": user_id, "permission": "admin"}
+    )
+    containers = [o["id"] for o in organizations]
+    if not containers:
+        return []
+
+    fq = "owner_org:({ids})".format(ids=" OR ".join(containers))
+    data_dict = {"q": "*:*", "fq": fq, "include_private": True}
+    packages = toolkit.get_action("package_search")(context, data_dict)
+    datasets = [p["id"] for p in packages["results"]]
```
amercader, line 542: Can we do this filter at the DB level, with something like `Package.owner_org in (conatiners)` and `Group.id in (containers)`?

ckanext/unhcr/auth.py
amercader: Yes, I think we should restrict it. One thing is to publicize a user belonging to a container, but the requests should be private, specially pending and rejected. To keep it simple we can do sysdmins and team admins only for now.

ckanext/unhcr/templates/user/dashboard_requests.html
```diff
@@ -44,7 +47,36 @@ <h3 class="dataset-heading">
     </ul>
   {% else %}
```
amercader, line 48: No need to add the object type on the `No outstanding requests for XXX`  messages, it looks a bit redundant. Also  let's consolidate on lower case "requests"

Screenshot_2020-07-01 Manage - - Users - RIDL

ckanext/unhcr/plugin.py
amercader: It should definitely be `auth.datasets_validation_report`. Not sure how it ended up like this, probably too much copy / pasting

ckanext/unhcr/models.py
chris48s: :+1: done in 1fc6222

ckanext/unhcr/auth.py
chris48s: Cool. I've done this in a5d1f1a
I've only bothered to explicitly check for container admin because if you are a sysadmin, all auth checks return true. Is this conventional?
Also, the other cue I took from `organization_list_for_user` was that `context['user']` can be different from `data_dict['user_id']` but I think given your feedback this also probably doesn't make sense here, so I've also simplified this so that the action gets the relevant requests for `context['user']`. That probably makes more sense in this situation, right?

ckanext/unhcr/controllers/extended_user.py
```diff
@@ -11,19 +11,38 @@ class ExtendedUserController(UserController):
...
         context = {'model': model, 'user': toolkit.c.user}
         self._custom_setup_template_variables(context)
 
-        # Get requests
         try:
-            requests = helpers.get_pending_requests(all_fields=True)
-        except toolkit.NotAuthorized:
-            message = 'Not authorized to see pending requests'
-            return toolkit.abort(403, message)
+            new_container_requests = toolkit.get_action('pending_requests_list')(
```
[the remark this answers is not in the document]
chris48s: :+1:  8bd8fbf

ckanext/unhcr/templates/user/dashboard_requests.html
chris48s: :+1: a71e077

ckanext/unhcr/actions.py
chris48s: Not done this one yet.. I'm not necessarily against pulling the auth check out, especially if you think we just want to apply slightly different/stricter auth checks but I don't think I agree it would simplify things here compared to using the existing auth functions for these operations. I've not tested the code below, but if we add a new auth function for `access_request_update` using the logic you've suggested, its going to look _roughly_ like this:

```py
from ckan.authz import has_user_permission_for_group_or_org

def access_request_update(context, data_dict):
    user = context.get('user')
    request_id = toolkit.get_or_bust(data_dict, "id")
    request = model.Session.query(AccessRequest).get(request_id)

    if request.object_type == 'package':
        package = toolkit.get_action('package_show')(
            context, {'id': request.object_id}
        )
        org_id = package['owner_org']
    elif request.object_type == 'organization':
        org_id = request.object_id
    else:
        raise toolkit.Invalid("Unknown Object Type")

    return {
        'success': has_user_permission_for_group_or_org(
            org_id, user, 'admin'
        )
    }

```

and it only allows us to remove 2 lines from this function:

`toolkit.check_access('dataset_collaborator_create', context, data_dict)` and `toolkit.check_access('organization_member_create', context, data_dict)`

Everything else here stays the same. I think that makes it more complicated, not less.

ckanext/unhcr/actions.py
```diff
@@ -452,3 +454,151 @@ def datasets_validation_report(context, data_dict):
...
+def access_request_update(context, data_dict):
+    """
+    Approve or reject a request for access to a container or dataset
+
+    :param id: access request id
+    :type id: string
+    :param status: new status value (approved, rejected)
+    :type status: string
+    """
+    request_id = toolkit.get_or_bust(data_dict, "id")
```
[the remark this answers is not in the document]
chris48s: Not done this one yet.. I'm not necessarily against pulling the auth check out, especially if you think we just want to apply slightly different/stricter auth checks but I don't think I agree it would simplify things here compared to using the existing auth functions for these operations. I've not tested the code below, but if we add a new auth function for `access_request_update` using the logic you've suggested, its going to look _roughly_ like this:

```py
from ckan.authz import has_user_permission_for_group_or_org

def access_request_update(context, data_dict):
    user = context.get('user')
    request_id = toolkit.get_or_bust(data_dict, "id")
    request = model.Session.query(AccessRequest).get(request_id)

    if request.object_type == 'package':
        package = toolkit.get_action('package_show')(
            context, {'id': request.object_id}
        )
        org_id = package['owner_org']
    elif request.object_type == 'organization':
        org_id = request.object_id
    else:
        raise toolkit.Invalid("Unknown Object Type")

    return {
        'success': has_user_permission_for_group_or_org(
            org_id, user, 'admin'
        )
    }

```

and it only allows us to remove 2 lines from this function:

`toolkit.check_access('dataset_collaborator_create', context, data_dict)` and `toolkit.check_access('organization_member_create', context, data_dict)`

Everything else here stays the same. I think that makes it more complicated, not less.

chris48s:
I've pushed some more commits to this branch today but it is still WIP. I need to do a couple more bits to it next week before the next round of comments.

TODO:
* Have a look at https://github.com/okfn/ckanext-unhcr/pull/308#discussion_r448315952
* Implement the "Your request has been rejected" emails (not started this yet)
* Final check/test

Couples of notes:

* I rebased this PR on to master to resolve conflicts. 83212a9 was the last commit you looked at. 1fc6222 onwards are responding to review comments.
* In order to switch the terminology to 'organization' and 'package' I updated the DB init script, but there isn't a clean migration path if you've already applied the previous version to your working copy. If you already applied the previous version to your working copy, my suggestion is to:
    * add `AccessRequest.__table__.drop()` at the start of `create_tables()`,
    * restart local containers to re-init
    * remove `AccessRequest.__table__.drop()`.
    You might need to do first test run with `--reset-db`.

ckanext/unhcr/plugin.py
chris48s: Sorted. Done in https://github.com/okfn/ckanext-unhcr/pull/370

ckanext/unhcr/actions.py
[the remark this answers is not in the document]
amercader: Calling just one time `check_access` is a good simplification as we know early in the function that the user is authorized. Besides CKAN core [enforces](https://github.com/ckan/ckan/blob/master/ckan/logic/__init__.py#L480:L483) the good practice of actions having a separate auth function.

ckanext/unhcr/actions.py
chris48s: :+1:  done in 7f4956c

ckanext/unhcr/actions.py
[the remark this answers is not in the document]
chris48s: I've implemented this roughly as proposed above in 7f4956c, but I have some follow-up questions about this topic:

- Instead of querying the `AccessRequest` table twice (once in the action and once in the auth function), I've passed the `AccessRequest` object to the auth function in the `context` object, but I'm not 100% sure if this is correct:
    - Is it ok for `auth.access_request_update` and `actions.access_request_update` to take slightly different params in `data_dict`, or should the structure of `data_dict` be consistent across `actions` and `auth` functions with the same name?
    - To what extent should auth functions assume inputs are already validated and to what extent should they validate their own input? On the one hand, it seems wasteful to repeat the same validations in `auth.access_request_update` and `actions.access_request_update`, but then on the other hand we actually frequently call auth functions independently of their corresponding action function.
- You posted this link to CKAN raising `Action function {} did not call its auth function`, but I can't get this to happen. It doesn're fire with the implementation that relies on `toolkit.check_access('dataset_collaborator_create', context, data_dict)` and `toolkit.check_access('organization_member_create', context, data_dict)`. Then if I define an action and auth function with the same name and _don't_ call the auth function in the action, it also doesn't happen. When is this condition supposed to fire?

chris48s:
I reckon this is good for another round of comments.

Having added the dialog for to provide a reason when you reject an access request, it would be quite nice to also bring the "new Data Container" reject email into line with that and do the same thing on that too, but I've got to stop pulling this thread somewhere. I think we can defer that to a follow-up PR.

ckanext/unhcr/actions.py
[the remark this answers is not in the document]
amercader: I know that core gets the db model first and adds it to `context` before calling the auth function in some actions but this is an anti-pattern that leads to very difficult to track bugs when the context is reused. In this particular case it's probably fine but I'll rather do two model queries, one on the action and one on the auth function (sqlalchemy will have the object in the session cache anyway) that pass stuff around in the context.

data_dicts passed to actions and their auth functions should be consistent.
I'm not sure why the `Action function {} did not call its auth function` check is not fired in this case, the logic is not the easiest to follow, maybe you need to call get_action multiple times on the same request.

ckanext/unhcr/auth.py
```diff
@@ -256,3 +257,37 @@ def dataset_collaborator_create(next_auth, context, data_dict):
...
+
+    if request.object_type == 'package':
+        package = toolkit.get_action('package_show')(
+            context, {'id': request.object_id}
+        )
+        org_id = package['owner_org']
+    elif request.object_type == 'organization':
+        org_id = request.object_id
+    else:
+        raise toolkit.Invalid("Unknown Object Type")
```
amercader, line 287: Perhaps a ValidationError is more appropriate here but in any case neither are catch by the blueprint function so it will raise a 500 error

ckanext/unhcr/blueprint.py
```diff
@@ -74,10 +77,23 @@ def access_requests_reject(request_id):
     if (not hasattr(toolkit.c, "user") or not toolkit.c.user):
         return toolkit.abort(403, "Forbidden")
 
+    message = toolkit.request.form.get('message')
+    if not message:
+        return toolkit.abort(400, "'message' is required")
+
     try:
         toolkit.get_action('access_request_update')(
             {'user': toolkit.c.user}, {'id': request_id, 'status': 'rejected'}
```
amercader, line 86: Perhaps we should store the message as part of the request for keep an audit log, but we can do that in the future (if needed)

ckanext/unhcr/blueprint.py
```diff
@@ -74,10 +77,23 @@ def access_requests_reject(request_id):
...
         )
+
+        request = model.Session.query(AccessRequest).get(request_id)
```
amercader, line 89: `access_request_update` should return the updated dict to follow the pattern of the rest of `*_update` actions, and that will save you an extra call to `*_show`

amercader:
@chris48s just some minor changes and then we are good to merge!

ckanext/unhcr/actions.py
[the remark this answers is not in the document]
chris48s: updated: a9325e5

ckanext/unhcr/blueprint.py
chris48s: updated: 64d4607

ckanext/unhcr/auth.py
chris48s: I think its fine for this to be unhandled - its one of those

```py
else:
    # this should never happen
```

situations. `object_type` comes from the DB (you can't user-supply it) and there is an `Enum` constraint on the column:

```py
object_type = Column(
    Enum('package', 'organization', name='access_request_object_type_enum'),
    nullable=False,
)
```

Personally I perfer to write with a guard condition, like this:

```py
if request.object_type == 'package':
    # do a thing
elif request.object_type == 'organization':
    # do a different thing
else:
    raise toolkit.Invalid("Unknown Object Type")
```

than write:

```py
if request.object_type == 'package':
    # do a thing
else:
    # do a different thing
```

so that if we add another value to the `Enum` at some later date it is more obvious to the developer what needs to happen next: the function will fail instead of just doing the wrong thing, but there is no user interaction that can throw this.

ckanext/unhcr/auth.py
chris48s: silence is golden on this one?

ckanext/unhcr/auth.py
amercader: You are right, let's leave it

amercader:
All good, great work @chris48s !

Issue #304, opened 2020-05-27
DataPusher not working on AWS UAT


amercader:
When manually uploading a data file to the DataStore via DataPusher it gets stuck on the download stage.

![Screenshot_2020-05-27 Ethiopia WASH KAP 2017 - Unnamed resource - RIDL](https://user-images.githubusercontent.com/200230/83024745-108ffb80-a02f-11ea-9927-b5b2308a7658.png)

Logs show:

```
| 1590585301022 | Fetching from: https://ridl-uat.unhcr.org/dataset/89df5809-5c08-48c2-b90f-561ff1f6d6cd/resource/f9378617-321b-491f-a4a1-64516038db9f/download/unhcr_eth_kap_2017_gambella_tierkidi_data_cleaned_facsimile.csv        |
| 1590585304805 | Error notifying listener                                                                                                                                                                                             |
| 1590585304805 | Traceback (most recent call last):                                                                                                                                                                                   |
| 1590585304805 |   File "/usr/lib/python2.7/site-packages/apscheduler/scheduler.py", line 239, in _notify_listeners                                                                                                                   |   
| 1590585304805 |     cb(event)                                                                                                                                                                                                        |
| 1590585304805 |   File "/usr/lib/python2.7/site-packages/ckanserviceprovider/web.py", line 184, in job_listener                                                                                                                      |
| 1590585304805 |     db.mark_job_as_errored(job_id, error_object)                                                                                                                                                                     |
| 1590585304805 |   File "/usr/lib/python2.7/site-packages/ckanserviceprovider/db.py", line 413, in mark_job_as_errored                                                                                                                |
| 1590585304805 |     _update_job(job_id, update_dict)                                                                                                                                                                                 |
| 1590585304805 |   File "/usr/lib/python2.7/site-packages/ckanserviceprovider/db.py", line 348, in _update_job                                                                                                                        |
| 1590585304805 |     job_dict["error"] = json.dumps(job_dict["error"])                                                                                                                                                                |
| 1590585304805 |   File "/usr/lib/python2.7/json/__init__.py", line 244, in dumps                                                                                                                                                     |
| 1590585304805 |     return _default_encoder.encode(obj)                                                                                                                                                                              |
| 1590585304805 |   File "/usr/lib/python2.7/json/encoder.py", line 207, in encode                                                                                                                                                     |   
| 1590585304805 |     chunks = self.iterencode(o, _one_shot=True)                                                                                                                                                                      |
[...]

PR #314, opened 2020-06-04
simplify datastore api auth code (to fix datapusher job)

chris48s:
refs #304

Again, this is mostly just stuff I've already said in Slack or on a call, but I'll re-state it here for future reference:

* In the datapusher code we make a call to our CKAN instance's API which passes some params in the query string: https://github.com/ckan/datapusher/blob/817ac25fbfbb2dd0a5cbe9de18cf602c99ed8f97/datapusher/jobs.py#L215-L222
* When CKAN core processes an API call, it requires the action to be `side_effect_free` to accept the API params in the query string: https://github.com/ckan/ckan/blob/7900935de6270ee80ec64a5378251dafedf3a0e4/ckan/views/api.py#L263-L266 (essentially, we shouldn't mutate state via `GET`)
* Our `datastore_search` isn't tagged as `side_effect_free` and uncommenting https://github.com/okfn/ckanext-unhcr/blob/d04287154acc73c23e88a64839a598bcf8c05110/ckanext/unhcr/actions.py#L428 doesn't fix that
* This causes the API call to `datastore_search` to fail, which in turn means the datapusher can't push to the datastore

I still haven't really got to the bottom of why the `side_effect_free` decorator isn't working as expected here. I think there is some odd interaction when using `side_effect_free` on a `chained_action`. However...

* It seems like the point of the `datastore_*` action functions is to run our custom auth function, but this is acheived by registering them in `IAuthFunctions` anyway. https://github.com/okfn/ckanext-unhcr/blob/d04287154acc73c23e88a64839a598bcf8c05110/ckanext/unhcr/plugin.py#L365-L367
* There doesn't seem to be any good reason for `auth_allow_anonymous_access` given everything in RIDL requires authentication and we pass the datapusher job a token when we register a job.

I think we can just remove all of this.
This also has the useful effect that if we do this, we're now using the default actions from CKAN core, `datastore_search` etc are marked `side_effect_free` and the datapusher works as expected.

Although I've manually tested the API, I haven't added any unit tests in this PR. There are some disabled tests in
https://github.com/okfn/ckanext-unhcr/blob/d04287154acc73c23e88a64839a598bcf8c05110/ckanext/unhcr/tests/test_actions.py#L299-L336
which should help us here, but after spending a small amount of time trying to fix them I think I'm going to end up disappearing down another fairly deep rabbit hole there. In the interests of staying focussed and moving the AWS project forward, I'm going to suggest I leave them as they are but raise an issue to get them working another day.

I'm reasonably sure this makes sense, but please do sanity check as I do lack a certain amount of context on this one.

### DIFF
--- a/ckanext/unhcr/actions.py
+++ b/ckanext/unhcr/actions.py
@@ -415,29 +415,6 @@ def recently_changed_packages_activity_list_html(context, data_dict):
 
 # Datastore
 
-# We ended up with double auth checks for every action
-# because auth_audit requires calling non-unhcr auth functions too
-
-@toolkit.chained_action
-def datastore_info(action, context, data_dict):
-    toolkit.check_access('unhcr_datastore_info', context, data_dict)
-    return action(context, data_dict)
-
-
-@toolkit.chained_action
-#  @toolkit.side_effect_free
-def datastore_search(action, context, data_dict):
-    toolkit.check_access('unhcr_datastore_search', context, data_dict)
-    return action(context, data_dict)
-
-
-@toolkit.chained_action
-#  @toolkit.side_effect_free
-def datastore_search_sql(action, context, data_dict):
-    toolkit.check_access('unhcr_datastore_search_sql', context, data_dict)
-    return action(context, data_dict)
-
-
 @toolkit.side_effect_free
 def datasets_validation_report(context, data_dict):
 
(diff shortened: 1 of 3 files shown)


chris48s:
Cool. I've made an issue for datastore tests: #315

PR #318, opened 2020-06-08
fix labels on 'add member' form

### LINKED ISSUE #317: Fix formatting in Add Member form
Screenshot_2020-06-08 Add Member - East Horn of Africa, and Great Lakes - Data Containers - RIDL

It shouldn't take the whole container and the labels are misaligned. It's probably just a problem with the classes applied to the form.

chris48s:
closes #317

Screenshot at 2020-06-08 11-26-40

I think the widths are fine as they are. Its broadly the same as default CKAN (except we don't have the right panel for username)

Screenshot at 2020-06-08 11-26-57

If you want we can move the whole form content into a narrower container e.g:

Screenshot at 2020-06-08 11-26-04

### DIFF
--- a/ckanext/unhcr/templates/organization/member_new.html
+++ b/ckanext/unhcr/templates/organization/member_new.html
@@ -2,7 +2,7 @@
 {% set deposit = h.get_data_deposit() %}
 
 {% block form %}
-<form class="dataset-form form-horizontal add-member-form" method='post'>
+<form class="dataset-form add-member-form" method='post'>
   {# Username #}
   <div class="control-group control-medium">
     <label class="control-label" for="username">


amercader:
Just to be clear, do the changes in this PR make the container narrower like in the last screenshot? If so this can be merged

chris48s:
3f74760 only fixes the labels alignment (i.e: the first screenshot)

but I've just pushed a second commit f56bd68 which moves the form into a narrower div (i.e: the third screenshot). It looks like a big diff because a lot of indenting has changes, but its only 2 lines really: https://github.com/okfn/ckanext-unhcr/pull/318/commits/f56bd6809964eb8653bff8effbd17e5e7d607767?w=1 (diff ignoring whitespace)

PR #319, opened 2020-06-08
don't allow collaborators to be added to a deposited dataset

### LINKED ISSUE #316: Disable collaborators tab on deposited datasets
The tab is added on `package/edit_base.html`. We just need to check that the `type` of the dataset is not `deposited-dataset` to display it.

chris48s:
closes #316 

Rather than doing this in the template code, I've done it at the auth level. This probably makes sense anyway and its what the template code uses to decide whether to show the tab or not:

https://github.com/okfn/ckanext-collaborators/blob/e48db4ad309a76f79e34fb1fbf16b0d20959dbea/ckanext/collaborators/templates/package/edit_base.html#L5

### DIFF
--- a/ckanext/unhcr/auth.py
+++ b/ckanext/unhcr/auth.py
@@ -60,6 +60,7 @@ def restrict_access_to_get_auth_functions():
     overriden_auth_functions['package_create'] = package_create
     overriden_auth_functions['package_update'] = package_update
     overriden_auth_functions['package_activity_list'] = package_activity_list
+    overriden_auth_functions['dataset_collaborator_create'] = dataset_collaborator_create
 
     return overriden_auth_functions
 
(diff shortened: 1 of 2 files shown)


ckanext/unhcr/auth.py
```diff
@@ -60,6 +60,7 @@ def restrict_access_to_get_auth_functions():
     overriden_auth_functions['package_create'] = package_create
     overriden_auth_functions['package_update'] = package_update
     overriden_auth_functions['package_activity_list'] = package_activity_list
+    overriden_auth_functions['dataset_collaborator_create'] = dataset_collaborator_create
```
chris48s, line 63: Today I found out some of the custom auth functions are registered here whereas some of them are registered directly in `get_auth_functions()` in `plugin.py`. Is there a reason for this/any difference? Should they all really be registered here?

amercader: No, we should only register here overridden **get** functions from core. The end result is the same but it's clearer this way. All the other ones should be defined in `plugin.py` `get_auth_functions()`. Can you move the `*_create, *_update` ones there?

amercader:
Auth function is obviously better than the template. Let's hold merging this though until we discuss it with Mariann

PR #325, opened 2020-06-11
hard-code maintenance banner

chris48s:
This hard-codes a notification at the top of any page which inherits from page.html aside from the login page.

I've made this a PR just so it gives me a place to put notes but I don't actually think we should merge it to master. We should just do an ad-hoc deploy using `pip install -e git+https://github.com/okfn/ckanext-unhcr.git@maintenance-banner#egg=ckanext-unhcr` in the Dockerfile.

We should confirm with the client that 10am to midday CEST is actually the window they want to block out for this.

### DIFF
--- a/ckanext/unhcr/templates/page.html
+++ b/ckanext/unhcr/templates/page.html
@@ -36,6 +36,9 @@
                   {% block flash %}
                     <div class="flash-messages">
                       {% block flash_inner %}
+                        <div class="alert fade in alert-info">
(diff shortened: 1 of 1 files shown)


amercader:
I think it was actually 12am to 2pm but let's confirm

chris48s:
updated

chris48s:
Give we've deployed from this branch I'll close this now

Issue #324, opened 2020-06-11
Add a config option for summary emails


chris48s:
When we run `send_summary_emails()`
https://github.com/okfn/ckanext-unhcr/blob/876b288c7bd0daa32ea69ca050a60e752871c72a/ckanext/unhcr/commands.py#L88
check an config env var to decide whether to send or not - this will make it easier for us to turn emails on in production but off in AUT when sharing deploy scripts between the two.

PR #328, opened 2020-06-12
add a config setting for send_summary_emails

chris48s:
closes https://github.com/okfn/ckanext-unhcr/issues/324
refs https://github.com/okfn/docker-ckan-unhcr-aws/pull/3


ckanext/unhcr/commands.py
```diff
@@ -2,9 +2,10 @@
 
 import sys
 
+from ckan.common import config
```
amercader, line 5: Use `toolkit.config` here

PR #331, opened 2020-06-15
add 'keywords' table to metrics dashboard

### LINKED ISSUE #306: Metrics: Add table for datasets by topic
> It would be great, if possible, if the dashboard had statistics by keyword/topic (Water Sanitation Hygiene, Health, …)

Use the "Topic classifications" field

chris48s:
..and add a little helper function to DRY up making  facet-based tables.

closes #306

Might be a bit of a can of worms, but it seems like there's quite a big conceptual overlap here between tags and keywords. Is there a really clear distinction between how they should be used?

### DIFF
--- a/ckanext/unhcr/blueprint.py
+++ b/ckanext/unhcr/blueprint.py
@@ -9,6 +9,7 @@ from .metrics import (
     get_containers,
     get_containers_by_date,
     get_tags,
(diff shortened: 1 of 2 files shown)


amercader:
@chris48s tags are free text, keywords are taken from a predefined list.

PR #338, opened 2020-06-19
[#332] Fix resource editing

### LINKED ISSUE #332: Editing existing resources fails with 'All resources require an uploaded file'
Trying to edit the metadata on a resource that has already been uploaded fails with  'All resources require an uploaded file'. We're failing on this validator:

https://github.com/okfn/ckanext-unhcr/blob/876b288c7bd0daa32ea69ca050a60e752871c72a/ckanext/unhcr/validators.py#L182-L186

When I first discovered it, I was slightly worried we introduced this bug with the cloudstorage plugin, but it seems to reproduce with the default storage engine, so I think this has been a problem for some time and nobody has noticed.

If I change the validator to

```py
import ckan.lib.navl.dictization_functions as df

def upload_not_empty(key, data, errors, context):
    missing = df.missing
    index = key[1]
    url_type = data[('resources', index, 'url_type')]
    if url_type is not missing and url_type != 'upload':
        errors[('resources', index, 'url',)] = ['All resources require an uploaded file']
```

that fixes the issue with saving an existing resource, but it means we can create a new resource with no file attached to it at all, which isn't what we want either.

@amercader - any thoughts on how to fix this one? Can we detect whether we're creating  a new resource or editing an existing one from within the context of a validator?

amercader:
Fixes #332 

Modified the validator to take into account updates. Added tests for normal edits and to make sure you can not remove an upload when editing.

### DIFF
--- a/ckanext/unhcr/tests/test_actions.py
+++ b/ckanext/unhcr/tests/test_actions.py
@@ -375,6 +375,31 @@ class TestResourceUpload(base.FunctionalTestBase):
             )
         )
 
+    def test_upload_present_after_update(self):
+
+        dataset = factories.Dataset()
+
+        resource = factories.Resource(
+            package_id=dataset['id'],
+            upload=mocks.FakeFileStorage(),
+            url = "http://fakeurl/test.txt",
+            url_type='upload',
+        )
+
+        resource['name'] = 'updated'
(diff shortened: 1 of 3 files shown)


amercader:
All green now

PR #343, opened 2020-06-24
surface "download resource" activities in the front-end

### LINKED ISSUE #284: New admin activity stream, showing curation activities and downloads

chris48s:
This PR rebrands curation activities as "internal activities" and exposes resource download activities to container admins only, so:

- a container admin can see all the internal activities
- a dataset editor can only see the curation activities
- a container member can't see any internal activities
- an unprivileged user can't see any internal activities

Can you confirm that matches with your understanding of the feature requirements?

Closes #284

### DIFF
--- a/ckanext/unhcr/actions.py
+++ b/ckanext/unhcr/actions.py
@@ -256,11 +256,11 @@ def pending_requests_list(context, data_dict):
 
 @toolkit.side_effect_free
 def package_activity_list(context, data_dict):
-    get_curation_activities = toolkit.asbool(
-        data_dict.get('get_curation_activities'))
+    get_internal_activities = toolkit.asbool(
+        data_dict.get('get_internal_activities'))
     full_list = get_core.package_activity_list(context, data_dict)
     full_list = [a for a in full_list if a["activity_type"] != "download resource"]
-    curation_activities = [
+    internal_activities = [
         a for a in full_list if 'curation_activity' in a.get('data', {})]
     normal_activities = [
         a for a in full_list if 'curation_activity' not in a.get('data', {})]
@@ -271,8 +271,8 @@ def package_activity_list(context, data_dict):
             .get('data', {})
             .get('package_extra', {})
             .get('key') not in ('curation_state', 'curator_id'), normal_activities))
-    return (curation_activities
-        if get_curation_activities else normal_activities)
+    return (internal_activities
+        if get_internal_activities else normal_activities)
 
 
 @toolkit.side_effect_free
--- a/ckanext/unhcr/actions.py
+++ b/ckanext/unhcr/actions.py
@@ -254,25 +254,45 @@ def pending_requests_list(context, data_dict):
 
 # Activity
 
+
+def _package_internal_activity_list(context, data_dict):
+    full_list = get_core.package_activity_list(context, data_dict)
+    activities = [
+        a for a in full_list
+        if 'curation_activity' in a.get('data', {})
+        or a["activity_type"] == "download resource"
+    ]
+    return activities
+
+
+def _package_normal_activity_list(context, data_dict):
+    full_list = get_core.package_activity_list(context, data_dict)
+    activities = [
+        a for a in full_list
+        if 'curation_activity' not in a.get('data', {})
+        and a["activity_type"] != "download resource"
+    ]
+    # Filter out the activities that are related to `curation_state`
+    activities = list(
+        filter(
+            lambda activity: get_core.activity_detail_list(
+                context, {'id': activity['id']}).pop()
+                .get('data', {})
+                .get('package_extra', {})
+                .get('key') not in ('curation_state', 'curator_id'),
+            activities
+        )
+    )
+    return activities
+
+
 @toolkit.side_effect_free
 def package_activity_list(context, data_dict):
     get_internal_activities = toolkit.asbool(
         data_dict.get('get_internal_activities'))
-    full_list = get_core.package_activity_list(context, data_dict)
-    full_list = [a for a in full_list if a["activity_type"] != "download resource"]
-    internal_activities = [
-        a for a in full_list if 'curation_activity' in a.get('data', {})]
-    normal_activities = [
-        a for a in full_list if 'curation_activity' not in a.get('data', {})]
-    # Filter out the activities that are related `curation_state`
-    normal_activities = list(filter(
-        lambda activity: get_core.activity_detail_list(
-            context, {'id': activity['id']}).pop()
-            .get('data', {})
-            .get('package_extra', {})
-            .get('key') not in ('curation_state', 'curator_id'), normal_activities))
-    return (internal_activities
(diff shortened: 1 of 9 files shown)


ckanext/unhcr/auth.py
```diff
@@ -160,10 +160,10 @@ def package_update(next_auth, context, data_dict):
 
 
 def package_activity_list(context, data_dict):
-    if toolkit.asbool(data_dict.get('get_curation_activities')):
-        # Check if the user can see the curation activity,
+    if toolkit.asbool(data_dict.get('get_internal_activities')):
+        # Check if the user can see the internal activity,
         # for now we check if the user can edit the dataset
-        return auth_update_core.package_update(context, data_dict)
+        return {'success': toolkit.check_access('package_update', context, data_dict)}
```
chris48s, line 166: I made this change to account for the fact that there's an auth chain. We define our own `package_update` and so does `ckanext-collaborators` so we don't want to call core's `package_update` directly. `toolkit.check_access` does invoke the full chain. This does seem to work and we do seem to be doing the same thing elsewhere. For example:

https://github.com/okfn/ckanext-unhcr/blob/bb22acb7b37a4e38514d1a589aa8f93e3338bbeb/ckanext/unhcr/auth.py#L200-L201

..but I wonder if that's actually correct. Should we really `return {'success': False}` here instead of throwing an exception if the user doesn't have sufficient privileges? Any thoughts on the best way to handle it?

ckanext/unhcr/helpers.py
```diff
@@ -461,6 +461,16 @@ def create_curation_activity(
     toolkit.get_action('activity_create')(activity_context, data_dict)
 
 
+def download_resource_renderer(context, activity):
```
chris48s, line 464: After spending some more time on it, it turned out this was actually simpler than I initially thought. Think I was just making life hard for myself :slightly_smiling_face:

ckanext/unhcr/actions.py
```diff
@@ -251,25 +252,61 @@ def pending_requests_list(context, data_dict):
...
-            .get('data', {})
-            .get('package_extra', {})
-            .get('key') not in ('curation_state', 'curator_id'), normal_activities))
-    return (curation_activities
-        if get_curation_activities else normal_activities)
+    toolkit.check_access('package_activity_list', context, data_dict)
+    get_internal_activities = toolkit.asbool(
+        data_dict.get('get_internal_activities'))
+
+    package = context.get('package')
```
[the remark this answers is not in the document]
amercader: See my comment on #308 about relying on objects in the `context`. You can do `package = model.Package.get(id)` here, no need to call the action.

ckanext/unhcr/actions.py
```diff
@@ -251,25 +252,61 @@ def pending_requests_list(context, data_dict):
...
+                .get('package_extra', {})
+                .get('key') not in ('curation_state', 'curator_id'),
+            activities
+        )
+    )
+    return activities
+
+
 @toolkit.side_effect_free
 def package_activity_list(context, data_dict):
```
amercader, line 293: Now that `@toolkit.chained_action` support is better than when we wrote this let's use a chained action to call the core list and remove the core imports

```python
def package_activity_list(up_func, context, data_dict):
    full_list = up_func(context, data_dict)
```

ckanext/unhcr/auth.py
amercader: Good call changing it to allow chaining the auth functions. I think the end result is the same if we call check_access here, we end up with a `NotAuthorized` exception (raised by this check_access or by the parent one). But if you want to return a boolean just call check_access and return False if you catch a not authorized exception here.

ckanext/unhcr/actions.py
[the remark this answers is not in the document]
chris48s: When you say "You can do `package = model.Package.get(id)` here..." is there a rule/convention about when you're supposed to interact with the DB directly and when you're supposed to do it via the action layer?

Seems like more of an art than a science so far..

ckanext/unhcr/auth.py
chris48s: Its minor, but I think its better if all the functions at this layer return False instead of throwing. I applied the same change in the `resource_download` auth function too

ckanext/unhcr/actions.py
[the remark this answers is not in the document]
chris48s: done in 0a75405 anyhow

ckanext/unhcr/actions.py
chris48s: updated: bc4eb53

ckanext/unhcr/actions.py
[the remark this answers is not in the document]
amercader: For read operations, I guess the rule of thumb is to query the db directly on actions, unless you need to take into account potential changes that other extensions do. So for instance in this case we just need to get the owner_org so it's fine to work with the models. If for some reason you action eg needed all extra fields you probably want to call `package_show` to account for fields added by other extensions. 
For side-effect actions like update or create you probably want to always call the action to enable other customizations to take effect

PR #364, opened 2020-06-30
manage sysadmins from the UI

### LINKED ISSUE #348: manage sysadmins from the UI

chris48s:
closes #348

### DIFF
--- a/ckanext/unhcr/actions.py
+++ b/ckanext/unhcr/actions.py
@@ -12,6 +12,7 @@ import ckan.logic.action.delete as delete_core
 import ckan.logic.action.update as update_core
 import ckan.logic.action.patch as patch_core
 import ckan.lib.activity_streams as activity_streams
+import ckan.lib.dictization.model_dictize as model_dictize
 from ckanext.unhcr import helpers, mailer
 from ckanext.unhcr.models import AccessRequest
 from ckanext.scheming.helpers import scheming_get_dataset_schema
@@ -639,3 +640,32 @@ def access_request_update(context, data_dict):
         col.name: getattr(request, col.name)
         for col in request.__table__.columns
     }
+
+
+# User
+
+def user_update_sysadmin(context, data_dict):
+    """
+    Add or remove a sysadmin user
+    An authenticated sysadmin can promote an existing user to sysadmin
+    or remove sysadmins priveledges from a user who already has them
+
+    :param id: The id or name of the user
+    :type id: string
+    :param is_sysadmin: The new value of User.sysadmin
+    :type is_sysadmin: bool
+    """
+    m = context.get('model', model)
+    user_id, is_sysadmin = toolkit.get_or_bust(data_dict, ['id', 'is_sysadmin'])
+    is_sysadmin = toolkit.asbool(is_sysadmin)
+
+    toolkit.check_access('user_update_sysadmin', context, data_dict)
+
+    user_obj = m.User.get(user_id)
+    if not user_obj:
+        raise toolkit.ObjectNotFound("User not found")
+    user_obj.sysadmin = is_sysadmin
+    m.Session.commit()
+    m.Session.refresh(user_obj)
+
+    return model_dictize.user_dictize(user_obj, context)
--- a/ckanext/unhcr/auth.py
+++ b/ckanext/unhcr/auth.py
@@ -302,3 +302,9 @@ def access_request_update(context, data_dict):
             org_id, user, 'admin'
         )
     }
+
+
+# User
+
+def user_update_sysadmin(context, data_dict):
+    return {'success': False}
--- a/ckanext/unhcr/plugin.py
+++ b/ckanext/unhcr/plugin.py
@@ -383,6 +383,7 @@ class UnhcrPlugin(
         functions['dataset_collaborator_create'] = auth.dataset_collaborator_create
         functions['access_request_list_for_user'] = auth.access_request_list_for_user
         functions['access_request_update'] = auth.access_request_update
+        functions['user_update_sysadmin'] = auth.user_update_sysadmin
         return functions
 
     # IActions
@@ -411,6 +412,7 @@ class UnhcrPlugin(
             'organization_activity_list_html': actions.organization_activity_list_html,
             'recently_changed_packages_activity_list_html': actions.recently_changed_packages_activity_list_html,
             'datasets_validation_report': actions.datasets_validation_report,
+            'user_update_sysadmin': actions.user_update_sysadmin,
         }
 
     # IValidators
@@ -474,6 +476,8 @@ class UnhcrPlugin(
 
     def get_blueprint(self):
         return [
-            blueprints.unhcr_metrics_blueprint,
             blueprints.unhcr_access_requests_blueprint,
+            blueprints.unhcr_admin_blueprint,
+            blueprints.unhcr_metrics_blueprint,
+            blueprints.unhcr_user_blueprint,
         ]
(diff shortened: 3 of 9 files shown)


ckanext/unhcr/plugin.py
```diff
@@ -135,6 +135,12 @@ def before_map(self, _map):
         # user
         controller = 'ckanext.unhcr.controllers.extended_user:ExtendedUserController'
         _map.connect('dashboard.requests', '/dashboard/requests', controller=controller, action='list_requests', ckan_icon='spinner')
+        _map.connect('/user/sysadmin', controller=controller, action='sysadmin', conditions={'method': ['POST']})
+
+        # admin
+        controller = 'ckanext.unhcr.controllers.extended_admin:ExtendedAdminController'
+        _map.connect('/ckan-admin', controller=controller, action='index')
+        _map.connect('/ckan-admin/', controller=controller, action='index')
```
chris48s, line 143: In order to get this to work with or without the trailing slash, I had to explicitly register both versions here. If I only register `/ckan-admin`, `/ckan-admin/` falls back to the version from CKAN core instead of our custom one, and vice versa. I'm not sure why. Maybe its because we're overriding a built-in route? Am I doing this wrong? Any advice on how to deal with this correctly?

ckanext/unhcr/controllers/extended_admin.py
```diff
@@ -0,0 +1,15 @@
+from ckan.plugins import toolkit
+
+class ExtendedAdminController(toolkit.BaseController):
```
amercader, line 3: Let's not write any more controllers :)
The ~~future~~ present is blueprints!

amercader:
I've not finished reviewing but just to flag that I really want to port this upstream!

ckanext/unhcr/controllers/extended_admin.py
chris48s: Glad you raised this. I did think about this, but I kind of lost sight of it as a question... I'm a bit unclear about how to correctly deal with blueprints in a plugin with lots of routes. Now is a good time to clear that up.

It seems like the correct way to do this should be to have multiple blueprints grouping related routes. For example, CKAN core declares a dataset blueprint, an admin blueprint and so on.

As far as I can tell, the `IBlueprint` plugin interface only allows us to return a single `Blueprint` object for the entire plugin. This is why in PR #308 I swapped out `unhcr_metrics_blueprint` for a generic `unhcr_blueprint`. This seems like its going to get quite busy when we port all our custom controllers to blueprints with 2.9. Is it possible to construct nested blueprints? Is there a plan to allow plugins to declare an array of blueprints?

I was also being slightly lazy and trying to avoid a merge conflict with #308 ;) but I can pull the same change into here and rebase one PR or the other later.

chris48s:
Excellent. I'm happy to work on submitting a PR to core once we're happy with it in RIDL :)

ckanext/unhcr/controllers/extended_admin.py
amercader: Sorry, I'll go into more detail later but just to mention that `IBlueprint.get_blueprint()` can return a lists of blueprints!

chris48s: That changes the game then :tada: 
At the moment this is not obvious from the [docs](https://docs.ckan.org/en/latest/extensions/plugin-interfaces.html#ckan.plugins.interfaces.IBlueprint) so I've submitted a small PR to core updating the docs: https://github.com/ckan/ckan/pull/5489
I'll refactor this PR and #308 accordingly to make use of that feature.

ckanext/unhcr/plugin.py
chris48s: when i moved this to a blueprint, I managed to solve this with `strict_slashes=False` but I'm still a bit unclear why I did need it here, but not on other flask routes e.g: `/metrics`

ckanext/unhcr/controllers/extended_admin.py
chris48s: I've added 2478f1c moving `ExtendedAdminController` to a flask blueprint. I also want to move `ExtendedUser.sysadmin`, but I'm also thinking I want to do a bit of re-organising to split up `blueprint.py` into a directory with one-file-per Blueprint object. I reckon I'm going to just put this one on hold until we've got #308 reviewed then I'll circle back, rebase this and do the re-org then.

ckanext/unhcr/controllers/extended_user.py
```diff
@@ -26,6 +26,27 @@ def list_requests(self):
...
+        try:
+            toolkit.check_access('sysadmin', {})
+        except toolkit.NotAuthorized:
+            return toolkit.abort(403, 'Not authorized to promote user to sysadmin')
+
+        user = model.User.get(id_)
+        if not user:
+            return toolkit.abort(404, 'User not found')
+        user.sysadmin = status
+        model.Session.commit()
```
amercader, line 42: This should be wrapped in an action, and called from the the blueprint, otherwise it won't be accessible via the API. Here's an implementation from another project (essentially the same you have):

```python
def user_update_sysadmin(context, data_dict):
    """Add or remove a sysadmin user

    An authenticated sysadmin can promote an existing user to sysadmin
    """
    model = context['model']
    username, is_sysadmin = toolkit.get_or_bust(data_dict, ['username', 'is_sysadmin'])
    is_sysadmin = toolkit.asbool(is_sysadmin)

    logic.check_access('user_update_sysadmin', context, data_dict)

    user_obj = model.User.get(username)
    if not user_obj:
        raise toolkit.ObjectNotFound()

    user_obj.sysadmin = is_sysadmin
    model.Session.add(user_obj)
    model.Session.commit()

    return user_obj.as_dict()

```

ckanext/unhcr/controllers/extended_user.py
```diff
@@ -26,6 +26,27 @@ def list_requests(self):
             'requests': requests,
         })
 
+    def sysadmin(self):
```
amercader, line 29: Let's not add more stuff to the controllers and move this to the admin blueprint

chris48s: agreed - see https://github.com/okfn/ckanext-unhcr/pull/364#discussion_r453545148
Should have been clearer this one needs more work

chris48s:
I went back and revisited this based on the changes in https://github.com/okfn/ckanext-unhcr/pull/390 I've also taken into account all the suggested changes. This should be good for another look now.

ckanext/unhcr/actions.py
```diff
@@ -639,3 +639,34 @@ def access_request_update(context, data_dict):
...
+    toolkit.check_access('user_update_sysadmin', context, data_dict)
+
+    user_obj = m.User.get(user_id)
+    if not user_obj:
+        raise toolkit.ObjectNotFound("User not found")
+    user_obj.sysadmin = is_sysadmin
+    m.Session.commit()
+    m.Session.refresh(user_obj)
+
+    user_dict = user_obj.as_dict()
```
amercader, line 670: It's best to use the [dictization functions](https://github.com/ckan/ckan/blob/master/ckan/logic/action/update.py#L866) if possible as these have extra logic to add/remove fields (like the `display_name` one below)

ckanext/unhcr/auth.py
```diff
@@ -302,3 +302,13 @@ def access_request_update(context, data_dict):
...
+
+
+# User
+
+def user_update_sysadmin(context, data_dict):
+    try:
+        toolkit.check_access('sysadmin', {})
+        return {'success': True}
+    except toolkit.NotAuthorized:
+        return {'success': False}
```
amercader, line 314: No need to check if the user is a sysadmin. Sysadmins bypass these auth functions, so for sysadmin only actions is best to do:

```python
def user_update_sysadmin(context, data_dict):
    return {'success': False}
```

chris48s:
I've made those 2 small changes and squashed. Going to merge this now. I'll also add a task to my list to work on a PR upstreaming these changes to CKAN core.

PR #366, opened 2020-07-01
add a banner to UAT

### LINKED ISSUE #360: Make the UAT interface easily distinguishable for users
...because we had misunderstandings and accidents just recently. Maybe a permanent yellow banner at the top of the page...? Every page?

chris48s:
Refs https://github.com/okfn/docker-ckan-unhcr-aws/pull/8
Closes #360

#360 is a useful idea. I have more than once been confused about this and its something I do usually do for test/staging sites etc. We probably should have thought of it already. I've tried to strike a decent balance here between the banner being obvious enough to be useful at a glance and subtle enough that its not annoying if its permanently on every page in UAT.

### DIFF
--- a/ckanext/unhcr/helpers.py
+++ b/ckanext/unhcr/helpers.py
@@ -1,4 +1,5 @@
 import logging
+import os
 import re
 from urllib import quote
 from jinja2 import Markup, escape
@@ -747,6 +748,13 @@ def get_sysadmins():
 
 def get_ridl_version():
     return __VERSION__
+
+
+def get_envname():
+    envname = os.environ.get('ENV_NAME')
+    if not envname:
+        return 'dev'
+    return envname.lower()
 
 
 _paragraph_re = re.compile(r'(?:\r\n|\r(?!\n)|\n){2,}')
--- a/ckanext/unhcr/templates/page.html
+++ b/ckanext/unhcr/templates/page.html
@@ -10,6 +10,11 @@
 
 
 {% block page %}
+{% if h.get_envname() not in ('prod', 'dev') %}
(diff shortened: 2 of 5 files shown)


ckanext/unhcr/templates/page.html
```diff
@@ -10,6 +10,11 @@
 
 
 {% block page %}
+{% if h.get_envname() != 'prod' and h.get_envname() != 'dev' %}
```
amercader, line 13: Can you use `h.get_envname() not in ('prod', 'dev')` in jinja2?

ckanext/unhcr/helpers.py
```diff
@@ -749,6 +749,13 @@ def get_ridl_version():
     return __VERSION__
 
 
+def get_envname():
+    envname = toolkit.config.get('ckanext.unhcr.envname')
+    if not envname:
+        return 'dev'
+    return envname
```
amercader, line 756: You should lower case the value, otherwise the checks in the template won't work

ckanext/unhcr/helpers.py
```diff
@@ -749,6 +749,13 @@ def get_ridl_version():
     return __VERSION__
 
 
+def get_envname():
+    envname = toolkit.config.get('ckanext.unhcr.envname')
```
amercader, line 753: See https://github.com/okfn/docker-ckan-unhcr-aws/pull/8#pullrequestreview-440831134

chris48s:
Nice suggestions :+1: I've updated this PR and https://github.com/okfn/docker-ckan-unhcr-aws/pull/8 accordingly.

PR #367, opened 2020-07-01
handle ObjectNotFound cleanly in resource_download

chris48s:
While I was testing something else, I noticed that we're not handling the 404 case properly because `auth.resource_download` doesn't actually throw a `ObjectNotFound` if the resource doesn't exist.

I've done this for now, but I wonder if I should either:

- Call `get_action('resource_show')` inside `auth.resource_download` (is it convention in CKAN that auth functions should always throw a `ObjectNotFound` if the thing we're checking auth against doesn't exist?)
- Just check the auth/403 case here and move the `ObjectNotFound` check to later in the function when we actually try to do something with the resource (which will make one less query in the success case)

### DIFF
--- a/ckanext/unhcr/controllers/extended_package.py
+++ b/ckanext/unhcr/controllers/extended_package.py
@@ -107,7 +107,7 @@ class ExtendedPackageController(PackageController):
     def resource_download(self, id, resource_id, filename=None):
         """
         Wraps default `resource_download` endpoint checking
-        the custom `resoruce_download` auth function
+        the custom `resource_download` auth function
         """
         context = {'model': model, 'session': model.Session,
                    'user': toolkit.c.user, 'auth_user_obj': toolkit.c.userobj}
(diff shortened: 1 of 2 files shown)


ckanext/unhcr/controllers/extended_package.py
```diff
@@ -107,14 +107,15 @@ def resource_copy(self, id, resource_id):
...
         Wraps default `resource_download` endpoint checking
-        the custom `resoruce_download` auth function
+        the custom `resource_download` auth function
         """
         context = {'model': model, 'session': model.Session,
                    'user': toolkit.c.user, 'auth_user_obj': toolkit.c.userobj}
 
         # Check resource_download access
         try:
             toolkit.check_access(u'resource_download', context, {u'id': resource_id})
```
amercader, line 117: I thought [this function](https://github.com/okfn/ckanext-unhcr/blob/master/ckanext/unhcr/auth.py#L190) would raise a `NotFound` exception if the resource was not found?
If not I think it's a bug on the `resource_download` auth function because how can you be authorized something it doesn't exist?

chris48s: Right, yeah. Good point :+1: 
The reason I was seeing this problem was because I was logged in as a sysadmin. If I log in as a normal user this works as expected. I think If you're a sysadmin all auth checks just return true (without actually running the auth functions?)

I guess this must just be an oddity of being logged in as a sysadmin which exists in lots of places then. It seems quite common to rely on the auth function to check for existence of the object.

PR #389, opened 2020-07-16
Add external users

### LINKED ISSUE #371: Add External users
New user type: `external`. This might be as simple as

```py
@property
def external(self):
    return not self.email.endswith("@unhcr.org")
```

or we might need to use User extras (CKAN 2.9)

Find a way to monkey-patch `check_access()` to always fail if `user.external` except for some subset of allowed actions. Can we do it in `update_config()`?

It would be best if we can do this in the plugin rather than having to apply the patch in the docker setup/repo.

chris48s:
Closes #371
Refs https://github.com/okfn/docker-ckan-unhcr-aws/pull/9

The main thing that's going on in this PR is I'm adding a new class of user: `external`. As an opening gambit, external users can log in, but then can't see or do anything. We can start to selectively enable certain actions as a next step.

One of the things I've thought about since we last spoke and hopefully achieved here is testing this functionality in a way that:

- Gives us confidence this works correctly
- Will flag problems if CKAN or plugins change under us
- Doesn't require us to explicitly add an additional `external_user` test case to every single action test

I'll mark up with some more comments inline.

### DIFF
--- a/ckanext/unhcr/plugin.py
+++ b/ckanext/unhcr/plugin.py
@@ -527,6 +527,10 @@ class UnhcrPlugin(
         # For curating users
         # Adding "deposited-dataset" label for data curators
         if user_obj:
+
+            if user_obj.external:
+                return ['']
+
             context = {u'user': user_obj.id}
             deposit = helpers.get_data_deposit()
             orgs = toolkit.get_action('organization_list_for_user')(context, {})
--- a/ckanext/unhcr/plugin.py
+++ b/ckanext/unhcr/plugin.py
@@ -240,6 +240,9 @@ class UnhcrPlugin(
 
     def get_helpers(self):
         return {
+            # Core overrides
+            'new_activities': helpers.new_activities,
+            'dashboard_activity_stream': helpers.dashboard_activity_stream,
             # General
             'get_data_container': helpers.get_data_container,
             'get_all_data_containers': helpers.get_all_data_containers,
--- a/ckanext/unhcr/plugin.py
+++ b/ckanext/unhcr/plugin.py
@@ -4,12 +4,14 @@ import json
 import logging
 
 from ckan.common import config
+from ckan.model import User
 import ckan.plugins as plugins
 import ckan.plugins.toolkit as toolkit
 from ckan.lib.plugins import DefaultTranslation
 from ckan.lib.plugins import DefaultPermissionLabels
 
 # 🙈
+import ckan.authz as authz
 from ckan.lib.activity_streams import (
     activity_stream_string_functions,
     activity_stream_string_icons,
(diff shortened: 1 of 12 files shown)


ckanext/unhcr/plugin.py
```diff
@@ -25,6 +27,53 @@
...
+    return domain not in internal_domains
+
+
+def restrict_external(func):
+    '''
+    Decorator function to restrict external users to a small number of allowed_actions
+    '''
+
+    allowed_actions = [
+        'package_search',
```
chris48s, line 60: TODO: populate with actions that internal users can perform

ckanext/unhcr/plugin.py
```diff
@@ -461,6 +516,10 @@ def get_user_dataset_labels(self, user_obj):
         # For curating users
         # Adding "deposited-dataset" label for data curators
         if user_obj:
+
+            if user_obj.external:
+                return ['']
```
chris48s, line 521: External users can't see anything
TODO: allow them to see deposited datasets they have created themselves

ckanext/unhcr/tests/test_auth.py
```diff
@@ -130,8 +116,48 @@ def test_logged_in_users_private_dataset(self):
...
+            '/data-container/{}'.format(container['name']),
+            '/organization/{}'.format(container['name']),
+            '/group/{}'.format(container['name']),
+        ]
+        for endpoint in endpoints_404:
+            # these throw a 404 rather than a 403
+            resp = app.get(endpoint, extra_environ=env, status=404)
+
+        endpoints_200 = [
+            '/feeds/dataset.atom',
```
chris48s, line 154: TODO: add some more endpoints here as we selectively enable actions

ckanext/unhcr/tests/test_auth.py
```diff
@@ -255,6 +281,55 @@ def test_resource_download(self):
             auth.resource_download({'user': external_user['name']}, resource)
         )
 
+    def test_external_users_core_actions(self):
```
chris48s, line 284: This test should fail if there is a change to the internals in core which means our monkey-patch stops working

ckanext/unhcr/tests/test_auth.py
```diff
@@ -255,6 +281,55 @@ def test_resource_download(self):
...
+                    assert_raises(
+                        toolkit.NotAuthorized,
+                        toolkit.check_access,
+                        action,
+                        context=context
+                    )
+                else:
+                    assert_equals(True, toolkit.check_access(action, context))
+
+    def test_external_users_plugin_actions(self):
```
chris48s, line 313: This test should fail if we add plugins that define completely new auth functions or override existing ones (rather than chaining) without ever calling a core auth function

ckanext/unhcr/tests/test_auth.py
```diff
@@ -130,8 +116,48 @@ def test_logged_in_users_private_dataset(self):
             response.body
         )
 
+    def test_external_users_endpoints(self):
```
chris48s, line 119: This test is not intended to be comprehensive. It may not be that useful, but it feels like a good idea to have some kind of smoke test at the this layer too.

ckanext/unhcr/plugin.py
```diff
@@ -25,6 +27,53 @@
 _ = toolkit._
 
 
+INTERNAL_DOMAINS = ['unhcr.org']
+
+
+def user_is_external(self):
```
amercader, line 33: Can we use a different var name than `self` :) I was really confused looking at this non-class

ckanext/unhcr/plugin.py
amercader: For reference when you implement this:

```
# On get_dataset_labels():

labels.append('creator-external:' + dataset_dict['creator_user_id']

# On get_user_dataset_labels()
if user.external:
     labels =[creator-external:' + user.id]
```

ckanext/unhcr/tests/base.py
```diff
@@ -28,6 +29,8 @@ def setup_class(cls):
         registry.prepare()
         registry.register(pylons.c, c)
 
+        toolkit.config['ckanext.unhcr.internal_domains'] = ['unhcr.org', 'ckan.org']
```
amercader, line 32: Use the `@core_helpers.change_config()` decorator for this, otherwise this leads to hard to track failures across tests

chris48s: We want this on for every test because every existing test we have that creates a user creates it with the assumption that user is "internal". I think `change_config()` can only be used as a function decorator, not a class decorator?

Is the correct approach here actually to set this in `test.ini`?

amercader: If a config needs to apply to all tests in the suite then `test.ini` is the place to put it, yes.
But why do we need all tests to include `ckan.org` as an internal domain and not use the default value (`unchr.org`)?

chris48s: The user factory creates all users with a `@ckan.org` email address i.e:

```py
>>> from ckan.tests import factories as core_factories
>>> user = core_factories.User()
>>> print(user['email'])
test_user_00@ckan.org
```

Considering `ckan.org` as "internal" under test seemed like the path of least resistance here, but if you prefer we could define our own `UserFactory` which creates user objects with a `@unhcr.org` email under test and switch the existing tests to use that instead of the core factory.

amercader: I see, let's add a config in test.ini then

chris48s:
Merging this to the `external-data-deposit` branch so I can continue basing more work on it. We can always revisit some of this stuff before we merge `external-data-deposit` to `master` if necessary

Issue #393, opened 2020-07-23
Container visibility flag


chris48s:
We want to show EDD users a subset of containers. All the ones under Headquarters should be hidden as well as containers for specific agencies/projects under a country. For example. EDD users should be able to see "Ethiopia" but not "Ethiopia : WASH KAP", "Ethiopia: CBI" or "Ethiopia: SENS".
We will need to add a boolean flag to container objects for "visible to EDD users" to specify which containers are shown/hidden.

PR #394, opened 2020-07-24
WIP add container visibility flag

chris48s:
refs #393


ckanext/unhcr/schemas/data_container.json
```diff
@@ -1168,7 +1168,16 @@
             "label": "Tags",
             "preset": "tag_string_autocomplete",
             "form_snippet": "tags.html"
+        },
+        {
+            "field_name": "visible_external",
+            "label": "Visible to External users",
+            "preset": "select",
+            "required": true,
```
amercader, line 1176: Let's use a proper boolean for this:

```
    "validators": "boolean_validator",
    "output_validators": "boolean_validator", 
```

ckanext/unhcr/tests/factories.py
```diff
@@ -6,6 +6,7 @@ class DataContainer(factories.Organization):
     type = 'data-container'
     country = ['SVN']
     geographic_area = 'southern_africa'
+    visible_external = 'Yes'
```
amercader, line 9: Should be `True` when you add the validators above

ckanext/unhcr/schemas/data_container.json
chris48s: I spent quite a while trying to figure out if I could do this at the time, but I came to the conclusion we can't use a proper boolean because ckanext-scheming doesn't have a boolean preset https://github.com/ckan/ckanext-scheming#preset and the value column in the `group_extra` EAV table is a `text` column. I guess doing this essentially serializes to a string on the way in and de serializes to a bool on the way out? Can you just flesh this out slightly more for me, or point me at an example or something?

amercader: Yes, that's it. All extras are stored as a text blob on the db but the values are serialized / deserialized on the way in and out. So at the API JSON object level it's a boolean. I've definitely used this in datasets, and I'd expect orgs to behave in the same way

chris48s:
Note for future Chris:

⚠️ This is not complete.
6cbbd6d updates the schema/form code and works there but I've not updated/tested anything else in this PR.

TODO: Go back over all the other stuff in this PR and work out whether you're supposed to be setting 

```py
visible_external = True   # bool
visible_external = 'true' # str
```

in each of those different contexts + test/check and update accordingly.

chris48s:
Now #403 is in and I've rebased this onto it, I need to have a look at https://github.com/okfn/ckanext-unhcr/pull/403#discussion_r479171220 next. I think I need to get this one finished before we can complete the signup form work as we need the `visible_external` flag to correctly populate the container list on the signup form.

chris48s:
I've gone back to this and rebased it.
The only reason we needed to allow `group_list` was to call `group_list_authz` so I think the answer there is to revoke `group_list` and explicitly allow `group_list_authz` (instead of delegating to `group_list` for auth) - ec594e1 .

I think `organization_list_for_user` is used in so many places we just need to allow it and filtering what it returns is probably going to have unwanted consequences.

We're still waiting on what containers are going to be in the list to populate the CSV, but its probably most useful to review this so we can continue to develop other stuff that relies on the feature.

ckanext/unhcr/templates/scheming/form_snippets/visible_external.html
```diff
@@ -0,0 +1,9 @@
+<div class="form-group control-ful">
+  <label for="field-visible_external" class="control-label">Visible to External users</label>
+  <div class="controls">
+    <select id="field-visible_external" name="visible_external"  class="form-control">
+      <option value="true" {% if data.visible_external %}selected="selected"{% endif %}>Yes</option>
+      <option value="false" {% if not(data.visible_external) %}selected="selected"{% endif %}>No</option>
```
amercader, line 6: Does this evaluate to boolean `False`? I seem to recall I had to do `data.visible_external is sameas false` on another project but maybe that was unnecessary

ckanext/unhcr/helpers.py
```diff
@@ -51,14 +51,19 @@ def get_data_container(id):
...
     context = {'model': model, 'ignore_auth': True}
     orgs = toolkit.get_action('organization_list')(context,
-        {'type': 'data-container', 'all_fields': True})
+        {'type': 'data-container', 'all_fields': True, 'include_extras': True})
     for org in orgs:
-        if org['id'] not in exclude_ids:
-            data_containers.append(org)
+        if org['id'] in exclude_ids:
+            continue
+        if external_user and not org['visible_external']:
```
amercader, line 62: I get a `KeyError` here after updating to this branch:

```
File '/srv/app/src_extensions/ckanext-unhcr/ckanext/unhcr/templates/organization/snippets/curation_modals.html', line 26 in top-level template code
  {% set orgs = h.get_all_data_containers(
File '/srv/app/src_extensions/ckanext-unhcr/ckanext/unhcr/helpers.py', line 62 in get_all_data_containers
  if external_user and not org['visible_external']:
KeyError: 'visible_external'

```
I guess that's because I need to run the scripts that updates all data containers with this new field, but to make the transition easier and avoid exceptions I think we can assume a container is not visible to external users if the field is missing (ie default all existing containers to internal)

ckanext/unhcr/helpers.py
```diff
@@ -51,14 +51,19 @@ def get_data_container(id):
     return toolkit.get_action('organization_show')(context, {'id': id})
 
 
-def get_all_data_containers(exclude_ids=[], include_unknown=False):
+def get_all_data_containers(exclude_ids=[], include_unknown=False, external_user=True):
     data_containers = []
     context = {'model': model, 'ignore_auth': True}
     orgs = toolkit.get_action('organization_list')(context,
```
amercader, line 57: I sound like a broken record but this is slow, specially with `include_extras`. We'll optimize it in the future with a custom action.

amercader (commented):
Looks good @chris48s just one minor comment

ckanext/unhcr/helpers.py
chris48s: I've made a separate issue for it https://github.com/okfn/ckanext-unhcr/issues/418

ckanext/unhcr/helpers.py
chris48s: Yes good catch. Key not set or False should be treated the same - updated.

ckanext/unhcr/templates/scheming/form_snippets/visible_external.html
chris48s: Seems to work as expected.

If I explicitly set it to No I get `<option value="false" selected="selected">No</option>`
If the value is not set at all I get `<option value="false" selected="selected">No</option>`

If I explicitly set it to Yes I get `<option value="true" selected="selected">Yes</option>`

chris48s:
Rebased again and updated (to account for #414 ). c3ef403 and on are the new commits since the last review

amercader (approved):

PR #398, opened 2020-07-29
always link to user guide in the sidebar

### LINKED ISSUE #380: User guide
Make the link to the user guide at im.unhcr.org/ridl available on the interface of RIDL

chris48s:
following on from additional conversation on https://github.com/okfn/ckanext-unhcr/issues/380#issuecomment-664465105

- always link to user guide in the sidebar
- link direct to data deposit docs on data deposit pages

closes #380

### DIFF
--- a/ckanext/unhcr/templates/organization/edit_base.html
+++ b/ckanext/unhcr/templates/organization/edit_base.html
@@ -8,14 +8,22 @@
   {% endif %}
 {% endblock %}
 
-{% block secondary_content %}
-  {% if c.group_dict.id == deposit.id %}
-    {% snippet "organization/snippets/curation_sidebar.html", deposit=deposit, facet_titles=c.facet_titles, editing=True %}
-  {% elif c.group_dict.state == 'approval_needed' %}
-    {% snippet "organization/snippets/pending_sidebar.html", group_dict=group_dict %}
-  {% else %}
-    {{ super() }}
-  {% endif %}
+{% block secondary %}
+<aside class="secondary col-sm-3">
+  {% block secondary_content %}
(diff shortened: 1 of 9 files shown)


ckanext/unhcr/templates/organization/edit_base.html
```diff
@@ -8,14 +8,26 @@
...
-{% block secondary_content %}
-  {% if c.group_dict.id == deposit.id %}
-    {% snippet "organization/snippets/curation_sidebar.html", deposit=deposit, facet_titles=c.facet_titles, editing=True %}
-  {% elif c.group_dict.state == 'approval_needed' %}
-    {% snippet "organization/snippets/pending_sidebar.html", group_dict=group_dict %}
-  {% else %}
-    {{ super() }}
-  {% endif %}
+{% block secondary %}
+<aside class="secondary col-sm-3">
```
amercader, line 12: Maybe we can have one single `user_guide.html` snippet that takes care of checking if it's a deposit or no, and then just call the following from everywhere?
```
  {% block user_guide %}
    {% snippet "snippets/user_guide.html" %}
  {% enblock %}
```
Or if you want the check in the parent template something like:

```
  {% block user_guide %}
    {% snippet "snippets/user_guide.html", is_deposit=c.group_dict.id == deposit.id %}
  {% enblock %}
```

chris48s: Yeah that's tidier :+1: I think we do need to pass in `is_deposit` because sometimes the condition is `is_deposit=c.group_dict.id == deposit.id` and sometimes `is_deposit=pkg.type == 'deposited-dataset'` depending on the template.

Issue #372, opened 2020-07-08
Ensure all links from data-deposit list stay in the data deposit sandbox


chris48s:
At the moment, on https://ridl.unhcr.org/data-container/data-deposit
dataset titles link to
https://ridl.unhcr.org/dataset/foobar
instead of
https://ridl.unhcr.org/deposited-dataset/foobar

Currently, this doesn't matter because all users can view datasets. Once there is a class of user who can't view datasets, this will become important. While you're looking at this, review other links from the data deposits page(s). If there are necessarily links from the data deposit pages that external users will not have, hide them from this class of users.

PR #400, opened 2020-08-03
Ensure external users aren't directed from /data-deposit to /dataset

chris48s:
Refs #372

This branch is based off the work in https://github.com/okfn/ckanext-unhcr/pull/389 so 1b56308 - 978c873 are from that branch/PR. Only f38fe3b - 501c26a are new here.
Review of this is basically irrelevant until #389 is reviewed but I don't want to just keep blocking all progress on the EDD. Once that is sorted, I'll rebase this.

Because `url_for` can deal with a variety of different input styles, I decided to resolve the URL first and then manipulate it as a post-process rather than trying to work with the input params.

### DIFF
--- a/ckanext/unhcr/helpers.py
+++ b/ckanext/unhcr/helpers.py
@@ -745,7 +745,7 @@ def current_path(action=None):
     path = toolkit.request.path
     if action == '/dataset/new':
         path = '/dataset/new'
-    if path.startswith('/dataset/copy'):
+    if path.startswith('/dataset/copy') or path.startswith('/deposited-dataset/copy'):
         path = '/dataset/new'
     return path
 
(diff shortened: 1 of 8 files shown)


amercader:
patching `url_for` sounds like a good compromise until we upgrade to 2.9, where all this is much better supported.

Do we need the monkey-patch in `plugin.py` *and* the helper override?

chris48s:
Yeah unfortunately it does seem to be necessary to double-patch it. I guess when we use `h.url_for()` in a HTML template the chain of imports is different than when we use `toolkit.url_for()` in a python file.

That said, in going back and revisiting this I've realised we can simplify the override in `helpers.py` to just

```py
@core_helpers.core_helper
def url_for(*args, **kw):
    return core_helpers.url_for(*args, **kw)
```

because by the point we register the helper, we've already patched `core_helpers.url_for`. We still have to override it in two places, but at least we only need to implement the logic once. In some ways this actually makes flow _more_ confusing though. Good news that CKAN 2.9 might make this a bit less painful :)

chris48s:
also merging this one to `external-data-deposit`

Issue #375, opened 2020-07-08
Set up access controls for external users


chris48s:
EDD users will need to be able to:

**Data Deposit:**
- [x] Create datasets, but only deposited datasets (external) - validator on the type field?
- [x] Show their own deposited datasets: I think the current perms cover this, but we need to write new tests for external uses
- [x] List/Search their own deposited datasets: I think the current perms cover this, but we need to write new tests for external uses
- [x] Update their own deposited datasets: I think the current perms cover this, but we need to write new tests for external uses
- [x] Delete their own deposited datasets?

**User management:**
- [x] View/Change their own details
- [x] Reset password
- [x] Reset API key
- [x] Login
- [x] Logout

chris48s:
https://github.com/okfn/ckanext-unhcr/pull/389#discussion_r455897875
https://github.com/okfn/ckanext-unhcr/pull/389#discussion_r455898385
https://github.com/okfn/ckanext-unhcr/pull/389#discussion_r455898708

PR #403, opened 2020-08-28
Set up access controls for external users

chris48s:
Refs #375

In this PR, I have focussed on granting external users a minimal set of authorizations necessary to create deposited datasets and interact with their own datasets in the data deposit container.

I haven't attempted to do anything with the interface for accomplishing this task - I'm just focussing on permissions here. I also haven't done anything about the user management permissions yet. I think this is probably enough to review at one time.

Note this PR is targetted at the long-running `external-data-deposit` branch (not `master`) which I've already merged #389 and #400 onto but it won't affect what we're shipping to users yet.

This shouldn't be too hard to review hopefully - I've split it into lots of commits making small changes with plenty of explanation.

### DIFF
--- a/ckanext/unhcr/auth.py
+++ b/ckanext/unhcr/auth.py
@@ -166,7 +166,7 @@ def package_update(next_auth, context, data_dict):
     # Deposited dataset
     if dataset['type'] == 'deposited-dataset':
         curation = helpers.get_deposited_dataset_user_curation_status(
-            dataset, toolkit.c.userobj.id)
+            dataset, getattr(context.get('auth_user_obj'), 'id', None))
         if 'edit' in curation['actions']:
             return {'success': True}
         return {'success': False, 'msg': 'Not authorized to edit deposited dataset'}
--- a/ckanext/unhcr/auth.py
+++ b/ckanext/unhcr/auth.py
@@ -77,6 +77,20 @@ def organization_list_for_user(context, data_dict):
         return core_get.organization_list_for_user(context, data_dict)
 
 
+@toolkit.chained_auth_function
+def organization_show(next_auth, context, data_dict):
+    user = context.get('auth_user_obj')
+    if not user:
+        return next_auth(context, data_dict)
+    if user.external:
+        deposit = helpers.get_data_deposit()
+        if data_dict.get('id') in [deposit['name'], deposit['id']]:
+            return {'success': True}
+        else:
+            return {'success': False}
+    return next_auth(context, data_dict)
+
+
 def organization_create(context, data_dict):
     user_orgs = toolkit.get_action('organization_list_for_user')(context, {})
 
--- a/ckanext/unhcr/auth.py
+++ b/ckanext/unhcr/auth.py
@@ -128,18 +128,27 @@ def organization_create(context, data_dict):
 # Package
 
 def package_create(context, data_dict):
-
     # Data deposit
     if not data_dict:
-        # All users can deposit datasets
-        if toolkit.request.path == '/deposited-dataset/new':
-            return {'success': True}
+        try:
+            # All users can deposit datasets
+            if (
+                toolkit.request.path == '/deposited-dataset/new' or
+                toolkit.request.path.startswith('/deposited-dataset/edit/')
+            ):
+                return {'success': True}
+        except TypeError:
+            return {
+                'success': False,
+                'msg': 'package_create requires either a web request or a data_dict'
+            }
     else:
         deposit = helpers.get_data_deposit()
         if deposit['id'] == data_dict.get('owner_org'):
             return {'success': True}
 
     # Data container
+    context['model'] = context.get('model') or model
     return auth_create_core.package_create(context, data_dict)
 
 
--- a/ckanext/unhcr/plugin.py
+++ b/ckanext/unhcr/plugin.py
@@ -29,6 +29,27 @@ _ = toolkit._
 
 
 INTERNAL_DOMAINS = ['unhcr.org']
+ALLOWED_ACTIONS = [
+    'group_list_authz',
+    'group_list',
+    'group_show',
+    'organization_list_for_user',
+    'organization_show',
+    'package_create',
+    'package_delete',
+    'package_patch',
+    'package_resource_reorder',
+    'package_search',
+    'package_show',
+    'package_update',
+    'resource_create',
+    'resource_delete',
+    'resource_download',
+    'resource_patch',
+    'resource_show',
+    'resource_update',
+    'resource_view_list',
+]
 
 
 def user_is_external(user):
@@ -56,11 +77,6 @@ def restrict_external(func):
     '''
     Decorator function to restrict external users to a small number of allowed_actions
     '''
-
-    allowed_actions = [
-        'package_search',
-    ]
-
     def wrapper(action, context, data_dict=None):
         user = User.by_name(context.get('user'))
         if not user:
@@ -69,7 +85,7 @@ def restrict_external(func):
             return func(action, context, data_dict)
         if context.get('ignore_auth'):
             return func(action, context, data_dict)
-        if user.external and action not in allowed_actions:
+        if user.external and action not in ALLOWED_ACTIONS:
             return {'success': False, 'msg': 'Not allowed to perform this action'}
         return func(action, context, data_dict)
     return wrapper
--- a/ckanext/unhcr/plugin.py
+++ b/ckanext/unhcr/plugin.py
@@ -489,6 +489,7 @@ class UnhcrPlugin(
         functions['unhcr_datastore_search_sql'] = auth.unhcr_datastore_search_sql
         functions['datasets_validation_report'] = auth.datasets_validation_report
         functions['organization_create'] = auth.organization_create
+        functions['organization_show'] = auth.organization_show
         functions['package_activity_list'] = auth.package_activity_list
         functions['package_create'] = auth.package_create
         functions['package_update'] = auth.package_update
--- a/ckanext/unhcr/plugin.py
+++ b/ckanext/unhcr/plugin.py
@@ -582,7 +582,7 @@ class UnhcrPlugin(
         if user_obj:
 
             if user_obj.external:
-                return ['']
+                return ['creator-%s' % user_obj.id]
 
             context = {u'user': user_obj.id}
             deposit = helpers.get_data_deposit()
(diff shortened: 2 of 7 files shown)


ckanext/unhcr/plugin.py
```diff
@@ -29,6 +29,25 @@
...
+    'package_resource_reorder',
+    'package_search',
+    'package_show',
+    'package_update',
+    'resource_create',
+    'resource_download',
+    'resource_patch',
+    'resource_show',
+    'resource_update',
+    'resource_view_list',
```
chris48s, line 49: I haven't granted `package_delete` or `resource_delete` here (yet). I know we were slightly undecided on this. Tbh, I don't really see a problem with allowing users to delete as long as the permission labels are doing their job correctly. Shall we just grant them?

ckanext/unhcr/auth.py
```diff
@@ -114,11 +127,13 @@ def organization_create(context, data_dict):
...
 def package_create(context, data_dict):
-
     # Data deposit
     if not data_dict:
         # All users can deposit datasets
-        if toolkit.request.path == '/deposited-dataset/new':
+        if (
+            toolkit.request.path == '/deposited-dataset/new' or
+            toolkit.request.path.startswith('/deposited-dataset/edit/')
+        ):
```
chris48s, line 136: TODO: write a test for this.
This isn't really my favourite ever solution to this, but I can't see another way given `data_dict` is empty in this case. Any thoughts of another way around it? I haven't written an auth test for `package_create` yet because its hard to write a unit test for the auth function when it depends on the request like this. If we can't find another way around it, I'll revisit. Maybe it is best implemented as a controller test.

ckanext/unhcr/plugin.py
```diff
@@ -29,6 +29,25 @@
 
 
 INTERNAL_DOMAINS = ['unhcr.org']
+ALLOWED_ACTIONS = [
+    'group_list_authz',
+    'group_list',
+    'group_show',
+    'organization_list_for_user',
```
chris48s, line 36: Granting `group_list` and `organization_list_for_user` here is a bit too lax really, but I'm going to suggest we just do it like this for now to avoid making this PR too big by merging it with #394. Once we merge this PR to the `external-data-deposit` branch, I will rebase PR #394 onto this and tighten things up there. I think the correct solution is probably to move what I've done in a template helper in https://github.com/okfn/ckanext-unhcr/pull/394/commits/1b4b550cb54b94f4dbdee8aec585bb6de0bb55d7 up a level into the action/auth function.

ckanext/unhcr/plugin.py
amercader: Yes, I think it makes sense. The standard auth should ensure that external users can only delete their own ones

ckanext/unhcr/tests/test_auth.py
```diff
@@ -292,78 +301,308 @@ def test_external_users_core_actions(self):
...
+            owner_org='data-deposit',
+            owner_org_dest='data-target',
+            user=user
+        )
+        dataset = factories.Dataset(
+            owner_org=target['id'],
+            visibility='private'
+        )
+
+        resource_created_by_external_user = factories.Resource(
```
amercader, line 420: If you really want to ensure that the resource is created by an external user you need to pass `user=external_user` with the init params. Otherwise an internal sysadmin is used.
Same for the `someone_else` below

ckanext/unhcr/templates/package/read_base.html
```diff
@@ -14,11 +14,16 @@
 {% block content_primary_nav %}
   {% if pkg.type == 'deposited-dataset' %}
     {{ h.build_nav_icon('dataset_read', _('Dataset'), id=pkg.name) }}
-  	{{ h.build_nav_icon('%s_internal_activity' % dataset_type, _('Internal Activity'), dataset_id=pkg.name, icon='gavel') }}
+    {% if h.check_access('package_activity_list', {'user': c.user, 'id': c.pkg_dict.id}) %}
```
amercader, line 17: CKAN is moving towards not relying on `c` vars on templates, so unless it's something like `c.user` we should use `pkg.id` (or whatever var is actually passed to templates) to make easier future migrations

amercader (commented):
Great stuff @chris48s , I left some comments

ckanext/unhcr/templates/package/read_base.html
chris48s: 45ff190

ckanext/unhcr/auth.py
```diff
@@ -77,6 +77,19 @@ def organization_list_for_user(context, data_dict):
         return core_get.organization_list_for_user(context, data_dict)
 
 
+def organization_show(context, data_dict):
```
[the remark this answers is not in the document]
chris48s: 3501055

ckanext/unhcr/plugin.py
chris48s: 96e95af

ckanext/unhcr/tests/test_auth.py
chris48s: This is partly about bad variable naming really. I think by trying to make the variable names super-clear, I actually made things more confusing :upside_down_face:   Maybe I should have just called them `resource1` and `resource2`. :trollface: 

That said, this does raise an interesting point. As things stand, if `external_user` created the (deposited) dataset, they should be able to access any resources attached to that dataset, even if the resource itself was created by someone else (although this would be unusual, I guess). i.e: its important who created the dataset, but not important who created the resources attached to it. So I think the test was right (i.e: even if sysadmin created the resource on `deposited_dataset_created_by_external_user`, `external_user` can still work on it) That's how I understand the requirements anyway, but that wasn't really explicit in the test.

I've gone back and reworked this whole thing. I've tried to make all of the above explicit without having variables called `resource_attached_to_deposited_dataset_created_by_external_user`, but also that test was really testing at least 3 things, so I turned it into a class with 3 test methods. Hopefully this is better: 6e1a3ce

ckanext/unhcr/auth.py
chris48s: Unfortunately using `toolkit.get_endpoint()` doesn't help in this situation because it returns `(u'package', u'edit')` regardless of whether we are on `deposited-dataset/edit` or `dataset/edit`, which is the distinction we are trying to check for here..

In terms of writing a test for it, I'm hitting 2 issues:

1. The auth check for `package_create` behaves differently under test than it does interactively. This is because `ckan.auth.create_unowned_dataset` is `False` in dev but `True` under test. I guess that is because of https://github.com/ckan/ckan/blob/ckan-2.8.4/test-core.ini#L41 but I'm not sure why that is on under test by default.

If I just accept that's the way it is, wrap the test in `@helpers.change_config('ckan.auth.create_unowned_dataset', False)` and move on, the next problem I hit is:

2. We're still using pylons controllers in this bit of the codebase (because ckan 2.8) but nevertheless wrapping the assertion in `flask_app.test_request_context()` does _sort of_ work here (There doesn't seem to be a equivalent concept for `pylons_app` AFAICT?). The URL is set when we test `toolkit.request.path` in `package_create` but then we end up throwing `AttributeError: '_Globals' object has no attribute 'userobj'` here: https://github.com/ckan/ckan/blob/ckan-2.8.4/ckan/authz.py#L384 I'm unclear if `c` isn't populated correctly because `flask_app.test_request_context()` is creating the context for a flask request but then I'm calling a plyons controller (which would be fair enough) or if its a bug in core and we should be catching the `AttributeError` error in core as well as the `TypeError` to determine `c` is not set. I'm a bit out of my depth when it comes to the knowledge of internals at this point.

Either way, I've still not written a test for it and I'm giving up for the afternoon. (╯°□°)╯︵ ┻━┻

chris48s: I've decided to go outwards a layer and write integration tests for this, in lieu of a better option: d028f70

While I was doing this I realised that the other consequence of what we're doing here is that the `package_create` action has different behaviour if you call it from the web UI than it does if you call it via the API. Guess there isn't much we can do about that either :shrug:

amercader (approved):
Looks great @chris48s

ckanext/unhcr/auth.py
amercader: on 1:

Having a global auth setting in the core test config that is not the default is a bug IMO as you can't be sure you are testing the expected behaviour. I'll raise the issue but I guess changing it in core would involve broken tests and a bit of work. But in our case we should definitely have the same auth config in tests than live. Rather than use the `change_config` decorator let's set the appropriate settings in `test.ini` that mimic what the live envs use, ie our [custom settings](https://github.com/okfn/docker-ckan-unhcr-aws/blob/master/ckan/environment/base.env#L33) on top the [CKAN defaults](https://github.com/ckan/ckan/blob/ckan-2.8.4/ckan/config/deployment.ini_tmpl#L65). This way we can are sure that all tests (specially the auth ones) use a consistent config.

let me get back to you on 2

amercader: > While I was doing this I realised that the other consequence of what we're doing here is that the package_create action has different behaviour if you call it from the web UI than it does if you call it via the API. Guess there isn't much we can do about that either shrug

This is bad. But I guess that when calling it via the api, at the point the `package_create` auth function is called `data_dict` is populated and you can check the `type` right?

amercader: On 2, honestly integration tests might be the right call if we need to rely on a web request. `userobj` might be not set because the request stack in CKAN that indentifies the user is not called. Core tests that use this use a [test client web request](https://github.com/ckan/ckan/blob/ckan-2.8.4/ckan/tests/controllers/test_group.py#L425:L435) anyway

But the auth function should work as well outside the context of a web request and we should have tests for that, (ie calling the action directly) with something like:

```python

try:
    if toolkit.request.path
except XXX
    # if no web request you need to pass a data_dict so we can check the type
```

chris48s: OK then. I tried setting `ckan.auth.create_unowned_dataset = false` in our test settings for this project and did a test run just to see how many tests would fail. It was 29 which isn't an outrageous number of tests to update, but I think its too much scope screep for this PR. I'm going to raise an issue to fix/update that and just leave the 4 `@helpers.change_config('ckan.auth.create_unowned_dataset', False)` decorators for now.

In terms of the auth function working outside of the web ui, if I call check `package_create' with `data_dict == None` and no web request, should the fallback be that we throw an exception (a `toolkit.Invalid` or something?) or just `return {'success': False}`?

chris48s: I think I've kind of answered my own question on this one and decided that it makes more sense to `return {'success': False}` as either way we're not going to let the user do the thing so we should probably just return a normal value.

I've pushed ce09efa which improves this and adds more tests. I think this is probably looking fairly sensible now.
I've also raised #411 to deal with another day.

amercader: Yes, auth functions should return `{'success': True / False}`. +1 on address the test config later

PR #409, opened 2020-09-01
style Data-container as Data Container

### LINKED ISSUE #406: Pop-up text after data container deletion
Could you please remove the hyphen from 'data-container' and capitalize the C? Thanks!
image

chris48s:
Closes #406

This gives us a a workaround, for this:
https://github.com/ckan/ckan/blob/7900935de6270ee80ec64a5378251dafedf3a0e4/ckan/controllers/group.py#L626-L627
Do we actually store a 'name' for this at the CKAN level, or is the slug `data-container` the only thing we've got in the DB?


amercader (approved):
There's no name for this. It's just the value of the `type` field

Issue #378, opened 2020-07-08
Allow external users to create accounts


chris48s:
While we're getting started, we can create service accounts on the console

The requirements for this may still change, but at the moment, it looks like to launch, we will need to do something like:

- Create/expose a signup form allowing anyone to request an account (at the moment https://ridl.unhcr.org/user/register is blocked)
- Apply some kind of captcha/bot-proofing?
- Prevent UNHCR users from creating non-SAML accounts

When a user requests an account:

- We create the user as 'pending' and email an appropriate admin/internal user
- The internal user can approve or reject the account
- The requesting user gets an email with the result. If approved, they can log in

chris48s:
Requirements as described in the top post are OK

chris48s:
Did a bit of initial investigation on this..

First thing we'll need to do for this one is

- merge https://github.com/okfn/ckanext-saml2/pull/1
- tag a release
- upgrade that plugin in the docker repo

we can do that independently of this feature as that PR is just a bugfix to the plugin.

Once we've got that, we'll also need to set

`CKAN__AUTH__CREATE_USER_VIA_WEB=True`

on the docker repo. That does introduce meaningful behaviour change, so that will need to sit as a WIP on the docker repo until we're ready to deploy `external-data-deposit` but future work on that branch will depend on this setting.

Then we'll need to modify `page.html` so it doesn't call `{% snippet "login_form.html" %}` if we're logged out and we hit on `/user/register`

Once we've done that, I think that enables external signups (without any of the customisations/approval steps)

chris48s:
SAML2 upgrade done in https://github.com/okfn/docker-ckan-unhcr-aws/pull/14
WIP PR to change `CKAN__AUTH__CREATE_USER_VIA_WEB` at https://github.com/okfn/docker-ckan-unhcr-aws/pull/15

chris48s:
Do we want to customise the behaviour of `user/register` for already logged-in users?

![Screenshot at 2020-09-09 09-09-35](https://user-images.githubusercontent.com/6025893/92572334-335fbc00-f27c-11ea-9e5d-4350c32f2d8a.png)

chris48s:
Last remaining issue moved to #417

Issue #412, opened 2020-09-11
visually distinguish external users on /user


chris48s:
Once we settle on a style in #410 lets use the same badge/label to identify clearly external users on /user

chris48s:
related: maybe also make state clearer/more obvious on user detail page?

![Screenshot at 2020-09-21 17-00-23](https://user-images.githubusercontent.com/6025893/93791336-015f3a00-fc2c-11ea-83e4-dc568282c14c.png)

Issue #377, opened 2020-07-08
Create UI for EDD users


chris48s:
Once an external user has an account, they can log in ad see..

- ~Landing page?~
- ~Cut down version of https://ridl.unhcr.org/data-container/data-deposit~
- Replace filters/left panel conent with some kind of explanation of
    - What is this thing?
    - What does it do?
    - How do I use it?
- ~Remove all menu items linking to stuff they can't access~
- Slightly different headers/branding
- Some subset of user management pages (some customisation needed?)

PR #414, opened 2020-09-11
External User Signup process

chris48s:
Refs #378
Refs https://github.com/okfn/docker-ckan-unhcr-aws/pull/15

**Progress:**

- [x] Create customised `RegisterView`
- [x] Don't allow internal users to create an external user account
- [x] Create new users with pending status ( `ckan.model.State.PENDING` )
- [x] Drop new users to a confirmation/pending approval page after signup. Don't try to log them in
- [x] Collect 2 extra fields on signup:
    - Free text box: who are you/why do you want an account?
    - Containers: pick a/some containers relevant to your work
- [x] Better text/copy (in progress: following up with client)
    - What are we calling external accounts?
    - Better labels for message/container fields
    - What's the CTA for internal users trying to create an external account?
    - Sidebar content, some kind of explanation of what this is (it will be the first page a new external user sees)
- [x] Use message/containers to notify appropriate container admins to approve the request
- [ ] Enable Captcha (CKAN core already suports this - think we just need to set `ckan.recaptcha.publickey` and `ckan.recaptcha.privatekey`: https://docs.ckan.org/en/2.8/maintaining/configuration.html?highlight=recaptcha#ckan-recaptcha-publickey ) - TODO: deploy repo
- [x] Interface to allow admin users to approve/reject account requests, confirmation email template
    - [x] Add a user value to the object enum + migration
    - [x] Add an optional data:JSONB column + migration
    - [x] Modify `RegisterView.post()` to create an access request with `access_request_create`
    - [x] Surface user AccessRequests through the front-end ( `ExtendedUser.list_requests` / `dashboard_requests.html` )
    - [x] Modify `access_request_update` action/auth functions so we can approve/reject user account requests
        - approve moves the status from pending to active
        - reject moves the status from pending to deleted
    - [x] Emails for approved/rejected requests
    - [x] Automated test for all that
- [x] Automated tests (the CKAN core tests for `RegisterView` will be a good start for some of this as the new `RegisterView` replicates a lot of core functionality) but there will be more to do. So far I have writen zero tests
- [x] What should the behaviour be if you hit `/user/register` and you're already logged in?

### DIFF
--- a/ckanext/unhcr/auth.py
+++ b/ckanext/unhcr/auth.py
@@ -306,20 +306,51 @@ def access_request_update(context, data_dict):
             context, {'id': request.object_id}
         )
         org_id = package['owner_org']
+        return {
+            'success': has_user_permission_for_group_or_org(
+                org_id, user, 'admin'
+            )
+        }
     elif request.object_type == 'organization':
         org_id = request.object_id
-    else:
-        raise toolkit.Invalid("Unknown Object Type")
+        return {
+            'success': has_user_permission_for_group_or_org(
+                org_id, user, 'admin'
+            )
+        }
+    elif request.object_type == 'user':
+        return external_user_update_state(context, {'id': request.object_id})
 
-    return {
-        'success': has_user_permission_for_group_or_org(
-            org_id, user, 'admin'
-        )
-    }
+    raise toolkit.Invalid("Unknown Object Type")
 
 
 def access_request_create(context, data_dict):
     return {'success': bool(context.get('user'))}
+
+
+def external_user_update_state(context, data_dict):
+    m = context.get('model', model)
+    request_userobj = context.get('auth_user_obj')
+    if not request_userobj:
+        return {'success': False}
+
+    target_user_id = toolkit.get_or_bust(data_dict, "id")
+    target_userobj = m.User.get(target_user_id)
+    if not target_userobj:
+        raise toolkit.ObjectNotFound("User not found")
+
+    # request_userobj is the user who is trying to perform the action
+    # target_userobj is the user we're trying to modify
+
+    if not target_userobj.external:
+        return {'success': False, 'msg': "Can only perform this action on an external user"}
+    if target_userobj.state != m.State.PENDING:
+        return {'success': False, 'msg': "Can only change state of a 'pending' user"}
+
+    if not helpers.user_is_container_admin(request_userobj.id):
+        return {'success': False, 'msg': "Must be admin of at least one container"}
+
+    return {'success': True}
 
 
 # Admin
--- a/ckanext/unhcr/auth.py
+++ b/ckanext/unhcr/auth.py
@@ -347,10 +347,26 @@ def external_user_update_state(context, data_dict):
     if target_userobj.state != m.State.PENDING:
         return {'success': False, 'msg': "Can only change state of a 'pending' user"}
 
-    if not helpers.user_is_container_admin(request_userobj.id):
-        return {'success': False, 'msg': "Must be admin of at least one container"}
+    access_requests = model.Session.query(AccessRequest).filter(
+        AccessRequest.user_id==target_userobj.id,
+        AccessRequest.object_id==target_userobj.id,
+        AccessRequest.status=='requested',
+        AccessRequest.object_type=='user',
+    ).all()
 
-    return {'success': True}
+    if not access_requests or len(access_requests) > 1:
+        return {
+            'success': False,
+            'msg': "User must be associated with exactly one pending access request"
+        }
+
+    for container in access_requests[0].data['containers']:
+        if has_user_permission_for_group_or_org(
+            container, request_userobj.id, 'admin'
+        ):
+            return {'success': True}
+
+    return {'success': False}
 
 
 # Admin
--- a/ckanext/unhcr/blueprints/user.py
+++ b/ckanext/unhcr/blueprints/user.py
@@ -49,7 +49,7 @@ class RegisterView(BaseRegisterView):
             context, {'type': 'data-container', 'all_fields': True}
         )
         deposit = get_data_deposit()
-        return sorted(
+        containers = sorted(
             [
                 {'value': o['id'], 'text': o['display_name'] }
                 for o in orgs
@@ -58,6 +58,8 @@ class RegisterView(BaseRegisterView):
             ],
             key=lambda o: o['text']
         )
+        containers.insert(0, {'value': '', 'text': 'Select...'})
+        return containers
 
     def post(self):
         context = self._prepare()
@@ -93,17 +95,44 @@ class RegisterView(BaseRegisterView):
             )
             return self.get(data_dict, {'email': [message]}, {'email': message})
 
+        if not data_dict.get('container'):
+            message = "A region must be specified"
+            return self.get(data_dict, {'container': [message]}, {'container': message})
+
+        context['defer_commit'] = True
+        data_dict['state'] = context['model'].State.PENDING
+
         try:
-            data_dict['state'] = context['model'].State.PENDING
-            toolkit.get_action(u'user_create')(context, data_dict)
+            model.Session.begin_nested()
+            user = toolkit.get_action(u'user_create')(context, data_dict)
+
+            access_request_data_dict = {
+                'object_id': user['id'],
+                'object_type': 'user',
+                'message': data_dict['message'],
+                'role': 'member',
+                'data': {'containers': [data_dict.get('container')]}
+            }
+            toolkit.get_action(u'access_request_create')(
+                {'user': user['id'], 'ignore_auth': True, 'defer_commit': True},
+                access_request_data_dict
+            )
+
+            model.Session.commit()
         except toolkit.NotAuthorized:
+            model.Session.rollback()
             toolkit.abort(403, _(u'Unauthorized to create user %s') % u'')
         except toolkit.ObjectNotFound:
+            model.Session.rollback()
             toolkit.abort(404, _(u'User not found'))
         except toolkit.ValidationError as e:
+            model.Session.rollback()
             errors = e.error_dict
             error_summary = e.error_summary
             return self.get(data_dict, errors, error_summary)
+        except:
+            model.Session.rollback()
+            raise
 
         if toolkit.c.user:
             # #1799 User has managed to register whilst logged in - warn user
--- a/ckanext/unhcr/blueprints/user.py
+++ b/ckanext/unhcr/blueprints/user.py
@@ -102,6 +102,8 @@ class RegisterView(BaseRegisterView):
 
         context['defer_commit'] = True
         data_dict['state'] = context['model'].State.PENDING
+        deposit = get_data_deposit()
+        containers = [data_dict.get('container'), deposit['id']]
 
         try:
             model.Session.begin_nested()
@@ -112,7 +114,7 @@ class RegisterView(BaseRegisterView):
                 'object_type': 'user',
                 'message': data_dict['message'],
                 'role': 'member',
-                'data': {'containers': [data_dict.get('container')]}
+                'data': {'containers': containers}
             }
             toolkit.get_action(u'access_request_create')(
                 {'user': user['id'], 'ignore_auth': True, 'defer_commit': True},
@@ -136,7 +138,7 @@ class RegisterView(BaseRegisterView):
             raise
 
         recipients = mailer.get_user_account_request_access_email_recipients(
-            [data_dict.get('container')]
+            containers
         )
         for recipient in recipients:
             subj = mailer.compose_user_request_access_email_subj()
--- a/ckanext/unhcr/blueprints/user.py
+++ b/ckanext/unhcr/blueprints/user.py
@@ -2,11 +2,13 @@
 
 import logging
 from flask import Blueprint
+from ckan import model
 import ckan.lib.captcha as captcha
 import ckan.lib.navl.dictization_functions as dictization_functions
 import ckan.logic as logic
 import ckan.plugins.toolkit as toolkit
 from ckan.views.user import RegisterView as BaseRegisterView
+from ckanext.unhcr.helpers import get_data_deposit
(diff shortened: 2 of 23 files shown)


ckanext/unhcr/blueprints/user.py
```diff
@@ -30,8 +42,169 @@ def sysadmin():
...
+                recipient,
+                user,
+                user,
+                data_dict['message']
+            )
+            mailer.mail_user_by_id(recipient['name'], subj, body)
+
+        if toolkit.c.user:
+            # #1799 User has managed to register whilst logged in - warn user
+            # they are not re-logged in as new user.
```
chris48s, line 156: I've copied this behaviour from CKAN core
https://github.com/ckan/ckan/blob/7900935de6270ee80ec64a5378251dafedf3a0e4/ckan/views/user.py#L321-L345
but I wonder if it would make more sense to customise this behaviour for RIDL. If you're already logged in and you have a `@unhcr.org` email, you won't be able to create an account anyway, even if you do log out. Any thoughts on sensible behaviour here?

ckanext/unhcr/blueprints/user.py
```diff
@@ -30,8 +42,169 @@ def sysadmin():
...
+        user_is_sysadmin = False
+        if toolkit.c.user:
+            try:
+                toolkit.check_access('sysadmin', {'user': toolkit.c.user})
+                user_is_sysadmin = True
+            except toolkit.NotAuthorized:
+                pass
+
+        if toolkit.c.user and not data and not user_is_sysadmin:
+            # #1799 Don't offer the registration form if already logged in
```
chris48s, line 184: Comment also applies here - copied from CKAN core. Should we customise?

ckanext/unhcr/auth.py
```diff
@@ -306,22 +306,53 @@ def access_request_update(context, data_dict):
...
+    # request_userobj is the user who is trying to perform the action
+    # target_userobj is the user we're trying to modify
+
+    if not target_userobj.external:
+        return {'success': False, 'msg': "Can only perform this action on an external user"}
+    if target_userobj.state != m.State.PENDING:
+        return {'success': False, 'msg': "Can only change state of a 'pending' user"}
+
+    if not helpers.user_is_container_admin(request_userobj.id):
+        return {'success': False, 'msg': "Must be admin of at least one container"}
```
chris48s, line 351: The way I've done this, we only _show_ the `AccessRequest` objects to the admins of the containers specified at signup time. At this level, I'm only checking the requesting user is an admin of one or more containers to allow them to perform this action on an external user. Its a bit fiddly to tighten this up because the container specified at signup time is attached to an `AccessRequest` object but here its just one user object performing an operation on another user object. The `AccessRequest` is out-of-context at this point. I'm slightly in two minds about what to do with it. Any thoughts on it?

ckanext/unhcr/mailer.py
```diff
@@ -282,22 +299,38 @@ def compose_request_access_email_body(object_type, recipient, obj, requesting_us
...
+
+
+def compose_account_approved_email_subj():
+    return '[UNHCR RIDL] - User account approved'
+
+
+def compose_account_approved_email_body(recipient):
+    context = {}
+    context['recipient'] = recipient
+    context['login_url'] = toolkit.url_for('/service/login', qualified=True)
```
chris48s, line 333: I've hard coded `/service/login` here but its probably not exactly right. I'm struggling to figure out what the abstracted route is here. `toolkit.url_for('user.login')` isn't what we want - that dumps us to the SAML login in prod.

ckanext/unhcr/blueprints/user.py
```diff
@@ -30,8 +42,169 @@ def sysadmin():
...
+        orgs = toolkit.get_action('organization_list')(
+            context, {'type': 'data-container', 'all_fields': True}
+        )
+        deposit = get_data_deposit()
+        containers = sorted(
+            [
+                {'value': o['id'], 'text': o['display_name'] }
+                for o in orgs
+                if o['id'] != deposit['id']
+                # TODO: and o['visible_external']
```
chris48s, line 58: This depends on #394 which is still WIP. For review purposes I'm just listing every container here, but assume we'll reduce this to 7 "region" containers soon.

ckanext/unhcr/blueprints/user.py
```diff
@@ -30,8 +42,169 @@ def sysadmin():
...
+                'Users with an @{domain} email may not register for an external account. '.format(
+                    domain=domain
+                ) + 'Log in to {site} using {email} and use your Active Directory password to access RIDL'.format(
+                    site=toolkit.config.get('ckan.site_url'),
+                    email=toolkit.request.form['email']
+                )
+            )
+            return self.get(data_dict, {'email': [message]}, {'email': message})
+
+        if not data_dict.get('container'):
```
chris48s, line 99: We're only collecting one container here, but all of the internals are able to cope with an array of containers, so its not a huge deal if we need to change this to a multi later. I think there is value in keeping the UI simple for external users though

chris48s:
Right this is finally ready for review. The PR is long, but hopefully each commit is fairly digestible

Just to help with the review, here's a recap of the business requirements for this feature:

- There is a signup page where anyone on the internet can request an account.
- When the user requests an account, we collect their email address, a free text box for more info (who are you?/why do you want this?/etc) and we collect a container they are interested in submitting data to
- We create the user as 'pending' and email sysadmins, curation team and the admins of the container they specified
- Sysadmins, curation team and the admins of the container they specified can approve or reject the account
- The requesting user gets an email with the result. If approved, they can log in

ckanext/unhcr/blueprints/user.py
```diff
@@ -30,8 +42,169 @@ def sysadmin():
     return toolkit.h.redirect_to('unhcr_admin.index')
 
 
+class RegisterView(BaseRegisterView):
+
+    def _get_container_list(self):
+        context = {'model': model, 'ignore_auth': True}
+        orgs = toolkit.get_action('organization_list')(
```
amercader, line 49: Not be solved here but this is really slow and makes a request for a form take longer than it should. We might want to use a static list or config option if there are only 7 options. We can think about it later.

ckanext/unhcr/blueprints/user.py
amercader: I agree

ckanext/unhcr/blueprints/user.py
amercader: This code is core is meant to address sysadmins creating new users directly on the UI. I think you are right in that it doesn't apply here. I guess we could move the check further up and prevent logged-in users (internal and external) to register new accounts. We could probably allow sysadmins though as I can see a use case for that but we can wait to have an actual requirement to do it.

ckanext/unhcr/blueprints/user.py
amercader: See comment above. In this case I think this makes sense as we don't want to allow logged in users to create accounts unless they are syadmins. We may have to review the language on logout_first.html though

ckanext/unhcr/mailer.py
```diff
@@ -252,17 +262,21 @@ def get_container_request_access_email_recipients(container_dict):
...
     return recipients
 
 def get_dataset_request_access_email_recipients(package_dict):
     return get_container_request_access_email_recipients({"id": package_dict["owner_org"]})
 
+def get_user_account_request_access_email_recipients(containers):
+    # This email is sent to admins of all containers in `containers` arg plus sysadmins
+    recipients = _get_sysadmins()
+    for c in containers:
+        recipients = recipients + get_container_request_access_email_recipients({"id": c})
```
amercader, line 276: `c` is loaded word in CKAN land :)

ckanext/unhcr/mailer.py
amercader: `toolkit.url_for('user.login')` seems right no? This is `/user/login`, which when `CKAN___SAML2__ENABLE_NATIVE_LOGIN=True` looks the same as `/service/login` (Login with SSO on top and user name / password on the bottom). I think we need to definitely redesign this page or have a separate one for externals but for now it allows us to validate the flow

ckanext/unhcr/templates/user/new_user_form.html
```diff
@@ -0,0 +1,42 @@
+{% import "macros/form.html" as form %}
```
amercader, line 1: We could definitely use some more blocks on the core template to avoid duplicating the whole thing...
Maybe send a quick PR to core and we can simplify this template when we switch to 2.9

amercader (approved):
All looks good @chris48s , I left some comments. I'll add a new comment with UI stuff that we can do in a separate PR

amercader:
Going through the registration flow I noticed several issues that we need to address at some point to make things clearer for new users. These are probably part of a separate issue, but collecting here for reference:

1. There's no way to know where to register when visiting the site. Perhaps initially we can just point testers to `/user/register` but before lunch we need to have a path for users to find this form. (eg Section on the home page for external users)
2. The login page should make clearer who needs to use what. Most users need the "Login with SSO" link, so the user name and password fields could be hidden or on a separate page
3. We should display a banner on a pending user page, right now is easy to miss the small "State: pending" notice on the left
4. Once logged in as an external user I'm redirected to the homepage, which shows an auth error. We should redirect to the Data Deposit, and hide the buttons from the header that are not relevant (Data Containers and Datasets)

chris48s:
Thanks for the review comments. I will pick up next week. Seems like there are no huge issues though. Just to respond to your comments about next steps..

> 1. There's no way to know where to register when visiting the site. Perhaps initially we can just point testers to /user/register but before lunch we need to have a path for users to find this form. (eg Section on the home page for external users)

> 2. The login page should make clearer who needs to use what. Most users need the "Login with SSO" link, so the user name and password fields could be hidden or on a separate page

Definitely. I think we probably need a home/landing page when you hit https://ridl.unhcr.org with links to funnel users to

* Login (UNHCR users)
* Register (external users)
* Login (external users)

and a bit of text explaining who needs to use what/when. I'll make an issue for this.

> 3. We should display a banner on a pending user page, right now is easy to miss the small "State: pending" notice on the left

Fully agreed :+1:  already noted in https://github.com/okfn/ckanext-unhcr/issues/412#issuecomment-696211114 

> 4. Once logged in as an external user I'm redirected to the homepage, which shows an auth error. We should redirect to the Data Deposit, and hide the buttons from the header that are not relevant (Data Containers and Datasets)

Again, agreed. I think this is covered by https://github.com/okfn/ckanext-unhcr/issues/377 (which is probably the main remaining large issue to address).

ckanext/unhcr/mailer.py
chris48s: e3a033d

ckanext/unhcr/blueprints/user.py
chris48s: > I guess we could move the check further up and prevent logged-in users (internal and external) to register new accounts. We could probably allow sysadmins though as I can see a use case for that but we can wait to have an actual requirement to do it.

OK then. What I've done is just made the behaviour way simpler. If you hit this page and you're already logged in, we throw a 403 error see 6deaecd . If we decide we need to make it more complicated, then we can revisit it. Does that seem reasonable?

ckanext/unhcr/auth.py
chris48s: > can there be multiple access requests for the same target user though?

Good question. As long as we've created it via the action, we can't have more than one open request for the same account
https://github.com/okfn/ckanext-unhcr/blob/db6c40cac04c43d7500fbdef1494aae8b82ce580/ckanext/unhcr/actions.py#L692-L700
so I think its fine to assume we must have exactly one AccessRequest object and fail otherwise.

Have updated in a88cffb

ckanext/unhcr/mailer.py
chris48s: For me 

https://127.0.0.1:5000/user/login and
https://127.0.0.1:5000/service/login
do the same thing in dev.

But in production,
https://ridl.unhcr.org/user/login and
https://ridl.unhcr.org/service/login
don't do the same thing.

ckanext/unhcr/templates/user/new_user_form.html
chris48s: opened: https://github.com/ckan/ckan/pull/5623

amercader:
@chris48s AFAIC this is good to merge

chris48s:
any further thought on https://github.com/okfn/ckanext-unhcr/pull/414#discussion_r495892054 ?

ckanext/unhcr/mailer.py
amercader: That's because in prod we don't have `CKAN___SAML2__ENABLE_NATIVE_LOGIN=True` yet right?

chris48s: No. `CKAN___SAML2__ENABLE_NATIVE_LOGIN=True` is set in production already:

https://github.com/okfn/docker-ckan-unhcr-aws/blob/8677f6b0b34bfda097a270ece41e7b66732c76b3/ckan/environment/base.env#L49

It is set in the base env, and we're no overriding it in the prod settings https://github.com/okfn/docker-ckan-unhcr-aws/blob/master/ckan/environment/prod.env

`CKAN__AUTH__CREATE_USER_VIA_WEB=True` is the setting that we are using in dev but haven't applied in prod, which should not affect this AFAIK.

amercader: Interesting. I need to investigate what's happening here. Let's use service/login for now then. If you want to use a named route you can use `toolkit.url_for('staff_login', qualified=True)` here (I know, ckanext-saml2 is terrible)

chris48s: `toolkit.url_for('staff_login', qualified=True)`
gives me https://localhost:5000staff_login

`toolkit.url_for(controller='ckanext.saml2.plugin:Saml2Controller', action='staff_login', qualified=True)`
gives me https://localhost:5000/packages which is _properly_ odd.

I'm going to leave this for now as it doesn't seem like the most crucial thing to be puzzling over at the moment.

PR #416, opened 2020-09-25
clearly indicate when users are external/pending on `/user` and `/user/<id>`

### LINKED ISSUE #412: visually distinguish external users on /user

chris48s:
Closes #412
This has a bit of a merge conflict with #414 so lets work on getting that one merged first and I'll rebase this.

### DIFF
--- a/ckanext/unhcr/templates/user/list.html
+++ b/ckanext/unhcr/templates/user/list.html
@@ -10,8 +10,8 @@
   {% for user in page.items %}
     <tr>
       <td>{{ h.linked_user(user['name'], maxlength=20) }}</td>
-      <td>{% if h.user_is_external(user.User) %}external{% else %}UNHCR{% endif %}</td>
-      <td>{{ user.User.state }}</td>
+      <td>{% if h.user_is_external(user.User) %}External{% else %}UNHCR{% endif %}</td>
+      <td>{{ user.User.state.capitalize() }}</td>
     </tr>
   {% endfor %}
 </table>
--- a/ckanext/unhcr/templates/user/list.html
+++ b/ckanext/unhcr/templates/user/list.html
@@ -1,0 +1,18 @@
+{% ckan_extends %}
+
+{% block users_list %}
+<table class="table table-header table-hover table-bordered">
+  <tr>
+    <th>Username</th>
+    <th>Organisation</th>
+    <th>State</th>
+  </tr>
+  {% for user in page.items %}
+    <tr>
+      <td>{{ h.linked_user(user['name'], maxlength=20) }}</td>
+      <td>{% if h.user_is_external(user.User) %}external{% else %}UNHCR{% endif %}</td>
+      <td>{{ user.User.state }}</td>
+    </tr>
+  {% endfor %}
+</table>
+{% endblock %}
--- a/ckanext/unhcr/templates/user/list.html
+++ b/ckanext/unhcr/templates/user/list.html
@@ -10,7 +10,7 @@
   {% for user in page.items %}
     <tr>
       <td>{{ h.linked_user(user['name'], maxlength=20) }}</td>
-      <td>{% if h.user_is_external(user.User) %}External{% else %}UNHCR{% endif %}</td>
+      <td>{% if user.User.external %}External{% else %}UNHCR{% endif %}</td>
       <td>{{ user.User.state.capitalize() }}</td>
     </tr>
   {% endfor %}
--- a/ckanext/unhcr/templates/user/read_base.html
+++ b/ckanext/unhcr/templates/user/read_base.html
@@ -1,0 +1,60 @@
+{% ckan_extends %}
+
+{% block user_info %}
+<div class="info">
+  <dl>
+    {% if user.name.startswith('http://') or user.name.startswith('https://') %}
+      <dt>{{ _('Open ID') }}</dt>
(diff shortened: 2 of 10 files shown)


chris48s:
I've rebase this to resolve the conflicts

ckanext/unhcr/templates/user/list.html
```diff
@@ -0,0 +1,18 @@
...
+  <tr>
+    <th>Username</th>
+    <th>Organisation</th>
+    <th>State</th>
+  </tr>
+  {% for user in page.items %}
+    <tr>
+      <td>{{ h.linked_user(user['name'], maxlength=20) }}</td>
+      <td>{% if h.user_is_external(user.User) %}external{% else %}UNHCR{% endif %}</td>
+      <td>{{ user.User.state }}</td>
```
amercader, line 14:
```suggestion
      <td>{% if h.user_is_external(user.User) %}External{% else %}UNHCR{% endif %}</td>
      <td>{{ user.User.state.capitalize() }}</td>
```

ckanext/unhcr/templates/user/read_base.html
```diff
@@ -0,0 +1,60 @@
...
+        UNHCR
+      {% endif %}</dd>
+  </dl>
+  <dl>
+    <dt>{{ _('State') }}</dt>
+    <dd>
+      {% if user.state != 'active' %}
+        <span class="label label-warning">{{ user.state }}</span>
+      {% else %}
+        {{ user.state }}
```
amercader, line 44:
```suggestion
        <span class="label label-warning">{{ user.state.capitalize() }}</span>
      {% else %}
        {{ user.state.capitalize() }}
```

ckanext/unhcr/templates/user/read_base.html
```diff
@@ -0,0 +1,60 @@
...
+  {% endif %}
+  <dl>
+    <dt>{{ _('Member Since') }}</dt>
+    <dd>{{ h.render_datetime(user.created) }}</dd>
+  </dl>
+  <dl>
+    <dt>{{ _('Organisation') }}</dt>
+    <dd>
+      {% if h.user_is_external(user) %}
+        <span class="label label-warning">external</span>
```
amercader, line 33:
```suggestion
        <span class="label label-warning">External</span>
```

amercader (approved):
Looks good @chris48s , I just added some suggestions to capitalize the labels

chris48s:
In https://github.com/okfn/ckanext-unhcr/pull/410 we went with lowercase using the same styles. Do you want me to update those to "External" (rather than "external") as well, or is there a reason for them to be different here?

chris48s:
I realised there is another issue with this :(

When we call `user_show`, it only returns the email field if we are logged in as a sysadmin or the user we are showing. If I am some other user, I can view another user's profile but `user_show` doesn't return the external field, or the email. This causes us to throw `Invalid: 'not a user object'` in `user_is_external()`. In that case where we've got a user object but no `external` field and no email, I guess we need to fall back again and query the DB directly to bypass the auth and get the `external` value?

amercader:
@chris48s the `external` property should always be set at the model level (ie `is_user_external` should call `self.email` and check the domain). Then we extend `user_show` to manually add `user['external'] = user_obj.external`. This way whenever we receive a user object, either a model object or a user dict we already have the `external` property available. We shouldn't need to call a `is_user_external` helper in the templates as this property will be in the user object already

Does this make sense?
So:

```python
# in plugin.py
def user_is_external(self):
    ''' 
    Returns True if user email is not in the managed internal domains.
    '''
    try:
        domain = self.email.split('@')[1]
    except AttributeError:
         # Internal sysadmin user does not have email
        if self.sysadmin:
            return False
        else:
            return True

    internal_domains = config.get('ckanext.unhcr.internal_domains', INTERNAL_DOMAINS),
        sep = ',')

    return domain not in internal_domains

class UnhcrPlugin(...):

    def update_config(self, config_):
        # ...
        User.external = property(user_is_external)


#in action.py

@toolkit.chained_action
def user_show(up_func, context, data_dict):
    user_dict = up_funct(context, data_dict)

    user_obj = context['user_obj']
    user_dict['external'] = user_obj.external
    return user_dict

```

amercader:
Regarding capitalization, I believe it's better to Capitalize labels, but if you need to change it in many other places don't bother now.

chris48s:
Yeah - I see where you're going with that, except we want the external flag from the user who we're showing not the user who is calling the action. i.e: we want to get .`external` for the `data_dict['id']` or `data_dict['name']` user because `context['user_obj']` is the calling user.
Any thoughts on how to deal with the case where we've called `user_list` with `return_query` and we're dealing with a ResultProxy instead of a dict or a User obejct? i.e: https://github.com/ckan/ckan/blob/7900935de6270ee80ec64a5378251dafedf3a0e4/ckan/logic/action/get.py#L894-L896
Anyways, I'll revisit this PR on Monday and have another go at it..

chris48s:
Scratch all that in the last post - I'm getting confused about `context['auth_user_obj']` vs `context['user_obj']` and in the case where we've got a `ResultProxy`, we can access the `User` model object, so that's also fine.

This should be an improvement

amercader (approved):
This is good to go, moving the external property to the model makes everything clearer :+1:

Issue #415, opened 2020-09-25
Landing page


chris48s:
from https://github.com/okfn/ckanext-unhcr/pull/414#issuecomment-698948089 :

> 1. There's no way to know where to register when visiting the site. Perhaps initially we can just point testers to /user/register but before lunch we need to have a path for users to find this form. (eg Section on the home page for external users)

> 2. The login page should make clearer who needs to use what. Most users need the "Login with SSO" link, so the user name and password fields could be hidden or on a separate page

I think we probably need a home/landing page when you hit https://ridl.unhcr.org with links to funnel users to

* Login (UNHCR users)
* Register (external users)
* Login (external users)

and a bit of text explaining who needs to use what/when.

PR #419, opened 2020-10-06
UI/Landing page for EDD users

chris48s:
Refs #377
Refs #415
Refs https://github.com/okfn/ckanext-unhcr/pull/414#issuecomment-698948089

This tidies up the UI for external users a bit. Maybe we also need some additional text around this stuff, but I'm going to suggest we defer to the client for guidance on that.

### DIFF
--- a/ckanext/unhcr/plugin.py
+++ b/ckanext/unhcr/plugin.py
@@ -48,6 +48,7 @@ ALLOWED_ACTIONS = [
     'resource_show',
     'resource_update',
     'resource_view_list',
+    'site_read',
 ]
 
 
--- a/ckanext/unhcr/templates/home/index.html
+++ b/ckanext/unhcr/templates/home/index.html
@@ -13,27 +13,29 @@
     <div role="main" class="hero">
       <div class="container">
         {% block search %}
-        {% set tags = h.get_facet_items_dict('tags', limit=3) %}
+          {% if not c.userobj.external %}
+            {% set tags = h.get_facet_items_dict('tags', limit=3) %}
 
-        <div class="module module-search module-narrow module-shallow box">
-          <form class="module-content search-form" method="get" action="{% url_for controller='package', action='search' %}">
-            <h3 class="heading">{{ _("Search the library of") }} {{ h.SI_number_span(dataset_count) }}
-          {{ _('dataset') if dataset_count == 1 else _('datasets') }}</h3>
-            <div class="search-input form-group search-giant">
-              <input id="field-main-search" type="text" class="form-control" name="q" value="" autocomplete="off" placeholder="Search keywords, metadata, version data etc." />
-              <button type="submit">
-                <i class="fa fa-search"></i>
-                <span class="sr-only">{{ _('Search') }}</span>
-              </button>
+            <div class="module module-search module-narrow module-shallow box">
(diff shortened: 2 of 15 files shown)


ckanext/unhcr/plugin.py
```diff
@@ -48,6 +48,7 @@
     'resource_show',
     'resource_update',
     'resource_view_list',
+    'site_read',
```
chris48s, line 51: Is there any issue if we grant `site_read` to EDD users?
My instinct is no, but if its a problem, we can work round it.

ckanext/unhcr/templates/home/index.html
```diff
@@ -13,27 +13,29 @@
...
-            <h3>{{ _('Popular tags') }}</h3>
-            {% for tag in tags %}
-              <a class="tag" href="{% url_for controller='package', action='search', tags=tag.name %}">{{ h.truncate(tag.display_name, 22) }}</a>
-            {% endfor %}
-          </div>
-        </div>
+          {% endif %}
         {% endblock %}
 
         <div>
```
chris48s, line 41: If we want to provide slightly different text for external users, we can do. For now, I've just hidden search and tags.

ckanext/unhcr/plugin.py
chris48s: Travis might have answered this question for me. I'll look into this tomorrow

chris48s: `site_read` enabled too much other stuff, so I bypassed the auth for this one view

amercader:
@chris48s FYI we override `site_read` author so we can add special cases of needed: https://github.com/okfn/ckanext-unhcr/blob/master/ckanext/unhcr/auth.py#L67

ckanext/unhcr/blueprints/home.py
```diff
@@ -0,0 +1,14 @@
+from flask import Blueprint
+import ckan.plugins.toolkit as toolkit
+from ckan.views.home import index as core_index, before_request
+
+unhcr_home_blueprint = Blueprint('unhcr_home', __name__, url_prefix=u'/')
+
+def index():
+    if toolkit.c.userobj and toolkit.c.userobj.external:
```
amercader, line 8: Perhaps if we add this check to our custom site_read we don't need to override the home blueprint at all?

chris48s: I hadn't thought of doing it that way, but you're right - we could. I don't have a really strong view on it, but looking at `site_read` it is another auth function like `package_create` that requires us to have a HTTP request to run the auth function, making it hard to test. In general I think we want to avoid doing more of that where we can, but we've also already gone down that road with that particular auth function so in that sense it doesn't make anything worse. Given that context, which is your preference?

amercader: I'm slightly more favorable to extend site_read to avoid overriding the blueprint, but I don't have a strong opinion either way so you can keep this if it helps with testing

amercader (approved):
Looks good @chris48s , I pushed a commit with some UI tweaks because it was too difficult to explain (I didn't know it myself tbh). I think needs more work, but wanted to separate both login options and make the register one less prominent.
Screenshot_2020-10-09 Please login to access this page - RIDL

PR #420, opened 2020-10-07
Don't allow external user to be org member or dataset collaborator

### LINKED ISSUE #413: don't show external/pending users on /data-container/membership
Currently /data-container/membership ('manage membership' page) shows external users and pending accounts, allowing them to be assigned to containers. This shouldn't be possible.
As well as not showing them in the front-end, it should probably also be a validation error in the `organization_member_create` action if we try to add an external user as a member of a container?
Are there any other places where we are showing a list of users where we need to filter external/pending users?

chris48s:
Closes #413

As far as I can tell, these are the only places we need to cover this but let me know if I have missed anything.

I hit the exact same issue in this PR as I did in #416 so I've cherry-picked the same commit from the branch on to here.
8c1fed6 in PR #416 and
9da34cb in this PR
are the exact same change.

Once one of them gets approved/merged I will rebase the other one to tidy up.

### DIFF
--- a/ckanext/unhcr/actions.py
+++ b/ckanext/unhcr/actions.py
@@ -18,7 +18,7 @@ import ckan.logic.action.patch as patch_core
 import ckan.lib.activity_streams as activity_streams
 import ckan.lib.dictization.model_dictize as model_dictize
 from ckanext.collaborators.logic import action as collaborators_action
-from ckanext.unhcr import helpers, mailer
+from ckanext.unhcr import helpers, mailer, utils
 from ckanext.unhcr.models import AccessRequest
 from ckanext.scheming.helpers import scheming_get_dataset_schema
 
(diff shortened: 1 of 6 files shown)


ckanext/unhcr/actions.py
```diff
@@ -837,3 +864,67 @@ def search_index_rebuild(context, data_dict):
...
+    return {
+        'q': [not_missing, unicode_safe],
+        'ignore_self': [ignore_missing],
+        'limit': [ignore_missing, natural_number_validator],
+        'include_external': [ignore_missing, boolean_validator],
+    }
+
+
+@core_logic.validate(unhcr_autocomplete_schema)
+def user_autocomplete(context, data_dict):
```
chris48s, line 888: It would have been quite nice to be able to do something like

```py
@toolkit.chained_action
def user_autocomplete(up_func, context, data_dict):
    result = up_func(context, data_dict)
    # remove things from result we don't want
    return result
```

or

```py
@toolkit.chained_action
def user_autocomplete(up_func, context, data_dict):
    # data_dict['extra_filters'] = some stuff...
    return up_func(context, data_dict)
```

here, but it wasn't possible so I've had to override/re-implement the whole thing

ckanext/unhcr/actions.py
```diff
@@ -837,3 +864,67 @@ def search_index_rebuild(context, data_dict):
...
+    ):
+    return {
+        'q': [not_missing, unicode_safe],
+        'ignore_self': [ignore_missing],
+        'limit': [ignore_missing, natural_number_validator],
+        'include_external': [ignore_missing, boolean_validator],
+    }
+
+
+@core_logic.validate(unhcr_autocomplete_schema)
```
chris48s, line 887: I wasn't totally sure what to do about validation here. I've not (knowingly) hit an action with this style of validation on it before. I copied the approach from core and added the extra field here, although it seems that if I hadn't the extra param I added would have still been available in an `__extras` property. Validation of the `data_dict` param on actions is something that I've been dealing with manually, but also haven't received much guidance on. Should we really be defining schemas for these where there is non-trivial validation to perform, or are they only used in certain situations? As far as I can tell, none of other the custom actions in this project have them.

ckanext/unhcr/actions.py
```diff
@@ -837,3 +864,67 @@ def search_index_rebuild(context, data_dict):
...
+    :param include_external: include external users in the output (optional,
+        default: ``False``)
+    :type include_external: bool
+
+    :rtype: a list of user dictionaries each with keys ``'name'``,
+        ``'fullname'``, and ``'id'``
+    '''
+    m = context.get('model', model)
+    user = toolkit.get_or_bust(context, "user")
+    include_external_users = data_dict.get('include_external', False)
```
chris48s, line 905: I figure the safest/most future-proof thing to do here is make the default `False`.

amercader: Yes, definitely

ckanext/unhcr/actions.py
amercader: What issues prevented you from used the chained action?

ckanext/unhcr/actions.py
amercader: I haven't used schema validation on params much myself. I think it was implemented for parameters that were repeated across actions like pagination etc to avoid boilerplate code. It's a matter of preference really.

amercader (approved):
Looks good, I'm just curious on what issues you had chaining `user_autocomplete

chris48s:
> I'm just curious on what issues you had chaining `user_autocomplete`

Sorry I probably probably should have explained more clearly. The chaining itself wasn't an issue. It _works_, but `user_autocomplete` as a function just wasn't flexible in its inputs or outputs in the way we needed.

So for

```py
@toolkit.chained_action
def user_autocomplete(up_func, context, data_dict):
    # data_dict['extra_filters'] = some stuff...
    return up_func(context, data_dict)
```

mechanically, I could pass an extra param in but `user_autocomplete` internally doesn't know what to do with the extra param - it just accepts what its hard-cdoded to accept, so that doesn't do anything. (I just made up `extra_filters` )

In terms of filtering the output

```py
@toolkit.chained_action
def user_autocomplete(up_func, context, data_dict):
    result = up_func(context, data_dict)
    # remove things from result we don't want
    return result
```

this seemed more promising but there's 2 issues. One is that `user_autocomplete` doesn't return full user objects or dicts - it just returns username and display_name so you can't do additional filtering on another field (in this case `external`) without making another round-trip to the DB. Also, if you do additional filtering as a post-process it breaks the `limit` param. i.e: if we've called `user_autocomplete()` with `data_dict['limit'] = 20` and it returns 20 results but then I subtract 15 records from the result because they are external users, I now return 5 rows not 20. When you consider that `user_autocomplete()` is performance-critical (we expect it to perform well enough to be called on keystroke essentially) I figured adding extra database IO was a big ask but realising it breaks the `limit` param too sealed the deal.

PR #422, opened 2020-10-16
Allow container admins to review deposits for their own container

### LINKED ISSUE #381: Allow container admins to review EDDs for their own container
Currently only the curation team can interact with data deposits.
Once we allow a wider cohort of users to submit data deposits, we expect there to be a higher volume of submissions and the curation team are unlikely to have sufficient context to review them all effectively.
At the moment, the only users who can interact with the `data-deposit` container are the curation team. There is a challenge here about how we expose some subset of that to other users.

chris48s:
Closes #381

This PR adds the ability for container admins to see and review/approve data deposits where the target container is the container they are an admin of.
In terms of the business requirements for this, we agreed that this change behaviour applies to all deposited datasets, so container admins gain the ability to interact with deposited datasets submitted by all users (both UNHCR staff and external users).

### DIFF
--- a/ckanext/unhcr/helpers.py
+++ b/ckanext/unhcr/helpers.py
@@ -312,7 +312,7 @@ def get_data_deposit():
     return deposit
 
 
-def get_data_curation_users():
+def get_data_curation_users(dataset):
     context = {'model': model, 'ignore_auth': True}
     deposit = get_data_deposit()
 
@@ -330,15 +330,25 @@ def get_data_curation_users():
         'object_type': 'user',
     })
 
+    container_admins = []
+    owner_org_dest = dataset.get('owner_org_dest')
+    if owner_org_dest and owner_org_dest != 'unknown':
+        container_admins = toolkit.get_action('member_list')(context, {
+            'id': owner_org_dest,
+            'capacity': 'admin',
+            'object_type': 'user',
+        })
+
     # Get users
     users = []
-    for item in depadmins + curators:
+    for item in depadmins + curators + container_admins:
         user = toolkit.get_action('user_show')(context, {'id': item[0]})
         users.append(user)
 
+    users = [dict(tup) for tup in {tuple(u.items()) for u in users}]  # de-dupe
+
     # Sort users
-    users = list(sorted(users,
-        key=itemgetter('display_name')))
+    users = list(sorted(users, key=itemgetter('display_name', 'name')))
 
     return users
 
@@ -350,7 +360,7 @@ def get_deposited_dataset_user_curation_status(dataset, user_id):
     # General
     status = {}
     status['error'] = get_dataset_validation_error_or_none(dataset, context)
-    status['role'] = get_deposited_dataset_user_curation_role(user_id)
+    status['role'] = get_deposited_dataset_user_curation_role(user_id, dataset)
     status['state'] = dataset['curation_state']
     status['final_review'] = dataset.get('curation_final_review')
     status['active'] = dataset['state'] == 'active'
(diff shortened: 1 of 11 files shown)


ckanext/unhcr/tests/test_controllers.py
```diff
@@ -40,6 +40,18 @@ def setup(self):
         self.creator = core_factories.User(name='creator', id='creator')
```
chris48s, line 40: Sorry. The diff for this entire file is horrible. Basically I've:

a) Added `'target_container_admin'`, `'target_container_member'` and `'other_container_admin'` cases to every relevant test
b) Replaced all the `# TODO: follow redirect and check for "action is not available"`s with actually doing that. It seemed worthwhile since I had to edit basically every single test case with that comment anyway

What I realised after I did all this and pushed it is I've added about 100 extra test-cases, most of which are controller tests (the logic for the data deposit was already very complicated with a lot of cases to cover and this PR increases that complexity). The test suite for this project was already slow, but this PR roughly doubles the length of a full test-suite run from ~20 mins to ~40 mins. I'm not really sure what to do about that. My instinct is that for the moment we just accept it (the test suite was already slow enough that I rarely run the whole thing locally) and review test suite speed and how we split the tests up when we migrate to CKAN 2.9 as we basically have to rewrite every test in the codebase then anyway to migrate from nose to pytest.

ckanext/unhcr/tests/test_helpers.py
```diff
@@ -384,3 +362,320 @@ def test_convert_dataset_to_microdata_survey_minimal(self):
                 },
             },
         })
+
+
+class TestDepositedDatasetHelpers(base.FunctionalTestBase):
```
chris48s, line 367: Several of these functions I modified in this PR didn't have any unit tests at all, but I wanted to write tests covering the changes I've made so I basically added a whole test suite for those functions not just the changes I made. These probably aren't completely exhaustive, but its a start..

amercader: Great, thanks for that

amercader (commented):
Overall this looks good @chris48s , I had a little play and it worked as expected. For my own sanity (and for future generations) I added docstrings to the data deposit helpers in d411188, please have a look to see I did not misunderstood anything.
The only thing of note is my question about simplifying the `container admin` role into `curator`

amercader:
Not sure where the last failures are coming from, I just added some docstrings :( 

Perhaps these tests are brittle because some different order issues [here](https://github.com/okfn/ckanext-unhcr/blob/381-container-admins-edd/ckanext/unhcr/tests/test_controllers.py#L112)?

chris48s:
I've spent no time on this yet, but 2 really quick notes..

1. Is the only change that _definitely_ you added some docstrings? There is a merge commit here that looks like it has a lot going on in it https://github.com/okfn/ckanext-unhcr/pull/422/commits/b5662e7607cfb238222ebea68775a2f52aeb5240 but I've not had a chance to try and pick apart what happened there. It might be that the diff is spurious?
2. I'm not 100% sure when you tagged 2.8.6 but I think the last time the CI build ran, https://github.com/okfn/ckanext-unhcr/blob/4f2927e23f0ba0dad25029cc3a90cebd68df37ba/bin/travis-build.bash#L17 would have picked 2.8.5 to run the tests against whereas this time it would have picked 2.8.6 so maybe the world changed under us between test runs?

I'll have a look at it tomorrow.

amercader:
@chris48s there was something odd with this branch, #427 is the cleaned up version with tests passing

ckanext/unhcr/helpers.py
```diff
@@ -330,15 +330,25 @@ def get_data_curation_users():
...
+    for item in depadmins + curators + container_admins:
         user = toolkit.get_action('user_show')(context, {'id': item[0]})
         users.append(user)
 
+    users = [dict(tup) for tup in {tuple(u.items()) for u in users}]  # de-dupe
+
     # Sort users
-    users = list(sorted(users,
-        key=itemgetter('display_name')))
+    users = list(sorted(users, key=itemgetter('display_name', 'name')))
```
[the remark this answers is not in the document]
chris48s: You can have two users whose fullname is John Smith, but they'll have unique usernames so doing that gives a sensible sort order.
I accept this is probably less of a big deal in real life, but my dev install currently has 42 different users called Joe Bloggs all with different usernames 😂

ckanext/unhcr/helpers.py
```diff
@@ -371,23 +381,39 @@ def get_deposited_dataset_user_curation_status(dataset, user_id):
     return status
 
 
-def get_deposited_dataset_user_curation_role(user_id):
+def get_deposited_dataset_user_curation_role(user_id, dataset=None):
```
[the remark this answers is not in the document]
chris48s: So I think there's 2 different ways to develop collapsing the container admin role into curator. In both cases, if we call `get_deposited_dataset_user_curation_role()` with the dataset param and the user is admin of the target container, `get_deposited_dataset_user_curation_role()` returns `'curator'`.. then there are 2 options:

One is we say if the user is an admin of >1 container and `get_deposited_dataset_user_curation_role()` is called with no dataset param then `get_deposited_dataset_user_curation_role()` should return `'curator'`. If we do that, it doesn't really address the thing I'm worried about.

The other way is you say if the user is an admin of >1 container and `get_deposited_dataset_user_curation_role()` is called with no dataset param then `get_deposited_dataset_user_curation_role()` should return `'depositor'`, or `'user'`. That is safer (and does solve the problem I'm worried about, I think), but it does mean depending on what page you're on the sidebar will tell you you have a different role, which was one of the things I was trying to avoid. We could just drop the whole

> As a {role} you can...

and say

> Available actions:

or something. I _think_ the only other problem we'd need to solve (based on preliminary investigation) is https://github.com/okfn/ckanext-unhcr/blob/a8998d8681e03d841a5af81e7df7b8762f060f04/ckanext/unhcr/templates/organization/snippets/curation_sidebar.html#L33-L39 but I think we could special case that.

chris48s:
This was getting confusing because I was pushing more commits to https://github.com/okfn/ckanext-unhcr/pull/427 but replying to comments on this PR. I've force pushed everything from the `381-container-admins-edd-2` up to this branch so this PR now has the correct commits on it _and_ the discussions

Issue #423, opened 2020-10-16
EDD: fix editing target containers


chris48s:
On the EDD branch at the moment, data deposits can only be re-targetted at containers with the `visible_external` flag. This seems to also be true for deposits made by internal users, so there's a bug to fix here.

The desired outcome is:

- ~External users can only initially target at data container with the `visible_external` flag~
- ~Internal users can initially target at any data container~
- ~When an internal curator user edits a deposit from an external user, they can re-target it at containers without the `visible_external` flag~
- If a container admin edits a dataset from an external user, they should only be able to re-target it at sub-containers of their own container
- ~If the external user edits it again after that has happened, this shouldn't break their ability to edit it/change the target container~

chris48s:
I think the current feedback on this one is we're happy with the current implementation so I'm going to close this. We can come back to it if we need to.

PR #426, opened 2020-10-19
fix editing target containers

chris48s:
Refs #423

This makes most of the improvements noted under #423 :

- External users can only initially target at data container with the `visible_external` flag
- Internal users can initially target at any data container
- When an internal curator user edits a deposit from an external user, they can re-target it at containers without the `visible_external` flag
- If the external user edits it again after that has happened, this doesn't break their ability to edit it/change the target container (the external user can still save the deposited dataset back to the internal container)

so for example, if an external user submits a dataset targetted at Americas then an internal admin changes that to Argentina but bounces it back to the external user for changes, the internal user can still save it back to Argentina (even though they wouldn't have normally been able to target a deposit at Argentina - only the top-level containers) .

Screenshot at 2020-10-19 15-18-50

The thing I haven't looked at yet is:

- If a container admin edits a dataset from an external user, they should only be able to re-target it at sub-containers of their own container

partly because it depends on #422 so we need to get that reviewed first, but also maybe we can do without that point for the pilot. Lets discuss when we talk this week.

### DIFF
--- a/ckanext/unhcr/helpers.py
+++ b/ckanext/unhcr/helpers.py
@@ -51,21 +51,43 @@ def get_data_container(id):
     return toolkit.get_action('organization_show')(context, {'id': id})
 
 
-def get_all_data_containers(exclude_ids=[], include_unknown=False, external_user=True):
+def get_all_data_containers(
+    exclude_ids=None,
+    include_ids=None,
+    include_unknown=False,
+    userobj=None,
+):
+    if not exclude_ids:
+        exclude_ids = []
+    if not include_ids:
+        include_ids = []
+    include_ids = [id_ for id_ in include_ids if id_ and id_ != 'unknown']
+    if not userobj:
+        userobj = toolkit.c.userobj
+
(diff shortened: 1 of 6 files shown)


ckanext/unhcr/helpers.py
```diff
@@ -51,21 +51,43 @@ def get_data_container(id):
     return toolkit.get_action('organization_show')(context, {'id': id})
 
 
-def get_all_data_containers(exclude_ids=[], include_unknown=False, external_user=True):
+def get_all_data_containers(
+    exclude_ids=None,
+    include_ids=None,
+    include_unknown=False,
+    userobj=None,
+):
```
chris48s, line 59: we weren't modifying `exclude_ids` in the function, but I decided to get rid of the [mutable default argument](https://docs.python-guide.org/writing/gotchas/) here while I noticed it.

amercader (approved):
This looks good @chris48s

PR #428, opened 2020-10-23
restrict choices for owner_org_dest when container admin is editing

chris48s:
Refs #423

### DIFF
--- a/ckanext/unhcr/helpers.py
+++ b/ckanext/unhcr/helpers.py
@@ -56,6 +56,7 @@ def get_all_data_containers(
     include_ids=None,
     include_unknown=False,
     userobj=None,
+    dataset=None,
 ):
     if not exclude_ids:
         exclude_ids = []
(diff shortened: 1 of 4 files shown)


ckanext/unhcr/helpers.py
```diff
@@ -56,6 +56,7 @@ def get_all_data_containers(
     include_ids=None,
     include_unknown=False,
     userobj=None,
+    dataset=None,
```
chris48s, line 59: I'm not really convinced by this entire function at this point. I think we probably need to revisit and clean this up but for the moment I've piled some more mess into it. I think its likely that requirements around this are going to change again (with a further increase in complexity) in the next few days anyway.

amercader: One thing is for sure we rarely get *all* data containers as suggested by the name :)

Refactoring later sounds good :+1:

amercader (approved):

Issue #436, opened 2020-11-12
Store "focal point" on user object


chris48s:
When an external user signs up for an account, collect a required field "focal point" (which person or team inside UNHCR liaises with this external user). This should be exposed on the user detail page.

A pre-requisite for both this issue and #379 is to store additional fields on the user model so the next thing we want to do to support both of these issues is to backport the `plugin_extras` feature from CKAN 2.9 onto 2.8 with a core patch on the deploy repo.

Questions: 
- Does it do any harm if we apply that patch to prod as well? It probably simplifies things if we can.
- Should we also move the existing free text box we collect on signup on to the user model too?

Issue #379, opened 2020-07-08
User account expiry dates


chris48s:
External accounts should be in some way time-limited.

We need some way to manage admins extending the expiry date/users requesting an extension.
Lets defer this one for the moment. Its not needed for a MVP
Hopefully we will have CKAN 2.9 User extras by the time we reach this.

chris48s:
* Put an initial time limit on the account when it is created (6 months)
* When the account is going to expire soon (1 month), notify the user when they log in and allow them to request an extension if they want. Users who have stopped using it won't see the notification. Users who are still active will. Maybe they need an email too?
* Allow admins to see external accounts sorted by expiry so they can see which ones are going to expire soon. Admins can edit the expiry date to extend or expire the account.

chris48s:
So far we are storing a date on signup but we're not actually doing anything with it yet

avdata99:
Already covered at #598

PR #443, opened 2020-11-19
store focal_point and expiry_date in user.plugin.extras

chris48s:
There is a companion PR to this at https://github.com/okfn/docker-ckan-unhcr-aws/pull/22 which adds the core patch we need to make this work.

- Collect "focal point" as required field on external user signup
- Store `focal_point` and `expiry_date` in `user.plugin_extras`
- Expose `focal_point` on user read page
- Don't do anything with `expiry_date` yet other than store it - there will be some further discussion with the client on the details of this

~Hopefully c178cdc will fix the CI build. I'll check on it in the morning~
~Edit: also I have a merge conflict to resolve~

Refs  #436
Refs #379


ckanext/unhcr/actions.py
```diff
@@ -952,4 +952,79 @@ def user_list(up_func, context, data_dict):
...
+    ]
+    if not extras:
+        extras = {}
+    out_dict = {}
+    for field in CUSTOM_FIELDS:
+        out_dict[field['name']] = extras.get(field['name'], field['default'])
+    return out_dict
+
+
+def unhcr_plugin_extras_patch(context, data_dict):
```
chris48s, line 984: The idea of these two actions (`unhcr_plugin_extras_patch` and `unhcr_plugin_extras_show`) is to give us some actions that expose read/write to users with `user_show`/`user_update`, but only on the 'unhcr' key within the `plugin_extras` object. This way if other plugins use their own namespace under `plugin_extras` but inherit the assumption from CKAN core that only sysadmin can read/write we don't expose an action which violates that assumption. I've implemented these how we talked about on slack: I'm using the CKAN actions API and elevating to the site user to perform these operations instead of authenticating as the `context` user. After I did that, it occurred to me that another possible implementation here would be to bypass the auth by querying the DB directly using `func.jsonb_set` / `func.jsonb_extract_path` to deep-write/read inside the JSON object. This might actually be a neater solution than spoofing a different user and using the CKAN actions. What do you think?

ckanext/unhcr/actions.py
```diff
@@ -952,4 +952,79 @@ def user_list(up_func, context, data_dict):
...
     user['external'] = context['user_obj'].external
+    if not data_dict.get('include_plugin_extras'):
+        user['plugin_extras'] = {}
+        user['plugin_extras']['unhcr'] = (
+            toolkit.get_action('unhcr_plugin_extras_show')(context, data_dict)
+        )
+    return user
+
+
+def _init_plugin_extras(user):
```
chris48s, line 963: The intent of these two functions (`_init_plugin_extras` and `_validate_plugin_extras`) is to deal with the fact that not all account will have these fields. Accounts we've already created in the past, accounts created by the SAML2 extension and internal accounts will probably have a `None` value for `plugin_extras`. In future as more plugins use this construct we might find some of them have an object here that doesn't have a 'unhcr' key. Using these when we write and when we read makes it easier to deal with that and present a consistent picture to the next layer of abstraction. Any layers downstream from the action API can just assume these keys always exist.

amercader: Good call :+1:

ckanext/unhcr/blueprints/user.py
```diff
@@ -90,32 +91,47 @@ def post(self):
...
+
+        if not data_dict.get('focal_point'):
+            error_message = "A focal point must be specified"
+            return self.get(data_dict, {'focal_point': [error_message]}, {'focal_point': error_message})
 
         context['defer_commit'] = True
         data_dict['state'] = context['model'].State.PENDING
         deposit = get_data_deposit()
         containers = [data_dict.get('container'), deposit['id']]
+        expiry_date = date.today() + timedelta(days=180)  # six months-ish
```
amercader, line 120: Shall we make this configurable? `ckanext.unhcr.external_accounts_expiry_delta` or whatever

ckanext/unhcr/blueprints/user.py
```diff
@@ -90,32 +91,47 @@ def post(self):
...
 
         if not data_dict.get('container'):
-            message = "A region must be specified"
-            return self.get(data_dict, {'container': [message]}, {'container': message})
+            error_message = "A region must be specified"
+            return self.get(data_dict, {'container': [error_message]}, {'container': error_message})
+
+        if not data_dict.get('focal_point'):
+            error_message = "A focal point must be specified"
+            return self.get(data_dict, {'focal_point': [error_message]}, {'focal_point': error_message})
```
amercader, line 114: Can we move this to a `user_create` chained action? That's where the rest of validation for the other fields happens
Also, we are bypassing users created via the API here

ckanext/unhcr/blueprints/user.py
```diff
@@ -90,32 +91,47 @@ def post(self):
...
         context['defer_commit'] = True
         data_dict['state'] = context['model'].State.PENDING
         deposit = get_data_deposit()
         containers = [data_dict.get('container'), deposit['id']]
+        expiry_date = date.today() + timedelta(days=180)  # six months-ish
 
         try:
             model.Session.begin_nested()
             user = toolkit.get_action(u'user_create')(context, data_dict)
+            toolkit.get_action(u'unhcr_plugin_extras_patch')(
```
amercader, line 125: Same as above, this should be done in `user_create`

ckanext/unhcr/helpers.py
```diff
@@ -405,6 +405,7 @@ def get_data_curation_users(dataset):
     users = []
     for item in depadmins + curators + container_admins:
         user = toolkit.get_action('user_show')(context, {'id': item[0]})
+        user.pop('plugin_extras', None)
```
amercader, line 408: `user_show` should promote any custom fields stored in `plugin_extras` to the root level, so the rest of the app doesn't need to know about them and can just use `user['focal_point']` etc

amercader (commented):
@chris48s I think it's worth trying to encapsulate as much logic as possible in the actions, particularly `user_show` and `user_update`, and to hide the `plugin_extras` implementation as much as  possible outside the actions to make future upgrades easier.
Our custom fields (`expiry_date` and `focal_point`) should be part of the user `data_dict` reaching `user_create` and  `user_update`, and these should take care of storing them in the `plugin_extras`. Reversely, when calling `user_show` the fields should be added to the main object so you can call `user['focal_point']` from anywhere. I think your logic shouldn't change much, just moving the code down one level to actions.

This aligns with the future goal of having user schemas support similar to datasets and orgs, and makes the rest of the code clearer.

ckanext/unhcr/blueprints/user.py
chris48s: I did this in https://github.com/okfn/ckanext-unhcr/pull/443/commits/165527b5204910cd391b2b3b78010080a7ecbe7b
Because the use test factory calls `user_create` that means we need to set `focal_point` whenever we create an external user under test. Not sure if it would be better to add it to the user factory for all users (even though we wouldn't attempt to set/store it for internal users), see if we can make a custom factory which conditionally sets that field for external users only, or whether it would really make sense to just validate it on the controller given we want users to specify it on signup but we can handle it not being there (I'm not really sure what it means for an external user to be created via the API. It would bypass the approval process, for example).

chris48s:
I forced pushed this because I've nearly started again with a different approach. Hopefully this is better. We can deal with the `user_update` action when we come to need it.

amercader (approved):
The changes look great @chris48s.
I think that for the users in test we can have a couple of factories in the extension that extend the core `User` one: `InternalUser` and `ExternalUser`, the latter sets `focal_point` automatically. That will make the tests cleaner and we don't need to worry about passing `focal_point` to all user instances.

What do you think?

chris48s:
> we can have a couple of factories in the extension that extend the core `User` one: `InternalUser` and `ExternalUser`

Yes - good plan :+1:  but I'll make an issue for it rather than do it right now.

PR #444, opened 2020-11-20
allow external users to edit profile info

chris48s:
Refs #375
Refs #377

Fairly basic stuff here, but I hit another issue with our favourite plugin on this one, so this is predicated on https://github.com/okfn/ckanext-saml2/pull/4 (or some other solution to this problem) to work properly.

### DIFF
--- a/ckanext/unhcr/plugin.py
+++ b/ckanext/unhcr/plugin.py
@@ -51,6 +51,9 @@ ALLOWED_ACTIONS = [
     'site_read',
     'tag_autocomplete',
     'tag_list',
+    'user_show',
+    'user_update',
+    'user_generate_apikey',
 ]
 
 
--- a/ckanext/unhcr/plugin.py
+++ b/ckanext/unhcr/plugin.py
@@ -485,6 +485,7 @@ class UnhcrPlugin(
         functions['user_update_sysadmin'] = auth.user_update_sysadmin
         functions['external_user_update_state'] = auth.external_user_update_state
         functions['search_index_rebuild'] = auth.search_index_rebuild
+        functions['user_show'] = auth.user_show
         return functions
 
     # IActions
--- a/ckanext/unhcr/templates/user/edit.html
+++ b/ckanext/unhcr/templates/user/edit.html
@@ -1,24 +1,28 @@
 {% ckan_extends %}
 {% block primary_content_inner %}
 
-  <legend>{{ _('Change details') }}</legend>
+  {% if user_dict.external %}
+    {{ form | safe }}
+  {% else %}
+    <legend>{{ _('Change details') }}</legend>
 
-  <div class="alert warning">
-    RIDL uses details from your UNHCR Active Directory account for authorization. If you need to change your password or user name please do so in Active Directory.
-  </div>
+    <div class="alert warning">
+      RIDL uses details from your UNHCR Active Directory account for authorization. If you need to change your password or user name please do so in Active Directory.
+    </div>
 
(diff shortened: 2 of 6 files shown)


ckanext/unhcr/plugin.py
```diff
@@ -51,6 +51,9 @@
     'site_read',
     'tag_autocomplete',
     'tag_list',
+    'user_show',
```
chris48s, line 54: Granting user_show also allows external users to read the profile of other users as well as their own.
I don't think that causes a problem, but I can have a look at restricting it if we think it does.

ckanext/unhcr/templates/user/edit_user_form.html
```diff
@@ -0,0 +1,46 @@
+{% import 'macros/form.html' as form %}
```
chris48s, line 1: This is another place where it would be nice if the template was split into blocks so we didn't have to override the whole thing. We should probably make a contribution similar to https://github.com/ckan/ckan/pull/5623 in core to add blocks in this template too.

chris48s:
This test failure happens in CI but doesn't reproduce locally. Currently trying to debug it

chris48s:
The reason the tests were failing in CI but not locally is because were applying https://github.com/okfn/docker-ckan-unhcr-aws/blob/master/ckan/patches/ckan/04_remove_user_request_handler.patch locally but not in CI.

I've pulled that in here too. Independently, I renamed the existing core patches in https://github.com/okfn/ckanext-unhcr/pull/443 so I won't repeat that change here but they'll all slot into numerical order when we merge both.

ckanext/unhcr/plugin.py
amercader: I think it's worth extending the `user_show` auth function and rejecting calls if the user in the context is external and the user on the data_dict are not themselves, to plug any potential leaks there

ckanext/unhcr/templates/user/edit.html
```diff
@@ -1,24 +1,28 @@
...
-      {% endif %}
-    {% endblock %}
-  </div>
+    <div class="form-actions">
+      {% block delete_button %}
+        {% if h.check_access('user_delete', {'id': user_dict.id})  %}
+          <a class="btn btn-danger pull-left" href="{% url_for controller='user', action='delete', id=user_dict.id %}" data-module="confirm-action" data-module-content="{{ _('Are you sure you want to delete this User?') }}">{% block delete_button_text %}{{ _('Delete') }}{% endblock %}</a>
+        {% endif %}
+      {% endblock %}
+      {% block generate_button %}
```
amercader, line 19: Don't external users need this? IIRC you mentioned that API keys are needed for cloudstorage

ckanext/unhcr/templates/user/edit_user_form.html
amercader: Sounds good!

ckanext/unhcr/templates/user/edit.html
chris48s: They do and it is, but I don't get the issue. We are granting external users `user_generate_apikey in this PR

Screenshot at 2020-12-02 16-58-14

amercader: Sorry, I'm not sure what I meant, this looks good

amercader (approved):
@chris48s This is good to go, but while testing it I noticed that external users can not see their own profile page (listed on the cog menu) and so can't see their details including API key. Is this by design?

Issue #434, opened 2020-11-12
Store user who approved the request on AccessRequest objects


chris48s:
This is a useful data point for all AccessRequests (user, dataset, container)
Specifically for users/EDD we should surface this info (who approved the account) on the user detail page for external users

PR #445, opened 2020-11-23
store/expose user who approved/rejected user account requests

chris48s:
- Store the user who approved or rejected an access request on the AccessRequest object (this also applies to container/organisation access requests - it is a useful data point there too)
- Expose the user who approved or rejected a user account request on the user detail page

Refs #434


amercader (approved):
Looks good, only two minor comments

PR #447, opened 2020-12-02
Scan resources with Clam AV

chris48s:
This feature has three parts that all fit together:

1. This pull request
2. https://github.com/okfn/docker-ckan-unhcr-aws/pull/23
3. The contents of https://github.com/okfn/ckan-clamav-service (which I've just pushed to `main` and tagged, but the fact it is a standalone repo doesn't make it exempt from review)

### DIFF
--- a/ckanext/unhcr/actions.py
+++ b/ckanext/unhcr/actions.py
@@ -4,6 +4,7 @@ import json
 import logging
 import requests
 from urlparse import urljoin
+from dateutil.parser import parse as parse_date
 from sqlalchemy import and_, desc, or_, select
 from sqlalchemy.dialects.postgresql import array
 from ckan import model
@@ -542,6 +543,44 @@ def _task_is_stale(task):
     return time_since_last_updated > assume_task_stale_after
 
 
+def _should_resubmit(context, task, metadata):
+    if task['state'] != 'complete':
+        return False
+
+    try:
+        resource_show = toolkit.get_action('resource_show')
+        resource_dict = resource_show(context, {'id': task['entity_id']})
+    except toolkit.ObjectNotFound:
+        return False
+
+    if resource_dict.get('last_modified') and metadata.get('task_created'):
+        try:
+            last_modified_datetime = parse_date(resource_dict['last_modified'])
+            task_created_datetime = parse_date(metadata['task_created'])
+            if last_modified_datetime > task_created_datetime:
+                log.debug('Uploaded file more recent: {0} > {1}'.format(
+                        last_modified_datetime,
+                        task_created_datetime,
+                    )
+                )
+                return True
+        except ValueError:
+            pass
+    elif (
+        resource_dict.get('url')
+        and metadata.get('original_url')
+        and resource_dict['url'] != metadata['original_url']
+    ):
+        log.debug('URLs are different: {0} != {1}'.format(
+                resource_dict['url'],
+                metadata['original_url'],
+            )
+        )
+        return True
+
+    return False
+
+
 def scan_submit(context, data_dict):
     resource_id = toolkit.get_or_bust(data_dict, "id")
     toolkit.check_access('scan_submit', context, data_dict)
@@ -551,15 +590,10 @@ def scan_submit(context, data_dict):
     callback_url = toolkit.url_for('/api/3/action/scan_hook', qualified=True)
     site_user = toolkit.get_action('get_site_user')({'ignore_auth': True}, {})
 
-    payload = json.dumps({
-        'api_key': site_user['apikey'],
-        'job_type': 'scan',
-        'result_url': callback_url,
-        'metadata': {
-            'ckan_url': site_url,
-            'resource_id': resource_id,
-        }
-    })
+    try:
+        resource_dict = toolkit.get_action('resource_show')(context, {'id': resource_id})
+    except toolkit.ObjectNotFound:
+        return False
 
     task = {
         'entity_id': resource_id,
@@ -571,6 +605,19 @@ def scan_submit(context, data_dict):
         'value': '{}',
         'error': 'null',
     }
+
+    payload = json.dumps({
+        'api_key': site_user['apikey'],
+        'job_type': 'scan',
+        'result_url': callback_url,
+        'metadata': {
+            'ckan_url': site_url,
+            'resource_id': resource_id,
+            'task_created': task['last_updated'],
+            'original_url': resource_dict.get('url'),
+        }
+    })
+
     try:
         existing_task = toolkit.get_action('task_status_show')(context, {
             'entity_id': resource_id,
@@ -643,8 +690,8 @@ def scan_hook(context, data_dict):
     task['value'] = json.dumps(data_dict)
     task['error'] = json.dumps(data_dict.get('error'))
 
-    context['ignore_auth'] = True
-    toolkit.get_action('task_status_update')(context, task)
+    task = toolkit.get_action('task_status_update')({'ignore_auth': True}, task)
+    context['session'].refresh(context['task_status'])
 
     if task['state'] == 'error':
         recipients = toolkit.aslist(toolkit.config.get('ckanext.unhcr.error_emails', []))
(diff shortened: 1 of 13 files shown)


ckanext/unhcr/actions.py
```diff
@@ -524,6 +527,184 @@ def datasets_validation_report(context, data_dict):
...
 
+def _fail_task(context, task, error):
+    task['error'] = json.dumps(error)
+    task['state'] = 'error'
+    task['last_updated'] = str(datetime.datetime.utcnow())
+    return toolkit.get_action('task_status_update')(context, task)
+
+
+def _task_is_stale(task):
+    assume_task_stale_after = datetime.timedelta(seconds=3600)
```
chris48s, line 538: We could pull this out to a setting if we think we want to tweak it, or we could share the datapusher setting?

ckanext/unhcr/plugin.py
```diff
@@ -512,13 +514,20 @@ def get_actions(self):
...
+            'resource_update': actions.resource_update,
             'user_update_sysadmin': actions.user_update_sysadmin,
             'external_user_update_state': actions.external_user_update_state,
             'search_index_rebuild': actions.search_index_rebuild,
             'user_autocomplete': actions.user_autocomplete,
             'user_list': actions.user_list,
             'user_show': actions.user_show,
         }
+        if 'cloudstorage' in config['ckan.plugins']:
+            functions['cloudstorage_finish_multipart'] = actions.cloudstorage_finish_multipart
```
chris48s, line 529: This is a similar issue to the one described in https://github.com/okfn/ckanext-unhcr/pull/289
We run with the cloudstorage plugin, but it is disabled in CI

ckanext/unhcr/actions.py
```diff
@@ -524,6 +527,184 @@ def datasets_validation_report(context, data_dict):
...
+        'entity_id': resource_id,
+        'task_type': 'clamav',
+        'key': 'clamav'
+    })
+
+    task['state'] = status
+    task['last_updated'] = str(datetime.datetime.utcnow())
+    task['value'] = json.dumps(data_dict)
+    task['error'] = json.dumps(data_dict.get('error'))
+
```
chris48s, line 644: I haven't bothered copying over the resubmit logic from
https://github.com/ckan/ckan/blob/c04b6d64aadc7a647b8896279002626ca3be0a56/ckanext/datapusher/logic/action.py#L229-L248
I'm not 100% sure what it is trying to guard against.
Any thoughts on it?

ckanext/unhcr/templates/emails/resource/infected_file.html
```diff
@@ -0,0 +1,21 @@
...
+      Resource <a href="{{ resource_url }}">{{ resource_name }}</a> was scanned and found to be infected.
+      The result of the scan was:
+<pre>
+{{ clamav_report }}</pre>
+  {% endcall %}
+
+  {% call email.paragraph() %}
+    To protect users, we have disabled file downloads for this resource.
+  {% endcall %}
+
```
chris48s, line 20: It feels like there should be some kind of call to action here, but I'm not sure what it is..

ckanext/unhcr/utils.py
```diff
@@ -49,3 +50,22 @@ def user_is_external(user):
             return True
 
     return domain not in get_internal_domains()
+
+
+def resource_is_blocked(context, resource_id):
```
chris48s, line 55: I wasn't really sure what to do with this function.

* I don't think it is an action in itself.
* If `resource_download` was an action, I'd chain it on to that but it isn't.
* I don't really want to chain it on `resource_show` - it is still useful to be able to view the metadata of an infected resource (I think without that it becomes hard to work out what dataset it is attached to, who uploaded it, etc)
* If I chain it on the `resource_download` auth function (and return `False` if the file is infected) sysadmins bypass the auth function, and I think we always want to block even if the user is sysadmin?
* I want to use the same logic in both resource_download controller functions (local storage and cloudstorage) so I don't want to duplicate this block in both controllers

Any thoughts on where it should live? I've not written any tests for it yet as it may need rethinking slightly.

amercader: If the only place where we need to check if a resource is infected is in the download handlers then I think is fine for now to leave it in here. If we need this information in some other place (eg indexed to search for infected resources, in the templates to show a warning etc) then we might want to think of a more generic place that makes it easier to access it than calling the function each time. Eg maybe in a chained `resource_show` or `package_show` but still returning the metadata, but with a virtual field added on infected resources or something.

But for now let's leave it like this.

amercader (commented):
Looks good overall @chris48s , I left some comments

ckanext/unhcr/actions.py
```diff
@@ -524,6 +527,184 @@ def datasets_validation_report(context, data_dict):
...
+    if task['state'] == 'error':
+        recipients = toolkit.aslist(toolkit.config.get('ckanext.unhcr.error_emails', []))
+        for address in recipients:
+            subj = '[UNHCR RIDL] Error performing Clam AV Scan'
+            core_mailer.mail_recipient(
+                'admin',
+                address,
+                subj,
+                json.dumps(data_dict, indent=4)
```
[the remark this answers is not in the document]
chris48s: I used `indent=4` so it is "human-readable" :laughing:

ckanext/unhcr/actions.py
```diff
@@ -524,6 +528,236 @@ def datasets_validation_report(context, data_dict):
...
+        'key': 'clamav'
+    })
+
+    task['state'] = status
+    task['last_updated'] = str(datetime.datetime.utcnow())
+    task['value'] = json.dumps(data_dict)
+    task['error'] = json.dumps(data_dict.get('error'))
+
+    task = toolkit.get_action('task_status_update')({'ignore_auth': True}, task)
+    context['session'].refresh(context['task_status'])
```
chris48s, line 693: This had me stumped for most of the day.

I still don't know if this is a bug in `task_status_update` or `task_status_show` or if it is intended behaviour that I find suprising and that's just how CKAN handles model sessions :shrug: Basically the issue I hit here was I call `task_status_show`, task state is 'pending'. I call `task_status_update`, to update the task state to 'complete' but then if we have to resubmit the job next time I call `task_status_show` (in `scan_submit`) the state is back to 'pending'. Refreshing the model object in the context after we update it here fixes it.

As I say, I'm not really sure what the intention is here. I think if this is the intended behaviour of the core actions then the resubmit case in the datapusher may also not work although I've not really looked into it.

ckanext/unhcr/actions.py
chris48s: Right yeah the first time I looked at this I assumed that if the resource gets updated then we just kick off a second job but there is a race here if the resource is modified before the first scan completes and instead of queueing another scan we don't bother queueing it because we've already got a pending task :+1: I've implemented this in 9d6db63 and added a note about a problem I hit while doing it.

ckanext/unhcr/actions.py
amercader: I don't know the expected behaviour on core regarding this, but your change makes sense and is probably broken in core. I need to look at some datapusher related errors (https://github.com/ckan/ckan/issues/5597) at some point that might be related to this.

amercader (approved):
Looks good to go @chris48s ! I also reviewed https://github.com/okfn/ckan-clamav-service and looks good as well.

I don't think there's nothing here that is tied to RIDL logic right? Meaning that we could easily split this into a separate CKAN extension if we wanted to make it reusable at some point (not that we need now)

chris48s:
> I don't think there's nothing here that is tied to RIDL logic right?

Mostly. We could definitely pull out the `scan_submit` and `scan_hook` actions/auth functions into a standalone plugin if we wanted to use https://github.com/okfn/ckan-clamav-service with another CKAN install. There are a couple of non-generic things in this PR. For example hooking it onto `cloudstorage_finish_multipart` and the block logic on `ExtendedStorageController` are specific to CloudStorage. The email uses a RIDL template, etc

PR #450, opened 2020-12-04
clarify 'submit' CTA in deposit alert

chris48s:
One of the bits of feedback from the pilot is it needs to be clearer (for all users internal and external) that once you've finished editing a deposited dataset you need to actively submit it to the curation team for review. This is a fairly crude attempt to make the call to action clearer, but other suggestions welcome.

### DIFF
--- a/ckanext/unhcr/templates/package/read.html
+++ b/ckanext/unhcr/templates/package/read.html
@@ -14,13 +14,12 @@
       {% elif curation['state'] == 'draft' %}
 
         {% if curation['role'] == 'depositor' %}
+            <a style="float: right" href="#curation-submit" role="button" class="btn btn-warning" data-toggle="modal" title="{{ link_help }}" data-target="#curation-submit">
+              Submit Dataset
+            </a>
+
           <p>{% trans %}This deposited dataset is in <strong>draft</strong> state. You can make further edits if necessary.{% endtrans %}</p>
-          <p>{% trans %}Once all edits are made
-            <a href="#curation-submit" role="button" class="btn btn-warning" data-toggle="modal" title="{{ link_help }}" data-target="#curation-submit">
-              Submit the dataset
-            </a> for review by the curation team.{% endtrans %}
-          </p>
-          <p>{% trans %}<strong>Note:</strong> You won't be able to edit the dataset once submitted. You can submit a dataset for curation even if it does not pass the validation requirements.{% endtrans %}</p>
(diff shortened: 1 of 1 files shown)


ckanext/unhcr/templates/package/read.html
```diff
@@ -14,8 +14,13 @@
...
         {% if curation['role'] == 'depositor' %}
-          <p>{% trans %}This deposited dataset is in <strong>draft</strong> state. You can make further edits if necessary. When you want to submit it to the curation team, click on the "Submit Dataset" button on the left.{% endtrans %}</p>
-        <p>{% trans %}<strong>Note:</strong> You won't be able to edit the dataset once submitted. You can submit a dataset for curation even if it does not pass the validation requirements.{% endtrans %}</p>
+          <p>{% trans %}This deposited dataset is in <strong>draft</strong> state. You can make further edits if necessary.{% endtrans %}</p>
+          <p>{% trans %}Once all edits are made
+            <a href="#curation-submit" role="button" class="btn btn-warning" data-toggle="modal" title="{{ link_help }}" data-target="#curation-submit">
+              Submit the dataset
+            </a> for review by the curation team.{% endtrans %}
+          </p>
+          <p>{% trans %}<strong>Note:</strong> You won't be able to edit the dataset once submitted. You can submit a dataset for curation even if it does not pass the validation requirements.{% endtrans %}</p>
```
amercader, line 23: I personally don't like the button embedded in the text. What do you think about something like this?

Screenshot_2020-12-08 Test submitted by external - RIDL


```suggestion
            <a style="float: right" href="#curation-submit" role="button" class="btn btn-warning" data-toggle="modal" title="{{ link_help }}" data-target="#curation-submit">
              Submit Dataset
            </a>

          <p>{% trans %}This deposited dataset is in <strong>draft</strong> state. You can make further edits if necessary.{% endtrans %}</p>
          <p>{% trans %}Once all edits are made submit the dataset for review by the curation team.{% endtrans %}</p>
```

Perhaps the proper approach to this issue is having a separate color for the main CTA button in the sidebar (with a proper color / text color of course):

Screenshot_2020-12-08 Test submitted by external - RIDL(1)

But this is a bigger change as we will need to update the rest of the forms on the site for consistency.

chris48s: Agreed. I also prefer your layout :+1: 
I'll make an issue about making the "btn-success" style more distinct across the site

Issue #449, opened 2020-12-03
send emails in background jobs


chris48s:
There are some places where we are sending multiple emails in a user-facing operation e.g:

https://github.com/okfn/ckanext-unhcr/blob/5b0788de59aa37f1347763fa2bddf8d05f2e8a9a/ckanext/unhcr/controllers/extended_package.py#L227-L236

https://github.com/okfn/ckanext-unhcr/blob/07f9f9351c13b427ea8434b67e0404401968c5f8/ckanext/unhcr/blueprints/user.py#L151-L160

This makes the interaction slower than it needs to be because we have to wait to send off N emails in sequence. This will be worse in prod where we have a larger number of container/sys admins.
If we can farm these out to a job, we could get a good speedup here.

PR #452, opened 2020-12-08
send emails in background jobs

chris48s:
Refs #449 
I've picked out everywhere we are sending multiple emails in a loop and moved the email to background jobs. This takes email sending off the critical execution path which gives us big speedup for these operations if we have many emails to send.
It does make the testing a bit more fiddly though.


amercader (approved):
Nice one @chris48s

Issue #454, opened 2020-12-08
Create seperate test factories for InternalUser and ExternalUser


chris48s:
https://github.com/okfn/ckanext-unhcr/pull/443#discussion_r530960062
https://github.com/okfn/ckanext-unhcr/pull/443#pullrequestreview-546968385

PR #457, opened 2020-12-11
set a default target container for deposits if we can

chris48s:
- Store containers for external users on `user.plugin_extras`
- Clean up tests by adding an `ExternalUser` factory (refs  #454 ). On reflection, I think it is OK to just keep using the core factory for internal users. We don't need a specific `InternalUser` factory.
- Select container by default when users create a new deposited dataset

### DIFF
--- a/ckanext/unhcr/actions.py
+++ b/ckanext/unhcr/actions.py
@@ -1194,6 +1194,7 @@ def user_show(up_func, context, data_dict):
 
     user['focal_point'] = extras['focal_point']
     user['expiry_date'] = extras['expiry_date']
+    user['containers'] = extras['containers']
 
     return user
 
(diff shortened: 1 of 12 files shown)


ckanext/unhcr/actions.py
```diff
@@ -1194,6 +1194,7 @@ def user_show(up_func, context, data_dict):
 
     user['focal_point'] = extras['focal_point']
     user['expiry_date'] = extras['expiry_date']
+    user['containers'] = extras['containers']
```
amercader, line 1197: Is it worth calling it `default_containers` to make clear what this is about? I don't have a strong opinion on it but it seems like it might help avoid confusion with standard container membership for future devs (or our future selves)

amercader (approved):
Looks good @chris48s , just one minor comment

ckanext/unhcr/actions.py
chris48s: agreed - merging with that change :+1:

PR #458, opened 2020-12-14
swap out travis/coveralls for GH actions/codecov

chris48s:
Bye travis :wave:  Even if we ignore the time we spend queueing with travis, this _massively_ reduces the time it takes to run a CI build :rocket: 

For now I have just replicated the approach of maintaining our core patches for test in this repo and applying them to a generic CKAN 2.8 image. It is no worse than what we were already doing. We can look at building on the actual docker containers we use in dev/prod as a further improvement.


.github/workflows/test.yml
```diff
@@ -0,0 +1,76 @@
...
+
+    container:
+      image: openknowledge/ckan-dev:2.8
+
+    services:
+      solr:
+        image: ckan/ckan-solr-dev:2.8
+
+      postgres:
+        image: ckan/ckan-postgres-dev:2.8
```
chris48s, line 16: We're running Postgres 11 in the dev containers and on RDS in production but this container runs Postgres 12 so the test environment is now slightly different. We could look at running the tests on 11 but also given it seems CKAN 2.8 (and our application code) all runs on Postgres 12 with no issues we could just upgrade in production. Maybe we schedule that upgrade for some time in Jan after we launch the EDD? I don't there is any reason not to do that.

amercader: Are you sure? 2.8 should contain pg 11:  https://github.com/ckan/ckan-postgres-dev/
And the image contents seem to confirm it: https://hub.docker.com/layers/129662454/ckan/ckan-postgres-dev/2.8/images/sha256-a03f40c80d426ccd058329e0c5ef9d501882979b207cf5ee99351440c154cf47?context=explore

IIRC there were changes in the requirements needed (psycopg2) to run pg 12 on CKAN 2.8 so I'm curious why this is working now :thinking:

chris48s: I see. I thought that docker image was build from this Dockerfile:
https://github.com/okfn/docker-ckan/blob/master/postgresql/Dockerfile
which is why I assumed it was 12.

If it is running on 11, then ignore that comment: All 3 envs are in sync, so that's good :)

amercader (commented):
Yay! Let's merge this and we can improve it with using our own images later on

amercader (approved):

PR #459, opened 2020-12-15
fix race conditions when using defer_commit/background jobs

### LINKED ISSUE #448: Fix API file uploads
Creating a resource with a large file upload via the API succeeds, but the resource gets marked as deleted. Small file uploads are fine.
I think this is a complete explanation of what's going on here.

- When we call `resource_create`, CKAN calls `package_update` with `defer_commit: True`: https://github.com/ckan/ckan/blob/c04b6d64aadc7a647b8896279002626ca3be0a56/ckan/logic/action/create.py#L314-L316
- `package_update` kicks off the package `after_update` job, even though we haven't persisted the package yet: https://github.com/ckan/ckan/blob/c04b6d64aadc7a647b8896279002626ca3be0a56/ckan/logic/action/update.py#L331-L334
- Our package `after_update` job is now running
- Meanwhile back in `resource_create` we block on this statement for ~10 seconds or something while we persist a large file to Amazon S3 https://github.com/ckan/ckan/blob/c04b6d64aadc7a647b8896279002626ca3be0a56/ckan/logic/action/create.py#L326-L327
- Before we've finished that, our package `after_update` job has called `package_show` on the package we haven't committed yet https://github.com/okfn/ckanext-unhcr/blob/4225819138e17236388933068ad7b8400197bbb7/ckanext/unhcr/jobs.py#L62 and `package_update` to save it back to the DB https://github.com/okfn/ckanext-unhcr/blob/4225819138e17236388933068ad7b8400197bbb7/ckanext/unhcr/jobs.py#L69
- The `package_update` action is called with a resources list that doesn't include the resource we're in the process of creating. Deep in the innards of CKAN, it picks up the inconsistency and sets it to deleted for us here https://github.com/ckan/ckan/blob/c04b6d64aadc7a647b8896279002626ca3be0a56/ckan/lib/dictization/model_save.py#L97-L100
- Meanwhile back in `resource_create` we have now finished saving the resource so we commit the transaction https://github.com/ckan/ckan/blob/c04b6d64aadc7a647b8896279002626ca3be0a56/ckan/logic/action/create.py#L329
- `package_create` then returns the package dict without calling `model.Session.refresh()`. Hence the state being 'active' in the API response despite the fact we already set it to 'deleted'

So.. lots of fun times there. I think there's a few different things we could consider doing here of varying levels of complexity and robustness:

- Don't run background jobs if `context['defer_commit'] == True`. Seems like this is always a recipe for race conditions and inconsistent DB content. The downside is any time we use `defer_commit` (sometimes this is a choice made for us in core) we won't ever populate the fields we're calculating in the background job.
- When we call `package_update`, call it with a blank resource list and `allow_partial_update=True`. That should stop us from marking resources as deleted if they haven't been committed yet, but it does mean that when resources are uploaded via the API the data_range and weight fields will have been calculated without including the most recently uploaded resource (because we'll still be calling `package_show` before we've commited).
- Pass the `defer_commit` flag to the background job. Calling `package_update` with `defer_commit` in https://github.com/okfn/ckanext-unhcr/blob/4225819138e17236388933068ad7b8400197bbb7/ckanext/unhcr/jobs.py#L68 might work (although I'm not sure if the background job will share the session). Even if it does we will have the same issue that the data_range and weight fields will have been calculated without including the most recently uploaded resource (because we still call `package_show` before we've commited).
- I think the reason this doesn't happen when we upload via the front-end is because the last thing we do in `finish_multipart` is call `package_patch`. We could just make `resource_create` a chained action that does all its gubbins and then calls `resource_update` at the end to set `state='active'`. `resource_update` has all the same problems, but because that last call to `resource_update` only touches the metadata it doesn't need to sit there waiting on IO with s3 so https://github.com/okfn/ckanext-unhcr/blob/4225819138e17236388933068ad7b8400197bbb7/ckanext/unhcr/jobs.py#L16 covers us long enough to get everything persisted before we call `package_show`. Then we will set the resource to deleted but then set it to active again at the end and re-generate the calculated fields off the full resource list. As long as jobs execute in order rather than asynchronously we should end up in the right state at the end. This is probably the most hacky solution but probably will eventually leave us with correct data (as long as the jobs execute in order) and I think this is what is actually happening when we upload via the web UI.
- It seems like it should be possible to try and work out if the transaction is still in progress and wait until it is finished before we kick off the job (with some timeout to prevent an infinite loop). I did have a go at trying to work this out when we call `after_update`, but `context['session'].dirty` returns `IdentitySet([])`. Am I looking at the wrong thing here? If I can work this out, it seems like this would be the most robust approach.

In general, it feels like buggy behaviour in CKAN that it kicks off background jobs when defer_commit is True before the commit happens. This could be particularly bad if we never commit and roll back the transaction after the job has started, but given the commit may occur outside `package_update` I guess it is hard to be sure they are being kicked off in the correct place.

chris48s:
This PR fixes the situation where we kick off the `process_dataset_on_update` background job before we've commited a transaction when uploading resources via the API.

Unfortunately the suggestion in https://github.com/okfn/ckanext-unhcr/issues/448#issuecomment-740576521 doesn't help. The problem isn't that we call `after_update` before `model.repo.commit()` in `package_update` when `defer_commit == False` (although that is its own bug). The problem is that in this situation we're calling `package_update` with `defer_commit: False`. Even if we kick the job off in a chained `package_update`, we're still running the job before the transaction is committed. When we call `package_update` with `defer_commit: True` the transaction still doesn't get committed until here: https://github.com/ckan/ckan/blob/c04b6d64aadc7a647b8896279002626ca3be0a56/ckan/logic/action/create.py#L329

That said, I've attempted to capture the intent or spirit of your proposed solution here: Still use background jobs, but make sure we fire them in a place after the transaction has been committed.
Basically, what we're doing now is:
- If we call `package_update/defer_commit: False` we register `process_dataset_on_update` as a background job after the `package_update`
- If we call `package_update/defer_commit: True` we don't register the background job
- `resource_create` and `resource_update` also register `process_dataset_on_update` as a background job, so if we called one of those actions the job still gets registered (and it happens after all the transactions have been committed)

This solution works for all the situations that we currently deal with, but it  could be brittle because potentially we could add a plugin in future which change these assumptions (or possibly a behaviour change in a future version of CKAN may bite us). For example if we add another action which calls `package_update/defer_commit: True` we need to add a corresponding hook for that action. If some plugin overrides `resource_create`/`resource_update` to implement `defer_commit: True` we need to account for that. If we don't like this, we could go back and have a look at another of the proposed solutions in #448

In order to do this I've had to define a second plugin due to the name conflict between the `IPackageController` and `IResourceController` hooks, so there is a related PR over at https://github.com/okfn/docker-ckan-unhcr-aws/pull/25 to add the new plugin. It is a bit awkward though. Maybe our one monoplugin should really be split up into smaller logical units?

No changes were needed for resources uploaded via the frontend using cloudstorage multipart uploads.

Closes #448

### DIFF
--- a/ckanext/unhcr/plugin.py
+++ b/ckanext/unhcr/plugin.py
@@ -92,6 +92,7 @@ class UnhcrPlugin(
     plugins.implements(plugins.IFacets)
     plugins.implements(plugins.ITemplateHelpers)
     plugins.implements(plugins.IPackageController, inherit=True)
+    plugins.implements(plugins.IResourceController, inherit=True)
     plugins.implements(plugins.IAuthFunctions)
     plugins.implements(plugins.IActions)
     plugins.implements(plugins.IValidators)
@@ -450,12 +451,27 @@ class UnhcrPlugin(
 
         return search_params
 
-    def after_create(self, context, data_dict):
-        if not context.get('job'):
-            if data_dict.get('state') == 'active':
-                toolkit.enqueue_job(jobs.process_dataset_on_create, [data_dict['id']])
 
-        if data_dict.get('type') == 'deposited-dataset':
+    # IPackageController, IResourceController
+    # note: if we add more hooks that are in the interface for both
+    # IPackageController and IResourceController (e.g: before_update)
+    # we need to account for the fact that data_dict might be a package
+    # and might be a resource.
+    # Refs https://github.com/okfn/ckanext-unhcr/pull/459
+
+    def after_create(self, context, data_dict):
+        if 'owner_org' in data_dict and 'package_id' not in data_dict:
+            self._package_after_create(context, data_dict)
+
+        if 'resource_type' in data_dict and 'package_id' in data_dict:
+            self._resource_after_create(context, data_dict)
+
+    def _package_after_create(self, context, pkg_dict):
+        if not context.get('job') and not context.get('defer_commit'):
+            if pkg_dict.get('state') == 'active':
+                toolkit.enqueue_job(jobs.process_dataset_on_create, [pkg_dict['id']])
+
+        if pkg_dict.get('type') == 'deposited-dataset':
             user_id = None
             if context.get('auth_user_obj'):
                 user_id = context['auth_user_obj'].id
@@ -464,17 +480,38 @@ class UnhcrPlugin(
                     {'ignore_auth': True}, {'id': context['user']})
                 user_id = user['id']
             if user_id:
-                helpers.create_curation_activity('dataset_deposited', data_dict['id'],
-                    data_dict['name'], user_id)
+                helpers.create_curation_activity('dataset_deposited', pkg_dict['id'],
+                    pkg_dict['name'], user_id)
 
-    def after_delete(self, context, data_dict):
+    def _resource_after_create(self, context, res_dict):
         if not context.get('job'):
-            toolkit.enqueue_job(jobs.process_dataset_on_delete, [data_dict['id']])
+            if res_dict.get('state') == 'active':
+                toolkit.enqueue_job(jobs.process_dataset_on_update, [res_dict['package_id']])
+
 
     def after_update(self, context, data_dict):
+        if 'owner_org' in data_dict and 'package_id' not in data_dict:
+            self._package_after_update(context, data_dict)
+
+        if 'resource_type' in data_dict and 'package_id' in data_dict:
+            self._resource_after_update(context, data_dict)
+
+    def _package_after_update(self, context, pkg_dict):
+        if not context.get('job') and not context.get('defer_commit'):
+            if pkg_dict.get('state') == 'active':
+                toolkit.enqueue_job(jobs.process_dataset_on_update, [pkg_dict['id']])
+
+    def _resource_after_update(self, context, res_dict):
         if not context.get('job'):
-            if data_dict.get('state') == 'active':
-                toolkit.enqueue_job(jobs.process_dataset_on_update, [data_dict['id']])
+            if res_dict.get('state') == 'active':
+                toolkit.enqueue_job(jobs.process_dataset_on_update, [res_dict['package_id']])
+
+
+    def after_delete(self, context, data_dict):
+        if 'owner_org' in data_dict and 'package_id' not in data_dict:
+            if not context.get('job'):
+                toolkit.enqueue_job(jobs.process_dataset_on_delete, [data_dict['id']])
+
 
     # IAuthFunctions
 
--- a/ckanext/unhcr/tests/test_actions.py
+++ b/ckanext/unhcr/tests/test_actions.py
@@ -1762,3 +1762,145 @@ class TestClamAVActions(base.FunctionalTestBase):
                     "metadata": {}
                 }
             )
+
+
+class TestHooks(base.FunctionalTestBase):
+
+    def setup(self):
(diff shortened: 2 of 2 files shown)


ckanext/unhcr/plugin.py
```diff
@@ -358,7 +358,7 @@ def _children_name_list(children):
         return search_params
 
     def after_create(self, context, data_dict):
-        if not context.get('job'):
+        if not context.get('job') and not context.get('defer_commit'):
```
chris48s, line 361: `package_create` does support `defer_commit: True` but as far as I can tell, nothing ever calls it with that arg

ckanext/unhcr/plugin.py
```diff
@@ -379,7 +379,7 @@ def after_delete(self, context, data_dict):
             toolkit.enqueue_job(jobs.process_dataset_on_delete, [data_dict['id']])
```
chris48s, line 379: Delete actions don't support `defer_commit: True` so I haven't changed anything here.

ckanext/unhcr/plugin.py
```diff
@@ -379,7 +379,7 @@ def after_delete(self, context, data_dict):
...
 
     def after_update(self, context, data_dict):
-        if not context.get('job'):
+        if not context.get('job') and not context.get('defer_commit'):
```
chris48s, line 382: There are two actions in CKAN core which call `package_update` with `defer_commit: True`:

- `resource_create`
- `resource_update`

ckanext/unhcr/plugin.py
```diff
@@ -503,3 +503,17 @@ def get_blueprint(self):
...
+class UnhcrResourcePlugin(plugins.SingletonPlugin):
+    plugins.implements(plugins.IResourceController, inherit=True)
+
+    def after_create(self, context, resource):
+        if not context.get('job'):
+            if resource.get('state') == 'active':
+                toolkit.enqueue_job(jobs.process_dataset_on_update, [resource['package_id']])
+
+    def after_update(self, context, resource):
+        if not context.get('job'):
```
chris48s, line 517: Neither `resource_create` nor `resource_update` respect `defer_commit: True`. They both always commit.

ckanext/unhcr/plugin.py
```diff
@@ -503,3 +503,17 @@ def get_blueprint(self):
...
+            if resource.get('state') == 'active':
+                toolkit.enqueue_job(jobs.process_dataset_on_update, [resource['package_id']])
```
chris48s, line 519: I haven't bothered implementing a `after_delete()` hook because `resource_delete` doesn't call package_update with `defer_commit: True`

amercader:
@chris48s This is messy, but what we are dealing with in CKAN core is even messier so happy to get this merged.

I wonder if once we migrate to 2.9 `package_revise` can help us with the concurrent edit support:

https://docs.ckan.org/en/2.9/api/index.html#ckan.logic.action.update.package_revise
https://github.com/ckan/ckan/pull/4618

> In order to do this I've had to define a second plugin due to the name conflict between the IPackageController and IResourceController hooks, so there is a related PR over at okfn/docker-ckan-unhcr-aws#25 to add the new plugin.

Do you get an actual exception if two interfaces implement the same hook? I can't recall. If not can we duck-type the data_dict to decide if the hook was called with a dataset or a resource?

PR #468, opened 2021-01-19
Fix EDD file uploads

chris48s:
Following on from #447 I missed some permissions we needed to grant to external users


ckanext/unhcr/plugin.py
```diff
@@ -48,9 +54,11 @@
     'resource_show',
     'resource_update',
     'resource_view_list',
+    'scan_submit',
     'site_read',
     'tag_autocomplete',
     'tag_list',
+    'task_status_show',
```
chris48s, line 61: I think it is ok to grant `task_status_show` to external users, but I can look at doing it with `ignore_auth: True` instead if you want

ckanext/unhcr/plugin.py
```diff
@@ -29,6 +29,12 @@
 
 
 ALLOWED_ACTIONS = [
+    'cloudstorage_initiate_multipart',
+    'cloudstorage_upload_multipart',
+    'cloudstorage_finish_multipart',
+    'cloudstorage_abort_multipart',
+    'cloudstorage_check_multipart',
+    'cloudstorage_clean_multipart',
```
chris48s, line 37: refs https://github.com/okfn/ckanext-cloudstorage/pull/6

ckanext/unhcr/plugin.py
amercader: I think it's safe to allow this

amercader (approved):
Looks good!

PR #469, opened 2021-01-20
EDD datapusher fixes

chris48s:
I missed these in local dev as I don't usually run with datapusher working locally :facepalm:


amercader (approved):
Makes sense

PR #473, opened 2021-01-26
Start migrating tests to pytest

chris48s:
We've already talked about this a couple of times, so there shouldn't be any big surprises here, but this PR splits the test suite into two: some to run with nose and some to run with pytest. This will allow us to migrate the test suite in small chunks instead of having to do it all at once.
I've started off by migrating `test_utils` and `test_schema` just because they are quite small modules. We also discovered some other fun things we need to deal with along the way :smile:


ckanext/unhcr/tests/pt/test_schema.py
```diff
@@ -61,16 +60,19 @@ def test_data_file_fields(self):
 
         dataset['resources'] = [resource]
 
-        with assert_raises(ValidationError) as e:
+        with pytest.raises(toolkit.ValidationError) as e:
             call_action('package_update', {'user': self.sysadmin['name']}, **dataset)
+        # workaround for DetachedInstanceError
+        # TODO: remove this when we upgrade to CKAN 2.9
+        model.Session.refresh(model.Session.revision)
```
chris48s, line 67: I settled on this as a workaround for the `DetachedInstanceError` we were getting. Maybe there's a better way..

ckanext/unhcr/tests/pt/test_utils.py
```diff
@@ -38,7 +40,13 @@ def test_resource_is_blocked_task_status_ok(self):
             url_type='upload',
         )
         toolkit.get_action('task_status_update')(
-            {'ignore_auth': True},
+            {
+                'ignore_auth': True,
+                # task_status_update wants a user object
+                # for no reason, even with 'ignore_auth': True
+                # give it an empty string to keep it happy
+                'user': ''
```
chris48s, line 48: In core, `task_status_update` requires `user` to be in `context` even if we're calling it with `'ignore_auth': True` https://github.com/ckan/ckan/blob/c04b6d64aadc7a647b8896279002626ca3be0a56/ckan/logic/action/update.py#L761
This wasn't an issue in the nose tests because it `context['user']` was getting populated with some value so I'm just passing a junk value here to work round it. Is this actually a bug in core (happy to submit a PR if it is) or should we always be calling actions with a user object even when we're passing `'ignore_auth': True` ?

amercader: I think in general actions (or more specifically their auth functions) should not need `context['user']` if you are passing `ignore_auth: True`. In this case the action doesn't even use it, and the auth function just uses it for the message. So I think it can be removed from the action with a PR. But is not a big deal either way.

amercader:
Looking good so far!

PR #474, opened 2021-01-27
Add organization_list_all_fields action

### LINKED ISSUE #418: Optimise calls to organization_list / all_fields:True
As noted in

https://github.com/okfn/ckanext-unhcr/pull/414#discussion_r494977929
https://github.com/okfn/ckanext-unhcr/pull/394#discussion_r495801074

There are several places in the codebase where we are calling

```py
toolkit.get_action('organization_list')(
    context,
    {'type': 'data-container', 'all_fields': True}
)
```

which is a slow operation with lots of containers. Can we write a custom action to improve the performance of these calls?

chris48s:
This PR adds a customised `organization_list` action.

```py
toolkit.get_action('organization_list_all_fields')(context, {})
```

does roughly the same thing as

```py
toolkit.get_action('organization_list')(
    context,
    {'type': 'data-container', 'all_fields': True, 'include_extras': True, 'order_by': 'title'}
)
```

(or at least the bits we need) but the call is much faster. This massively improves the render time of pages that call it :)

Closes #418

### DIFF
--- a/ckanext/unhcr/actions.py
+++ b/ckanext/unhcr/actions.py
@@ -7,6 +7,7 @@ from urlparse import urljoin
 from dateutil.parser import parse as parse_date
 from sqlalchemy import and_, desc, or_, select
 from sqlalchemy.dialects.postgresql import array
+from sqlalchemy.orm import aliased
 from ckan import model
 from ckan.authz import has_user_permission_for_group_or_org
 from ckan.plugins import toolkit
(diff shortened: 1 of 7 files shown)


ckanext/unhcr/tests/pt/actions/test_organization_actions.py
```diff
@@ -0,0 +1,47 @@
+# -*- coding: utf-8 -*-
```
chris48s, line 1: The location of this file alludes to some future refactoring I want to do (some of which we can probably do as we review everything for the CKAN 2.9 migration). I think I want to split `actions.py`, `auth.py`, `test_actions.py` and `test_auth.py` up into smaller chunks (like in core) so I'm starting that structure here.

ckanext/unhcr/actions.py
```diff
@@ -275,6 +276,82 @@ def organization_member_delete(context, data_dict):
...
+    )
+    result = session.execute(sql).fetchall()
+
+    organization_plugin = lib_plugins.lookup_group_plugin(group_type)
+    schema = organization_plugin.form_to_db_schema()
+
+    out_list = []
+    for row in result:
+        raw_dict = {k:v for k,v in row.items()}
+        validated_dict, errors = lib_plugins.plugin_validate(
```
amercader, line 340: When using ckanext-scheming you can actually do `organization_plugin.validate(...)` but it doesn't make much difference.

amercader (commented):
Looks good @chris48s

PR #476, opened 2021-01-28
update schemas for scheming 2

chris48s:
We now need to specify the height of multi-select boxes, otherwise they will be massive because the the `multiple_select` template has changed a bit.
refs https://github.com/okfn/docker-ckan-unhcr-aws/pull/27

### DIFF
--- a/ckanext/unhcr/schemas/deposited_dataset.json
+++ b/ckanext/unhcr/schemas/deposited_dataset.json
@@ -1,961 +1,962 @@
 {
-    "about": "CKAN dataset schema customized for the UN HCR RIDL project",
-    "scheming_version": 1,
-    "dataset_type": "deposited-dataset",
-    "about_url": "http://github.com/okfn/ckanext-unhcr",
+    "about": "CKAN dataset schema customized for the UN HCR RIDL project", 
+    "scheming_version": 1, 
+    "dataset_type": "deposited-dataset", 
+    "about_url": "http://github.com/okfn/ckanext-unhcr", 
     "dataset_fields": [
         {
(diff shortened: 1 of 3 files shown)


ckanext/unhcr/schemas/deposited_dataset.json
```diff
@@ -1,962 +1,963 @@
 {
-    "about": "CKAN dataset schema customized for the UN HCR RIDL project",
```
amercader, line 2: Can we avoid the whitespace changes here?

amercader (approved):

ckanext/unhcr/schemas/deposited_dataset.json
chris48s: Back in https://github.com/okfn/ckanext-unhcr/pull/365 I made some schema changes. At that point I manually edited both schemas and "helpfully" cleaned up the whitespace in `deposited_dataset.json`.

In retrospect that was a mistake because `deposited_dataset.json` is/should be generated from `dataset.json` by running https://github.com/okfn/ckanext-unhcr/blob/master/scripts/generate_deposited_dataset_schema.py What I should have done in that PR was:
- modified `deposited_dataset.json`
- run `scripts/generate_deposited_dataset_schema.py` and
- left the whitespace alone.

In this PR, I just modified `dataset.json` and ran `scripts/generate_deposited_dataset_schema.py` which un-does all the whitespace changes I made in https://github.com/okfn/ckanext-unhcr/pull/365/commits/00a7556c0b3fc945e1e23878092fd9f0393db3b9 but I think it is actually more useful that `deposited_dataset.json` is just the output of running `scripts/generate_deposited_dataset_schema.py`.

PR #480, opened 2021-02-08
don't allow >1 user account with the same email

### LINKED ISSUE #479: Don't allow >1 user accounts with the same email
At the moment the external user signup form allows more than one user account with the same email address. We should require this to be unique at signup time

chris48s:
closes  #479

### DIFF
--- a/ckanext/unhcr/actions.py
+++ b/ckanext/unhcr/actions.py
@@ -1289,6 +1289,17 @@ def user_show(up_func, context, data_dict):
 
 @toolkit.chained_action
 def user_create(up_func, context, data_dict):
+
+    m = context.get('model', model)
+    if len(m.Session.query(m.User).filter(m.User.email==data_dict['email']).all()) > 0:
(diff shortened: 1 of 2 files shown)


amercader (approved):
Looks good.

If an external user request gets rejected, does that purge the user or just mark it as deleted? (just flagging because then the same user could not re-apply with the same email)

chris48s:
Good question.
When we reject a request we just set the status of the account to `m.State.DELETED`. We don't delete the record so this would prevent re-applying with the same email.

I guess the options are:
- Leave it as it is: Once rejected, always rejected
- Completely delete the user record if rejected
- We could make the condition `.filter(_and(m.User.email==data_dict['email'], m.User.state==m.State.ACTIVE))` or whatever, but I guess we should really implement whatever CKAN 2.9 is going to do as we will inherit that behaviour soon. What assumption is made upstream?

chris48s:
Will do a follow-up PR to switch to completely deleting the user record

PR #483, opened 2021-02-12
Migrate some helper tests from nose to pytest

chris48s:
Just gradually picking away at this job..


ckanext/unhcr/tests/pt/test_helpers.py
```diff
@@ -0,0 +1,522 @@
...
+                    exclude_ids=[deposit['id']],
+                    dataset={'creator_user_id': container_admin['id']}
+                )
+            )
+        )
+
+        # TODO: this test_request_context() is a workaround for CKAN 2.8
+        # you should be able to remove this when we upgrade to CKAN 2.9
+        with test_request_context():
+            toolkit.c.userobj = None
```
chris48s, line 141: The problem that this works round is that https://github.com/ckan/ckan/blob/ckan-2.8.7/ckan/authz.py#L385-L399 throws `AttributeError`. I think in 2.9 this problem goes away
https://github.com/ckan/ckan/blob/ckan-2.9.2/ckan/authz.py#L401-L409
https://github.com/ckan/ckan/blob/ckan-2.9.2/ckan/authz.py#L150-L172
I suspect we're going to hit this a lot as we migrate the test suite though.

I guess the other option is to apply a core patch which makes `get_user_id_for_username()` handle the `AttributeError` instead of having to wrap loads of stuff with

```py
with test_request_context():
    toolkit.c.userobj = None
```

amercader: @chris48s let's patch 2.8.x in core so `get_user_id_for_username()` catches an `AttributeError`

chris48s: Cool. I went ahead and did that in
https://github.com/okfn/docker-ckan-unhcr-aws/pull/28
https://github.com/okfn/ckanext-unhcr/pull/485
and I've removed the workaround from this PR and #484 and rebased them both onto master

amercader (approved):

PR #484, opened 2021-02-16
Migrate validator tests from nose to pytest

chris48s:
Mmm. Low hanging fruit
The sweetest fruit of all :pineapple:

### DIFF
--- a/ckanext/unhcr/tests/pt/test_validators.py
+++ b/ckanext/unhcr/tests/pt/test_validators.py
@@ -1,126 +1,123 @@
-import os
-import pylons
+# -*- coding: utf-8 -*-
+
+import pytest
 from ckan import model
 from ckan.plugins import toolkit
-from paste.registry import Registry
-from nose.plugins.attrib import attr
 import ckan.lib.navl.dictization_functions as df
-from ckan.tests import factories as core_factories
-from nose.tools import assert_raises, assert_equals
 from ckan.tests.helpers import call_action
-from ckanext.unhcr.tests import base, factories
+from ckantoolkit.tests import factories as core_factories
+from ckanext.unhcr.tests import factories
 from ckanext.unhcr import validators
 
 
-class TestValidators(base.FunctionalTestBase):
-
-    # Config
-
-    @classmethod
-    def setup_class(cls):
-
-        # Hack because the hierarchy extension uses c in some methods
-        # that are called outside the context of a web request
-        c = pylons.util.AttribSafeContextObj()
-        registry = Registry()
-        registry.prepare()
-        registry.register(pylons.c, c)
-
-        super(TestValidators, cls).setup_class()
-
-    def setup(self):
-        super(TestValidators, self).setup()
-        self.normal_user = core_factories.User()
-        self.sysadmin = core_factories.Sysadmin()
-
-    # Deposited Datasets
+@pytest.mark.usefixtures(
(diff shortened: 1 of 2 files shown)


ckanext/unhcr/tests/pt/test_validators.py
```diff
@@ -131,43 +127,57 @@ def test_visibility_validator_restricted_false(self):
...
+)
+class TestOwnerOrgValidator(object):
+
+    """
+    TODO: every test in this class wraps the call to
+    validators.owner_org_validator(...) in
+    with test_request_context():
+        toolkit.c.userobj = None
+    This is a workaround for CKAN 2.8
+    We should be able to remove this when we upgrade to CKAN 2.9
```
chris48s, line 177: Same issue as https://github.com/okfn/ckanext-unhcr/pull/483#discussion_r575152701 again

ckanext/unhcr/tests/pt/test_validators.py
```diff
@@ -284,21 +295,30 @@ def test_owner_org_validator_dataset_collaborator_move_dataset(self):
...
+)
+class TestDepositedDatasetCurationIdValidator(object):
+
+    """
+    TODO: every test in this class wraps the call to
+    validators.deposited_dataset_curator_id(...) in
+    with test_request_context():
+        toolkit.c.userobj = None
+    This is a workaround for CKAN 2.8
+    We should be able to remove this when we upgrade to CKAN 2.9
```
chris48s, line 318: Same issue as https://github.com/okfn/ckanext-unhcr/pull/483#discussion_r575152701 again

I think given how often we are hitting this, a core patch for this one function is probably the way forward here

ckanext/unhcr/validators.py
```diff
@@ -276,9 +276,12 @@ def owner_org_validator(key, data, errors, context):
                 try:
                     return validators.owner_org_validator(key, data, errors, context)
                 except Invalid as e:
-                    if e.error == u'You cannot add a dataset to this organization':
+                    if e.error in (
+                        u'You cannot add a dataset to this organization',
+                        u'You cannot add a dataset to this data container',
```
chris48s, line 281: I've decided not to dig too deeply into why this error is sometimes `You cannot add a dataset to this organization` and sometimes `You cannot add a dataset to this data container` depending on the environment and just accept it :see_no_evil: 
Hopefully doing this will keep it from unexpectedly biting us later

amercader: lol is this a local / github actions inconsistency? or when running individual tests vs all the suite?

chris48s: Nope. If you run the test under nose the exception thrown at that point in the code is 'You cannot add a dataset to this organization'. Run the same test with pytest and the exception is 'You cannot add a dataset to this data container'
I tested interactively and I get 'You cannot add a dataset to this data container'
I was sure this feature worked, but maybe the feature was actually broken and only test was passing :upside_down_face: 

Maybe we should change it to just look for 'You cannot add a dataset to this data container' as pytest and interactive usage seem to 'agree' but I can't shake the feeling that maybe this is dependent on some non-obvious bit of context. Given they both basically effectively mean the same thing, I'm inclined to accept both.

amercader: This points to the scheming plugin not being loaded at this point for some reason in nose. Let's accept both and once we are on 2.9 settle on whatever is returned there.

amercader (approved):

PR #489, opened 2021-02-22
migrate auth tests to pytest

chris48s:
Baby steps here :baby: This is one of the larger test modules.


ckanext/unhcr/tests/pt/test_auth.py
```diff
@@ -598,41 +588,41 @@ def test_unit_user_is_not_container_admin(self):
...
     @helpers.change_config('ckan.auth.create_unowned_dataset', False)
-    def test_integration_new_deposit(self):
+    def test_integration_new_deposit(self, app):
         # everyone can create datasets in the data-deposit
         external_user = factories.ExternalUser()
-        resp = self.app.get(
+        resp = app.get(
             url='/deposited-dataset/new',
             extra_environ={'REMOTE_USER': external_user['name'].encode('ascii')},
             status=200,
```
chris48s, line 611: At the moment we're only running the tests on CKAN 2.8 so I'm just calling `app.get()` directly. Once we get to a stage where we are running on 2.9 I'll revisit all these and either migrate them or set up a compatibility wrapper, depending on how we end up doing it. Lets cross that bridge once we are a bit further down this road.

ckanext/unhcr/tests/pt/test_auth.py
```diff
@@ -598,41 +588,41 @@ def test_unit_user_is_not_container_admin(self):
...
+        assert not result['success']
 
         container2 = factories.DataContainer()
         result = auth.package_create(
             {'user': 'creator'}, {'owner_org': container2['name']}
         )
-        assert_equals(result['success'], False)
+        assert not result['success']
 
     def test_unit_no_data_dict(self):
```
chris48s, line 599: Not sure what to do with this one. This test is failing because
under nose `toolkit.request.path` is not set (which is what we want for this test - we're trying to test what happens when there's no `data_dict` and no request in context)
under pytest it is set and I don't seem to be able to unset it

ckanext/unhcr/tests/pt/test_auth.py
```diff
@@ -598,41 +588,41 @@ def test_unit_user_is_not_container_admin(self):
...
 
         internal_user = core_factories.User()
-        resp = self.app.get(
+        resp = app.get(
             url='/deposited-dataset/new',
             extra_environ={'REMOTE_USER': internal_user['name'].encode('ascii')},
             status=200,
         )
 
     @helpers.change_config('ckan.auth.create_unowned_dataset', False)
```
chris48s, line 621: This group of 3 tests that use `@helpers.change_config` all fail with:

```
___________________ TestPackageCreateAuth.test_integration_edit_deposit ___________________

args = (<test_auth.TestPackageCreateAuth object at 0x7f92eb722690>,), kwargs = {}

    @functools.wraps(func)
    def wrapper(*args, **kwargs):
        with changed_config(key, value):
>           return func(*args, **kwargs)
E           TypeError: test_integration_edit_deposit() takes exactly 2 arguments (1 given)

src_extensions/ckan/ckan/tests/helpers.py:389: TypeError
```

The decorator works if the test signature is just

```py
def test_integration_edit_deposit(self):
```

but then, of course, I don't have the test app in context. Any tips on how to get `@helpers.change_config` to play nice with a function that also takes a fixture param?

amercader: The pytest version of `@helpers.change_config` is `@pytest.mark.ckan_config`. See eg [here](https://github.com/ckan/ckanext-dcat/blob/6b7ec505f303fb18e0eebcebf67130d36b3dca82/ckanext/dcat/tests/test_logic.py#L72)

ckanext/unhcr/tests/pt/test_auth.py
amercader: That's weird. I can't reproduce it with this single test (I always get a `TypeError`). Maybe it's because you use the `app` fixture before on other tests and did some requests and the request object is hanging around. Can you make the checking in `package_create` more specific to account for this case? eg `toolkit.request.path` is not None` or something

chris48s: Hmm. How are you running just the one test?
If I run it on its own with `-k` I get the same result as running the full test suite:

```
ckanext-unhcr$ docker-compose -f ../../docker-compose.yml exec ckan-dev pytest --disable-warnings -s --ckan-ini=/srv/app/src_extensions/ckanext-unhcr/test.ini /srv/app/src_extensions/ckanext-unhcr/ckanext/unhcr/tests/pt/test_auth.py -k "TestPackageCreateAuth and test_unit_no_data_dict"
=================================== test session starts ===================================
platform linux2 -- Python 2.7.18, pytest-4.6.11, py-1.10.0, pluggy-0.13.1
rootdir: /srv/app/src_extensions/ckanext-unhcr
plugins: ckan-0.0.12, cov-2.11.1
collected 26 items / 25 deselected / 1 selected                                           

src_extensions/ckanext-unhcr/ckanext/unhcr/tests/pt/test_auth.py F

======================================== FAILURES =========================================
______________________ TestPackageCreateAuth.test_unit_no_data_dict _______________________

self = <test_auth.TestPackageCreateAuth object at 0x7f7d4f78d510>

    def test_unit_no_data_dict(self):
        creator = core_factories.User(name='creator')
        result = auth.package_create({'user': 'creator'}, None)
>       assert not result['success']
E       assert not True

src_extensions/ckanext-unhcr/ckanext/unhcr/tests/pt/test_auth.py:602: AssertionError
```

and if I dump `toolkit.request.path` in `package_create`, I've got `'/'` not `None`.

ckanext/unhcr/tests/pt/test_auth.py
chris48s: nice - thanks :+1: 

I knew it seemed wrong that I could only find a couple of examples of tests using `helpers.change_config` in core.
Presumably those occurrences in core (e.g: https://github.com/ckan/ckan/blob/91fc8e6a19e556689fa2ffa21ef22c635474a228/ckan/tests/logic/action/test_update.py#L1211 https://github.com/ckan/ckan/blob/cc000d3acf7401957d8321be53e8cabc2d9ebf3a/ckanext/example_idatastorebackend/test/test_plugin.py#L48-L49 ) should also be updated to use `@pytest.mark.ckan_config`

amercader: Yes, these should be updated in core (and `helpers.change_config()` removed)

ckanext/unhcr/tests/pt/test_auth.py
amercader: You are right, this fails consistently.

You are getting a request set in this test because the test class uses the `with_request_context` fixture, which creates a fake request. Perhaps the cleanest fix is to not apply this at the class level and apply it only to the individual tests that needed (unless there is a way in pytest to say dont_apply_this_class_fixture_to_this_particular_test

chris48s: I see.

Tbh none of the tests in this class really ideally _need_ `with_request_context` as such. It is just there so we can call `factories.DataContainer` in the `setup()`.

`factories.DataContainer` calls `organization_create`, which calls a whole bunch of stuff in the innards of CKAN which eventually ends up in a call to `package_search`. That lands us here:
https://github.com/okfn/ckanext-unhcr/blob/87560b1a27a24f6bc32a52173f577e5328e51196/ckanext/unhcr/plugin.py#L428
which assumes we have a web request in context and throws `TypeError: No object (name: tmpl_context or C) has been registered for this thread` if we don't. If we can build the test objects without a request then we don't need the fixture at all (and we can probably drop it in a bunch of other places too - maybe I'm using this fixture as a crutch too much?).
So I've pushed d3c3f36 as a workaround for that, but maybe there is a nicer way we could do it?

chris48s:
Gah d3c3f36 got this passing locally. I'll look at it again tomorrow _sigh_

chris48s:
Right the reason this is failing in CI is because the version of ckanext-hierarchy I'm running locally isn't the same one we're actually using and I've fixed this locally so I need to prioritize finalizing hierarchy and updating that ahead of this and then I can circle back to this..

chris48s:
TODO: update once https://github.com/okfn/ckanext-unhcr/pull/504 and https://github.com/okfn/docker-ckan-unhcr-aws/pull/32 merged

chris48s:
w00p. We can now call `factories.DataContainer()` without a web request :tada: 
I think there are probably now a whole bunch of other places where we can get away with removing the `with_request_context` fixture but I think I'm going to say this PR is "done" for now and then once we've merged the rest of the test suite migration PRs I will go back over everything and review that as another task.

PR #491, opened 2021-02-26
Migrate access request action tests from nose to pytest

chris48s:


ckanext/unhcr/actions.py
```diff
@@ -928,21 +941,21 @@ def access_request_list_for_user(context, data_dict):
...
-                AccessRequest.object_type == "organization",
-                AccessRequest.object_id.in_(containers),
+                access_requests_table.c.object_type == "organization",
+                access_requests_table.c.object_id.in_(containers),
             ),
             and_(
-                AccessRequest.object_type == "user",
-                AccessRequest.data["default_containers"].has_any(array(containers)),
+                access_requests_table.c.object_type == "user",
+                access_requests_table.c.data["default_containers"].has_any(array(containers)),
```
chris48s, line 953: I still can't fully explain why, but the way this function was written before works in a browser and it works under nose, but when I moved the tests over to pytest I started getting `ProgrammingError: (psycopg2.ProgrammingError) table name "access_requests" specified more than once`. I guess it is probably to do with the test setup as opposed to the test runner in itself but I wasn't able to get to the bottom of it.
In the end I was able to solve this by rewriting the action to consistently use the SQLAlchemy Core syntax everywhere instead of mixing-and-matching the ORM syntax and Core syntax (and I ended up fiddling with a couple of other things while I was digging into it). Everything works, but it is not 100% satisfying as I can't really explain the root issue.

chris48s:
Gah I've got a "tests pass in isolation but fail when you run them all together" issue to debug here too :grimacing:

ckanext/unhcr/actions.py
amercader: OK, if we were mixing both styles it's good to consolidate in one anyway. I guess we are not using full ORM syntax because all the joins and complex queries. That sounds good to me.

amercader (approved):

ckanext/unhcr/actions.py
chris48s: yes I try to use the ORM syntax where possible but it sometimes it is necessary to drop back to the core query style

PR #498, opened 2021-03-01
improve display of license/external access level

### LINKED ISSUE #490: dodgy table formatting for "license"/"external access level" fields
imgpsh_fullsize_anim

chris48s:
closes #490

What I would ideally like to do here is put

```jinja
{% block resource_license %}
  <tr>
    <th scope="row">External Access Level</th>
    <td>
      <span>{{ h.get_choice_label('external_access_level', pkg.external_access_level) or 'Not available' }}</span>
    </td>
  </tr>
{% endblock %}
```

in `read_base.html` to override
https://github.com/ckan/ckanext-scheming/blob/a53ef5cc2e1af36060330101efab1c058e3d5cf3/ckanext/scheming/templates/scheming/package/resource_read.html#L40-L45

Unfortauntely because of the order we load plugins in, scheming overrides our templates rather than vice-versa. If I start changing the order we load plugins in, all kind of other crazy stuff happens. As such, I think this is the best tradeoff we can really make here :/ That said, if you've got any other ideas...

### DIFF
--- a/ckanext/unhcr/templates/package/snippets/curation_sidebar.html
+++ b/ckanext/unhcr/templates/package/snippets/curation_sidebar.html
@@ -153,6 +153,6 @@
   </div>
 
   {# License #}
-  {% snippet "snippets/license.html", pkg_dict=pkg %}
+  {% snippet "package/snippets/external_access.html", pkg_dict=pkg %}
 
 </div>
--- a/ckanext/unhcr/templates/package/snippets/external_access.html
+++ b/ckanext/unhcr/templates/package/snippets/external_access.html
@@ -1,0 +1,6 @@
+<section class="module module-narrow module-shallow license">
+  <h2 class="module-heading"><i class="fa fa-lock"></i>External Access Level</h2>
+  <p class="module-content">
+    <span>{{ h.get_choice_label('external_access_level', pkg_dict.external_access_level) or 'Not available' }}</span>
+  </p>
+</section>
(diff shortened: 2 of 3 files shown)


amercader:
If I'm not mistaken we are [loading `unhcr` before `scheming_datasets`](https://github.com/okfn/docker-ckan-unhcr-aws/blob/edcc55a7439ec5fe6110473828bd88582c6f7881/ckan/environment/base.env#L31) so you should be able to override scheming templates from unhcr. Maybe it's one of this annoying Jinja2 block scope issues. Have you tried to override `/package/resource_read.html` and not `read_base.html`?

chris48s:
I might have said `read_base.html` in the top post and meant `resource_read.html`. I can't remember now, but I just tested again and I can't override that block in `resource_read.html`.

My reading of the debug info is that the last template in the inheritance chain is scheming's and not ours:

Screenshot at 2021-03-08 10-07-02

amercader (approved):
We are definititely overriding scheming templates from unchr like [this one](https://github.com/okfn/ckanext-unhcr/blob/master/ckanext/unhcr/templates/scheming/organization/group_form.html) or the core `form_snippets` (owner_org.html, tags.html...)

I don't know what's the deal with the resource template but your changes are fine so let's not bother more about this

chris48s:
Hmm. It is still unsatisfying to not fully understand this but this is better than what is on master at the moment so :shrug:

PR #499, opened 2021-03-05
migrate data deposit controllers and tests

chris48s:
Several things going on in this PR:

f688cf9 - migrate the data deposit routes from pylons controllers to flask blueprints (this bit was actually very straightforward)

71c18ff - Convert `GET` endpoints that change state to `POST`, assemble form targets with `url_for()` instead of hard-coding  (bit of trying to leave things better than I found them here)

Then 92d3b4c is just there to separate "moving things from one file to another" from "actually making functional changes" to make things easier to review. This means that you can look at the diff for a0a6087 and see _what actually changed_ :tada: 

This is where it gets ugly. There's basically 2 categories of tests we're dealing with here. There's ones where we're just doing something to the effect of 

```py
for user in [list of users]:
    self.try_to_do_thing_to_dataset(user, 403)  # make sure we can't do the thing
```

for these ones we can run that whole thing in one test because we never change the dataset object (this is faster than one test per user).

The other pattern is where we do something to the effect of:

```py
for user in [list of users]:
    self.try_to_do_thing_to_dataset(user, 200)
```

In this case, because for each user we actually successfully change the dataset object we want to reset it after each iteration so I've written those ones as 

```py
@pytest.mark.parametrize("user", [list of users])
def test_thingy(self, user):
    self.try_to_do_thing_to_dataset(user, 200)
```

This is slower, but it resets the test data after each one, which is what we need if the test modifies it.

957e723 is my attempt to try and make things run a bit quicker by reducing how much database IO we have to do for each test method. This does help a bit, but not enough. This test suite is still painfully slow and the pytest version runs way slower than it was when we were running it under nose despite my attempt to improve it :( 

Not really sure where to go with this next..


chris48s:
oh good.. and the tests fail in CI
That'll be fun to debug then :sob:

amercader:
I had a look at the speed issue. I don't think there a single magic thing that we can do to speed things up but I think we can try to refactor the tests to make them reduce calls and/or mock things and shave some seconds per test that will hopefully compound into minutes.

For instance in this type of tests:

```python
    def test_approve_draft_not_granted(self):                                                                                                                                                                                                                                        
        for user in ['sysadmin', 'depadmin', 'curator', 'creator', 'target_container_admin']:                                                                                                                                                                                        
            self.check_approve_draft_not_granted(user, 302, "action is not available")                                                                                                                                                                                               
        for user in ['depositor', 'target_container_member', 'other_container_admin']:                                                                                                                                                                                               
            self.check_approve_draft_not_granted(user, 403)                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                     
    def check_approve_draft_not_granted(self, user, status, error=None):                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                     
        # Prepare dataset                                                                                                                                                                                                                                                            
        self.patch_dataset({                                                                                                                                                                                                                                                         
            'unit_of_measurement': 'individual',                                                                                                                                                                                                                                     
            'keywords': ['3', '4'],                                                                                                                                                                                                                                                  
            'archived': 'False',                                                                                                                                                                                                                                                     
            'data_collector': ['acf'],                                                                                                                                                                                                                                               
            'data_collection_technique': 'f2f',                                                                                                                                                                                                                                      
            'external_access_level': 'open_access',                                                                                                                                                                                                                                  
        })                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                     
        # Approve dataset                                                                                                                                                                                                                                                            
        resp = self.make_request('approve', user=user, status=status)                                                                                                                                                                                                                
        if error:                                                                                                                                                                                                                                                                    
            assert error in resp.body    
```

The `patch_dataset()` call needs to be done only once if I'm understanding it correctly so it can be moved here:

```python
    def test_approve_draft_not_granted(self):                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                     
        # Prepare dataset                                                                                                                                                                                                                                                            
        self.patch_dataset({                                                                                                                                                                                                                                                         
            'unit_of_measurement': 'individual',                                                                                                                                                                                                                                     
            'keywords': ['3', '4'],                                                                                                                                                                                                                                                  
            'archived': 'False',                                                                                                                                                                                                                                                     
            'data_collector': ['acf'],                                                                                                                                                                                                                                               
            'data_collection_technique': 'f2f',                                                                                                                                                                                                                                      
            'external_access_level': 'open_access',                                                                                                                                                                                                                                  
        })                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                     
        for user in ['sysadmin', 'depadmin', 'curator', 'creator', 'target_container_admin']:                                                                                                                                                                                        
            self.check_approve_draft_not_granted(user, 302, "action is not available")                                                                                                                                                                                               
        for user in ['depositor', 'target_container_member', 'other_container_admin']:                                                                                                                                                                                               
            self.check_approve_draft_not_granted(user, 403)                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                     
    def check_approve_draft_not_granted(self, user, status, error=None):                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                     
        # Approve dataset                                                                                                                                                                                                                                                            
        resp = self.make_request('approve', user=user, status=status)                                                                                                                                                                                                                
        if error:                                                                                                                                                                                                                                                                    
            assert error in resp.body    
```

`make_request()` makes a call to `package_show` which AFAICT is not always used. All these things are not a big deal on their own but when doing so many combinations they do add up.

I did some very limited profiling (files attached in case you want to spend some time on them) and as expected we spend a lot of time doing requests, rendering templates etc. We are after all doing a full web request for every single combination. I wonder if rather than doing a full render we could mock the rendering and return just the error strings, or actually mock the render so it does nothing and use a mock of `h.flash_error()` to see what it was called with. 

I haven't tested any of these things, just thinking out loud


[test_profile.zip](https://github.com/okfn/ckanext-unhcr/files/6163759/test_profile.zip)

chris48s:
I've rebased this to resolve the merge conflicts. The build is now passing. There is probably some room to improve speed further but sorting out the recursion issue has got us down to the point where the full CI run completes in ~20 mins so I don't think I want to spend a lot of time on chasing smaller optimisations right now.

Does this still need further review or are we good to merge this one?

amercader:
Let's merge this! If there are any improvements we'll deal with them later

Issue #399, opened 2020-07-31
CKAN 2.9 upgrade


chris48s:
We're a way off starting this, but I've had a few thoughts swimming around in my mind, so now is a good time to start collecting them in an issue.

There are several categories of work to think about here:

- Filling in gaps in test coverage before we start
- Core patches
  Review core patches in https://github.com/okfn/docker-ckan-unhcr-aws/tree/master/ckan/patches/ckan
  Do we still need them? Do they need to be updated?
- Documented backwards-incompatible api changes to public API
- Un-documented backwards-incompatible changes to internals
  There are a number of places where we're patching/overriding internals - the internals may have changed and our overrides/patches will need to change too. Reviewing places where we import from core is one thing. Working out what bits of templates we're overriding is probably going to be more painful
- Removing customisations: Sometimes we've done something custom to fix a bug/add a feature that wasn't in core, or added it via a plugin. Now it is in core and we can remove the customisation/plugin
- Toolchain updates e.g:
    - Pylons controllers --> Flask blueprints
    - Custom migrations --> Alembic
    - Nose --> Pytest
    - Paster commands --> Click
    - Fanstatic --> Webassets
- Front-end updates (e.g: 2.7 --> 2.8 changed the bootstrap version. Anything like that we need to account for?)
- Plugins. We use:
    - ckanext-pdfview
    - ckanext-hierarchy
    - ckanext-facetcollapse
    - ckanext-collaborators
    - ckanext-scheming
    - ckanext-harvest
    - ckanext-ddi
    - ckanext-saml2
    - ckanext-cloudstorage (fork)
    They all need to be updated. Question: our plan is to go ckan 2.8/python 2 --> ckan 2.9/python 2 --> ckan 2.9/python 3. If a plugin we rely on drops python 2 compatibility in the first 2.9-compatible releases, does that scupper the whole thing?

amercader:
Useful info: https://github.com/ckan/ckan/wiki/Python-3-migration-guide-for-extensions

chris48s:
* In CKAN 2.9 collaborators is a native feature but the implementation differs a bit from ckanext-collaborators so some changes will need to be made around this, particularly for code that interacts a lot with collaborators like the "access requests" concept.
* There are some changes to Activities in CKAN 2.9 which we will need to accommodate in our customisations for activities
[...]

PR #500, opened 2021-03-12
switch cloudstorage out for s3filestore

chris48s:
This PR switches us from using okfn/ckanext-cloudstorage to keitaroinc/ckanext-s3filestore. Fairly minimal changes going on here but I've also taken the opportunity to get rid of a pylons controller and replace it with a flask blueprint

Refs https://github.com/okfn/docker-ckan-unhcr-aws/pull/29
Refs https://github.com/okfn/ckanext-unhcr/issues/399#issuecomment-797575581


ckanext/unhcr/plugin.py
```diff
@@ -223,11 +223,7 @@ def before_map(self, _map):
             '/deposited-dataset/{id}/resource/{resource_id}/download',
             '/deposited-dataset/{id}/resource/{resource_id}/download/{filename}',
         ]
-        if not plugins.plugin_loaded('cloudstorage'):
-            for route in download_routes:
-                _map.connect(route, controller=controller, action='resource_download')
-        else:
-            controller='ckanext.unhcr.controllers.extended_storage:ExtendedStorageController'
+        if not plugins.plugin_loaded('s3filestore'):
```
amercader, line 226: We don't need to add these Pylons routes any more now that we have a blueprint right?

amercader (approved):
@chris48s just one minor comment about routes, the rest looks great

ckanext/unhcr/plugin.py
chris48s: This block here

```py
if not plugins.plugin_loaded('s3filestore'):
    for route in download_routes:
        _map.connect(route, controller=controller, action='resource_download')
```

is adding the fallback routes for local storage. It is just there so we can develop using the local storage backend - you don't have to have an S3 bucket for your dev copy (although personally I do use a S3 bucket for my dev copy). This is still a pylons controller.


Further down in `get_blueprint()` where I am saying

```py
if plugins.plugin_loaded('s3filestore'):
    bp.append(blueprints.unhcr_s3_resource_blueprint)
```

that is adding the new blueprint for when we are using the S3 backend.

amercader: Got you

PR #501, opened 2021-03-15
Migrate a load of action tests to pytest

chris48s:
This is quite a big diff but I've split it into smaller commits so it will hopefully be easier to review.
Mostly this is just moving stuff from A to B but I will leave a couple of comments on the diff where there is something notable.


ckanext/unhcr/tests/pt/actions/test_package_actions.py
```diff
@@ -0,0 +1,256 @@
...
+@pytest.mark.usefixtures(
+    'clean_db', 'clean_index', 'with_request_context', 'unhcr_migrate'
+)
+class TestMicrodata(object):
+
+    # General
+
+    def setup(self):
+        # Config
+        toolkit.config['ckanext.unhcr.microdata_api_key'] = 'API-KEY'
```
chris48s, line 24: I did try switching this to use `@pytest.mark.ckan_config('ckanext.unhcr.microdata_api_key', 'API-KEY')` instead but for some reason it wasn't applying the setting and I don't know why 😕
Maybe we hit the same issue with `core_helpers.change_config` and that's why we were doing it like this. I don't think it is a huge deal, but if you have any ideas let me know

ckanext/unhcr/actions.py
```diff
@@ -776,7 +776,7 @@ def scan_hook(context, data_dict):
     task['value'] = json.dumps(data_dict)
     task['error'] = json.dumps(data_dict.get('error'))
 
-    task = toolkit.get_action('task_status_update')({'ignore_auth': True}, task)
+    task = toolkit.get_action('task_status_update')({'ignore_auth': True, 'user': None}, task)
```
chris48s, line 779: The two changes in this file are more workarounds for https://github.com/ckan/ckan/pull/5861
Assuming you plan to merge https://github.com/ckan/ckan/pull/5861 at some point another approach would be to solve this with a core patch instead.

chris48s:
oh good. I the tests fail in CI. I'll have a look tomorrow

ckanext/unhcr/actions.py
amercader: I merged the core PR and it should be on 2.8.8 / 2.9.3

amercader (approved):
Looks good. In general I try to avoid using 'clean_index' and 'with_request_context' unless they are really needed (ie you are testing the search in different tests or need an actual requests in your tests) Every second we can shave will help I guess

chris48s:
Right. I've rolled back the workarounds for https://github.com/ckan/ckan/pull/5861 and replaced them with a core patch which I have also pushed to the docker repo: https://github.com/okfn/docker-ckan-unhcr-aws/commit/b1a6c7492e077063a5b097f2daf05a23c13ae3c0 We can remove that once we upgrade to a CKAN version that includes it.,

I've also pushed https://github.com/okfn/ckanext-unhcr/pull/501/commits/5f74df004ed51a1233aea22194d332a1711c1ab6 to this branch as it was breaking the build here too. I'm going to merge this now and then rebase https://github.com/okfn/ckanext-unhcr/pull/499 over it.

chris48s:
Oh yeah I also have on my list to do a review of which fixtures are needed where once I resolve this issue with factories needing a request in context to build a test object https://github.com/okfn/ckanext-unhcr/pull/489#discussion_r598891470 although having solved this recursion issue I am a now slightly less concerned about the overall test performance.

Issue #315, opened 2020-06-04
Fix datastore permission tests


chris48s:
There are some disabled tests in

https://github.com/okfn/ckanext-unhcr/blob/d04287154acc73c23e88a64839a598bcf8c05110/ckanext/unhcr/tests/test_actions.py#L299-L336

try to get them working

PR #505, opened 2021-03-24
Fix datastore auth and tests

chris48s:
* fix datastore auth functions (do check the behaviour of these now matches the intent)
* run the datastore auth tests
* move datastore auth tests to pytest

It was not necessary to run the `datastore set-permissions` SQL to get these working in CI

Closes https://github.com/okfn/ckanext-unhcr/issues/315


ckanext/unhcr/auth.py
```diff
@@ -272,15 +272,32 @@ def resource_download(context, data_dict):
...
+@toolkit.chained_auth_function
+def datastore_info(next_auth, context, data_dict):
+    parent_auth = auth_datastore_core.datastore_auth(
+        context,
+        data_dict,
+        'resource_download'
+    )
+    if not parent_auth['success']:
+        return parent_auth
+    return next_auth(context, data_dict)
```
amercader, line 284: Given that the upstream `datastore_info` just calls the same `auth_datastore_core.datastore_auth()` function we are calling here, and that this function the only thing it does is call `check_access('resource_download')`, I wonder if we can simplify all these auth functions removing the chaining and just doing:

```python
def datastore_info(context, data_dict):
    return resource_download(context, {'id': data_dict['resource_id']})
```

We could use `check_access('resource_download')` but given how this is a custom ridl-specific auth function I think is probably fine.

ckanext/unhcr/tests/pt/actions/test_datastore_actions.py
```diff
@@ -0,0 +1,107 @@
...
+import mock
+from ckan import model
+from ckan.plugins import toolkit
+from ckan.tests import helpers as core_helpers
+from ckantoolkit.tests import factories as core_factories
+from ckanext.unhcr.tests import factories, mocks
+
+
+@pytest.mark.usefixtures('clean_db', 'with_request_context', 'unhcr_migrate')
+class TestDatastoreAuthRestrictedDownloads(object):
```
amercader, line 13: All assumptions here look good!

ckanext/unhcr/auth.py
chris48s: The reason I made them chained actions is because if you just try to define an auth function

```py
def datastore_info(context, data_dict):
    ...
```

without chaining, the app throws

```
Exception: The auth function 'datastore_info' is already implemented in 'unhcr'
```

I think this is because `datastore` is implemented as a plugin (even though it is in the core repo), so we have both `unhcr` and `datastore` in `ckan.plugins`:

```ini
ckan.plugins = ... unhcr datastore ...
```

so I think we do need to chain it and hand off to the next plugin.

amercader: Cool, let's chain it then, although I think that by calling first `auth_datastore_core.datastore_auth()` and then `next_auth()` we are calling the `resource_download` and `resource_read` auth functions, which should be redundant. But it's not a big deal.

chris48s: I've been back double checked this just in case we are making a pointless call and the core datastore_info/search auth functions only check `resource_show`, not `resource_download` (which is more restrictive in our case).

PR #507, opened 2021-03-24
migrate private resource tests to pytest

chris48s:
This is the last module of tests I am going to move.


ckanext/unhcr/tests/pt/test_controllers.py
```diff
@@ -1171,3 +1171,152 @@ def test_search_index_rebuild_sysadmin(self, app):
         # now package_search will tell us there is 1 dataset
         packages = toolkit.get_action('package_search')(context, data_dict)
         assert 1 == packages['count']
+
+
+@pytest.mark.usefixtures(
+    'clean_db', 'clean_index', 'with_request_context', 'unhcr_migrate'
+)
+class TestPrivateResources(object):
```
chris48s, line 1179: This test class was in test_actions but they aren't really action tests. They are mostly e2e tests using the test app so I've moved them to test_controllers.
I think once I've merged all this stuff I might split test_controllers up into some smaller files like I did with actions

ckanext/unhcr/tests/pt/test_controllers.py
```diff
@@ -1171,3 +1171,152 @@ def test_search_index_rebuild_sysadmin(self, app):
...
+
+        # We don't have data but we pass authorization
+        environ = {'REMOTE_USER': self.org_user['name'].encode('ascii')}
+        res = app.get(url, extra_environ=environ, status=404)
+
+        # We don't have data but we pass authorization
+        environ = {'REMOTE_USER': self.sysadmin['name'].encode('ascii')}
+        res = app.get(url, extra_environ=environ, status=404)
+
+    @pytest.mark.skip(reason="why is the private dataset still available in the test env?")
```
chris48s, line 1300: I've moved this test over as a skipped test because I'm still not 100% sure what to do with it but I'm happy to either fix the test or fix the code if we can decide what the correct behaviour is. This is the only skipped test

chris48s: Right. I think I have now figured out what is going on here:
When we create a dataset using the factory, the value still gets run through the actions/validators.
In the `visibility_validator`, we have this block:
https://github.com/okfn/ckanext-unhcr/blob/2808605cca8302346e75dad80cf886e887990870/ckanext/unhcr/validators.py#L187-L192
so even if we call `factories.Dataset(private=True, visibility='private')` by the time we actually save the object to the DB, the property is really `visibility='restricted'`. This is probably a sensible thing because it means if you create datasets via the API (which users are increasingly doing) you can't create a dataset with `visibility='private'` by not understanding that the "Private" label in the front end really corresponds to the `restricted` value in the API.

This behaviour is actually covered in the validator tests https://github.com/okfn/ckanext-unhcr/blob/2808605cca8302346e75dad80cf886e887990870/ckanext/unhcr/tests/pt/test_validators.py#L95 , but I've updated this test in 954580e to reflect the real behaviour, re-enabled it and added some comments to explain. I think this now makes sense, but can you check it. I may _still_ not be clear on what this is supposed to be doing.

amercader (approved):
This looks good and your comments on `test_access_visibility_private_admin` make sense. Sorry for the confusion on this

PR #510, opened 2021-03-29
test suite cleanup

chris48s:
this will probably be failing for now. Need to rebase it once #489 merged


chris48s:
TODO:
Once #505 is merged, the final bit of finishing off for this PR is to rebase on to that and then the `with_request_context` decorator can be removed from those tests too (but doing that is dependent on 608a541 so I can't do it in #505 )

amercader (approved):

chris48s:
Slight scope creep/change of plan here. Somewhere along the line I got confused and migrated the tests I was going to not bother migrating because they are definitely going to get deleted and left some tests we actually needed in the nose dir :facepalm:
Upshot of that is I still needed to move over the last remaining test class `TestExternalUserUpdateState` because those are tests we do still need so I've done that last class in this PR too. Nothing notable there: completely straightforward.
Then that means 100% of tests are now migrated to running under pytest so I've also taken the opportunity to cleanup the remnants of nose in this PR too :tada:

PR #513, opened 2021-04-13
CKAN 2.9: make the test suite a bit less broken

chris48s:
This PR targets the [ckan2.9 branch](https://github.com/okfn/ckanext-unhcr/compare/ckan2.9) which already has some changes applied to it.
Still lots of failing tests here, but more of them are failing because the code is broken rather than because the tests are broken so that's good :smile:

### DIFF
--- a/ckanext/unhcr/tests/pt/controllers/test_user.py
+++ b/ckanext/unhcr/tests/pt/controllers/test_user.py
@@ -116,7 +116,7 @@ class TestUserRegister(object):
     def test_register_success(self, app):
         mock_mailer = mock.Mock()
         with mock.patch('ckan.plugins.toolkit.enqueue_job', mock_mailer):
-            resp = app.post(toolkit.url_for('user.register'), self.payload)
+            resp = app.post(toolkit.url_for('user.register'), data=self.payload)
 
 
         # we should have created a user object with pending state
@@ -151,7 +151,7 @@ class TestUserRegister(object):
 
     def test_register_empty_message(self, app):
         self.payload['message'] = ''
-        resp = app.post(toolkit.url_for('user.register'), self.payload)
+        resp = app.post(toolkit.url_for('user.register'), data=self.payload)
         assert "&#39;message&#39; is required" in resp.body
         action = toolkit.get_action("user_show")
         with pytest.raises(toolkit.ObjectNotFound):
@@ -162,7 +162,7 @@ class TestUserRegister(object):
 
     def test_register_empty_focal_point(self, app):
         self.payload['focal_point'] = ''
-        resp = app.post(toolkit.url_for('user.register'), self.payload)
+        resp = app.post(toolkit.url_for('user.register'), data=self.payload)
         assert "A focal point must be specified" in resp.body
         action = toolkit.get_action("user_show")
         with pytest.raises(toolkit.ObjectNotFound):
@@ -173,7 +173,7 @@ class TestUserRegister(object):
 
     def test_no_containers(self, app):
         self.payload['container'] = ''
-        resp = app.post(toolkit.url_for('user.register'), self.payload)
+        resp = app.post(toolkit.url_for('user.register'), data=self.payload)
         assert "A region must be specified" in resp.body
         action = toolkit.get_action("user_show")
         with pytest.raises(toolkit.ObjectNotFound):
@@ -184,7 +184,7 @@ class TestUserRegister(object):
 
     def test_internal_user(self, app):
         self.payload['email'] = 'fred@unhcr.org'
-        resp = app.post(toolkit.url_for('user.register'), self.payload)
+        resp = app.post(toolkit.url_for('user.register'), data=self.payload)
         assert (
             "Users with an @unhcr.org email may not register for a partner account."
             in resp.body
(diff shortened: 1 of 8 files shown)


ckanext/unhcr/tests/pt/test_schema.py
```diff
@@ -97,9 +94,6 @@ def test_both_types_data_fields_missing(self):
 
         with pytest.raises(toolkit.ValidationError) as e:
             call_action('package_update', {'user': self.sysadmin['name']}, **dataset)
-        # workaround for DetachedInstanceError
-        # TODO: remove this when we upgrade to CKAN 2.9
-        model.Session.refresh(model.Session.revision)
 
         errors = e.value.error_dict['resources'][1]
```
chris48s, line 98: This test now fails for a different reason.

Under CKAN 2.8 `e.value.error_dict` here is

```py
{
    u'resources': [
        {},
        {'url': ['All data resources require an uploaded file'], u'date_range_start': [u'Missing value'], u'date_range_end': [u'Missing value'], u'version': [u'Missing value'], u'process_status': [u'Missing value'], u'identifiability': [u'Missing value']}
    ]
}
```

But then under 2.9 this is:

```py
{
    u'resources': [
        {'url': ['All data resources require an uploaded file'], u'date_range_start': [u'Missing value'], u'date_range_end': [u'Missing value'], u'version': [u'Missing value'], u'process_status': [u'Missing value'], u'identifiability': [u'Missing value']}
    ]
}
```

Is this an expected behaviour change between 2.8 and 2.9? It seems wrong.

ckanext/unhcr/tests/pt/controllers/test_user.py
```diff
@@ -15,36 +15,34 @@ class TestUserController(object):
...
 
         # promote them
         resp = app.post(
             '/user/sysadmin',
-            {'id': user['id'], 'status': '1' },
-            extra_environ=env,
-            status=302
+            data={'id': user['id'], 'status': '1' },
+            headers={"Authorization": admin['apikey'].encode('ascii')},
+            status=200,
```
chris48s, line 41: This is an interesting one. I haven't made this change to any of the other tests that use a redirect yet but having chased it through the stack it seems like there is a bug (or at least an unusual behaviour) in the old version of werkzeug we are pinned to here.

Basically, if you use `follow_redirects=True` (which CKAN 2.9 does by default https://github.com/ckan/ckan/blob/18c92bd5bac627ac28752f0d8493f34c562f9da1/ckan/tests/helpers.py#L259-L265 ), not all of the values we pass in `extra_environ` / `environ_overrides` get copied to the environ used by the second request (when we follow the redirect). Specifically relevant to us, if we pass `REMOTE_USER` in the env it gets used for the first request but then not the second one.

So:

```py
resp = app.post(
    '/user/sysadmin',
    data={'id': user['id'], 'status': '1' },
    extra_environ={'REMOTE_USER': admin['name'].encode('ascii')},
    status=200,
)
```

throws `AssertionError: Actual: 403. Expected: 200`

To work around this, I've switched to using an `Authorization` header instead of `REMOTE_USER` because headers do get preserved when we follow the redirect in `resolve_redirect` https://github.com/pallets/werkzeug/blob/0.15.5/src/werkzeug/test.py#L451 - so this:

```py
resp = app.post(
    '/user/sysadmin',
    data={'id': user['id'], 'status': '1' },
    headers={"Authorization": admin['apikey'].encode('ascii')},
    status=200,
)
```

works totally fine That gives us a workaround that I can apply to the other tests.. Another option would be to write a little wrapper that follows the redirect preserving `environ_overrides`/`extra_environ`. The thing that really surprises me here is it seems incredibly unlikely that we haven't hit this before anywhere in the test suite for CKAN core :confused: 

It is also possible this is fixed or different in a newer version of werkzeug (I've not tried it), but I assume we are pinned to this specific version for python 2 compatibility?

ckanext/unhcr/tests/pt/test_schema.py
amercader: This is the last related change in this area: https://github.com/ckan/ckan/commit/3faf857deff3b399cd4b3cce74e101fdfece4677 (Issue: https://github.com/ckan/ckan/issues/5334). If I understand the code correctly apparently is expected behaviour to just return the error dict for the resource that failed.

Although now that I think about it this was backported to 2.8 so the code should be the same. So that is probably not relevant. I don't know what's going on here.

ckanext/unhcr/tests/pt/controllers/test_user.py
amercader: Does this help here? https://github.com/ckan/ckan/issues/5748#issuecomment-771537923

chris48s: Yes :+1:  exactly this. Thanks
I suspect I now need to put a lot of those `with_request_context` fixtures I removed in https://github.com/okfn/ckanext-unhcr/commit/a237dfa8d8a6afe70f250ffbd6af5f1eed66e8ac back in :laughing:

ckanext/unhcr/tests/pt/test_schema.py
chris48s: OK. One thing that has occurred to me since I initially hit this is there are some scheming-related problems outstanding to solve which might be another variable here. I think I am going to leave this test failing for now and come back to again it once we've got that working correctly.

chris48s:
I haven't done anything with
`ckanext/unhcr/tests/pt/controllers/test_dataset.py`
`ckanext/unhcr/tests/pt/controllers/test_deposited_dataset.py`
Those also need updating, but every single test in both those modules is failing. I think its going to be easier to do something sensible with them once we have datasets and deposits working

chris48s:
still lots of stuff failing but we're getting somewhere

PR #514, opened 2021-04-13
CKAN 2.9: convert remaining paster commands to click

chris48s:
This PR targets the [ckan2.9 branch](https://github.com/okfn/ckanext-unhcr/compare/ckan2.9). I had already converted the `init-db` on there as setting up the DB was one of the bare minimum things I needed to get working. This PR finishes  the job


ckanext/unhcr/mailer.py
```diff
@@ -222,8 +222,8 @@ def get_summary_email_recipients():
     )
     curator_ids = [c[0] for c in curators]
 
-    all_users = toolkit.get_action('user_list')({ 'ignore_auth': True }, {})
     default_user = toolkit.get_action('get_site_user')({ 'ignore_auth': True })
+    all_users = toolkit.get_action('user_list')({ 'user': default_user['name'] }, {})
```
chris48s, line 226: PR comment:
This is a behavioural change between CKAN 2.8 and 2.9. In CKAN 2.8

```py
toolkit.get_action('user_list')({ 'ignore_auth': True }, {})
```

would return the user list with email addresses. In CKAN 2.9 it doesn't.

I've worked around it here by using the site user to call `user_list` but is this a bug in core or an intentional but undocumented change?

amercader: I think it was intentional to try to plug some holes that were leaking email addresses. You are going to hate me for this but can you try to use:

```python
toolkit.get_action('user_list')({ 'ignore_auth': True, 'keep_email': True }, {})
```

I know, I know... it's undocumented

chris48s: Yeah that does the job. I've updated the PR to solve using that method.
I guess the next question is: Is it undocumented because it is supposed to be secret or because it should be in the docs but isn't?

amercader: It's not supposed to be a secret. But docs are more focused on documenting the parameters that users can actually change from the API (ie the `data_dict`). This could be documented in the action docstring though (I think there are other `keep_*` flags supported in the context).

The proper fix for this is  to move this functionality to the data_dict but only enabled for sysadmins. There are loads of these context vars that need to be refactored, but that's a much bigger job.

amercader (approved):

PR #517, opened 2021-04-15
CKAN 2.9: theme fixes

chris48s:
This PR targets the [ckan2.9 branch](https://github.com/okfn/ckanext-unhcr/compare/ckan2.9).

This probably isn't comprehensive because there are a bunch of things I have not been able to test. Until we work out the scheming issues none of the dataset/resource pages or data deposit pages work yet so there might be further tweaks needed. I suggest we just merge this on the basis that it is fixing a bunch of stuff that was broken. Then once those other pages are working we can do a second pass on the theme if needed.


amercader (approved):

amercader:
Yes, let's merge it. I added a small one I bumped into

PR #518, opened 2021-04-15
CKAN 2.9: dataset and org pages

amercader:
Turns out scheming was working just fine. It was mostly the templates that needed updating and the old Pylons routes causing trouble.

Still need to fix the data-deposit route


chris48s:
Nice - thanks. I rebased this just because we already picked 8ab5d1a to the 2.9 branch with a different commit hash. Everything on here is making things less broken so I think I'm just going to merge what is here to the 2.9 branch because being able to create datasets and resources makes everything else I need to work on heaps easier :)

While I was testing this locally I did notice a few things which aren't necessarily problems with this PR but which I will quickly note here:

1. The "Data sensitivity" (dataset) field and "HXL-ated" (resource) fields are not marked as required, but they now throw a validation error if we don't fill them in. This doesn't reproduce on UAT (which is running the same scheming version on 2.8)

2. On https://ckan-dev:5000/data-container/ we are now saying
"CKAN Organizations are used to create, manage and publish..." instead of "RIDL Data Containers are used to create, manage and publish..."

Screenshot at 2021-04-15 16-51-39

Screenshot at 2021-04-15 16-51-57


3. Resource page styling needs some fixing:

Screenshot at 2021-04-15 16-52-53

4. Activity streams are broken (but we know this)

5. Under ckan 2.8
    - `/packages`
    - `/package`
    - `/packages/{id}`
    - `/package/{id}`
Were all redirects to `/dataset` or `/dataset/{id}`. Have these all been dropped in 2.9 or are these more things we need to fix?

Some of those things can be things for me to pick up

PR #520, opened 2021-04-16
CKAN 2.9: SAML2

amercader:
Basically update routes in the templates, clean up old ckanext-saml2 specific stuff and a small override on the user login form to hide a SSO button.

Refs okfn/docker-ckan-unhcr-aws#37


ckanext/unhcr/templates/login_form.html
```diff
@@ -6,15 +6,15 @@
...
+        href="{{ h.url_for('saml2auth.saml2login', came_from=h.get_came_from_param()) }}"
         title="Log in using your UNHCR Active Directory account">
         Log in as UNHCR staff
       </a>
     </p>
     <hr />
     <p>
       <a class="btn btn-inverse btn-lg"
-        href="/service/login"
+         href="{{ h.url_for('user.login') }}"
```
chris48s, line 17: If the /service/login route is completely gone now we should update this too
https://github.com/okfn/ckanext-unhcr/blob/d822476146071517b29fce0cd22258821a6230d9/ckanext/unhcr/mailer.py#L341

ckanext/unhcr/templates/user/snippets/login_form.html
```diff
@@ -0,0 +1,5 @@
+{% ckan_extends %}
+{% block login_button %}
+  {# ckanext-saml2auth adds a "SSO" button here that we want to hide #}
+  <button class="btn btn-primary" type="submit">{{ _('Login') }}</button>
```
chris48s, line 4: :+1: I put this comment here to remind myself to do this https://github.com/okfn/ckanext-unhcr/commit/a22b33a1e66d4251422d972a2430e0313a9bcfb0 We can revert it now

ckanext/unhcr/auth.py
```diff
@@ -59,9 +59,6 @@ def site_read(context, data_dict):
     if toolkit.request.path.startswith('/api'):
         # Let individual API actions deal with their auth
         return {'success': True}
-    elif toolkit.request.path == '/service/login':
-        # Allow local logins
-        return {'success': True}
```
chris48s, line 64: Just for my own understanding, why don't we need to replace this with /user/login ?

amercader: The auth exception for `user.login` when calling `site_read` is built-in in core: https://github.com/ckan/ckan/blob/dbe9fd8d225a621883fc30d53e4e8320091a320e/ckan/views/user.py#L83, no need to have it explicit here (/service/login was a custom route that needed special handling).

amercader:
@chris48s all done now

PR #538, opened 2021-04-22
CKAN 2.9: Drop collaborators plugin

chris48s:
Top-level summary of what is going on here: In CKAN 2.8 we were using ckanext-collaborators. In CKAN 2.9 part of that functionality is native in core so we're dropping the plugin.

This required a data migration, so also new in this PR is our first alembic migration. I pushed an extra commit to the 2.9 branch on the deploy repo to apply the alembic migrations on container boot https://github.com/okfn/docker-ckan-unhcr-aws/pull/35/commits/78fc1e9d8c10b4eaf16fbf2cfbf029ae76c97254


ckanext/unhcr/actions.py
```diff
@@ -189,7 +189,30 @@ def dataset_collaborator_create(up_func, context, data_dict):
...
+
+
+@toolkit.chained_action
+def package_collaborator_delete(up_func, context, data_dict):
+    up_func(context, data_dict)
+
+    mailer.mail_notification_to_collaborator(
+        data_dict['id'],
+        data_dict['user_id'],
+        capacity='collaborator',
```
chris48s, line 211: As noted on the call here, I've just hard-coded "collaborator" here to avoid re-implementing all the `package_collaborator_delete` param checking/validation.

We can rely on `data_dict['id']` and `data_dict['user_id']` because if `up_func()` ran without throwing an exception, they must exist and be valid

ckanext/unhcr/actions.py
```diff
@@ -1122,7 +1144,7 @@ def external_user_update_state(context, data_dict):
 
     toolkit.check_access('external_user_update_state', context, data_dict)
 
-    if state not in m.State.all:
+    if state not in [m.State.ACTIVE, m.State.DELETED]:
```
chris48s, line 1147: This was a built-in on the model in CKAN 2.8 
https://github.com/ckan/ckan/blob/ckan-2.8.7/ckan/model/core.py#L35
but it is gone in 2.9
https://github.com/ckan/ckan/blob/ckan-2.9.2/ckan/model/core.py

ckanext/unhcr/migration/unhcr/alembic.ini
```diff
@@ -0,0 +1,74 @@
+# A generic, single database configuration.
```
chris48s, line 1: I did try just deleting `ckanext/unhcr/migration/unhcr/alembic.ini` and `ckanext/unhcr/migration/unhcr/env.py` but CKAN fails to apply plugin migrations or generate new ones if they aren't there, even though the ones in CKAN core are the ones that are actually used so I've just left these here. I would like to try and understand more about the relationship between the migrations config in CKAN core and plugins at some point but at the moment I'm just accepting CKAN spits these files out and they need to be here.

ckanext/unhcr/migration/unhcr/versions/002_222e309d2220_migrate_dataset_member_to_package_member.py
```diff
@@ -0,0 +1,48 @@
+"""migrate dataset_member to package_member
```
chris48s, line 1: I decided to make the filename for our first migration `002_...` so that if we decide to consolidate/squash all our legacy migrations into an alembic migration at some point we can make that `001_...`

ckanext/unhcr/migration/unhcr/versions/002_222e309d2220_migrate_dataset_member_to_package_member.py
```diff
@@ -0,0 +1,48 @@
...
+        WHERE user_id IN (SELECT id FROM "user")
+        AND dataset_id IN (SELECT id FROM package)
+        AND modified IS NOT NULL;
+        """
+    ))
+    op.drop_table(u'dataset_member')
+
+
+def downgrade():
+    pass
```
chris48s, line 48: I could implement a `downgrade()` here that recreates the `dataset_member` table and copies the data back to it if we want, but I haven't bothered to spend the time on it. I think the chance of us wanting to do that is limited.

ckanext/unhcr/mailer.py
```diff
@@ -370,3 +372,34 @@ def compose_infected_file_email_body(recipient, resource_name, package_id, resou
...
+
+# Collaborators
+
+def _compose_collaborator_email_subj(dataset):
+    return u'{0} - Notification about collaborator role for {1}'.format(
+        toolkit.config.get('ckan.site_title'), dataset.title)
+
+def _compose_collaborator_email_body(user, dataset, role, event):
+    dataset_link = toolkit.url_for('dataset_read', id=dataset.id, qualified=True)
+    return render_jinja2('collaborators/emails/{0}_collaborator.html'.format(event), {
```
chris48s, line 385: Although we need to copy most of the email code from the old ckanext-collaborators plugin inline we we don't need to create the templates because we had already created custom versions of them in the past

ckanext/unhcr/tests/pt/test_mailer.py
```diff
@@ -451,3 +454,36 @@ def test_email_body(self):
...
+        dataset = factories.Dataset()
+        user = core_factories.User()
+        capacity = 'editor'
+
+        mailer.mail_notification_to_collaborator(
+            dataset['id'], user['id'], capacity, 'delete')
+
+        assert mock_mail_user.call_count == 1
+
+    # TODO: tests for body function
```
chris48s, line 484: I've just copy & pasted the tests from ckanext-collaborators here, but they're not amazing as they're just asserting we sent something. I'd like to improve this but don't want to get bogged down in it now. The emails code is as well tested as it was when we were using ckanext-collaborators so we haven't lost anything.

amercader (approved):
Looks good @chris48s :+1:

amercader:
FYI I copied your migration to another project and added the following extra check because I had forgotten to include the migration and users had started creating collaborators on 2.9:

```
        AND NOT (
            user_id IN (SELECT user_id FROM package_member) AND
            dataset_id IN (SELECT package_id FROM package_member)
        );
```
I don't think we need it in our case here

chris48s:
> I don't think we need it in our case here

Yes this would be useful if we had forgotten about it and deployed, but in our case, we will apply this migration before we ever boot a container running CKAN 2.9 so I don't think we need to worry about that situation.

Issue #531, opened 2021-04-20
CKAN 2.9: Fix data deposit controller tests


chris48s:
Now all/most of the routes are working it should be possible to update this test module from `webtest.TestApp` to `werkzeug.test.Client`

That should be the last large category of "tests that are failing because the tests are broken". Once that is done we can review the rest of the failures: there should be a (fairly large) handful of tests that are still failing, most of which indicate actual bugs that we need to fix. We can raise further issues from that

chris48s:
done in #539 - will raise new issues for further jobs

PR #539, opened 2021-04-23
CKAN 2.9: Fix data deposit controller tests

chris48s:
Refs #531

This gets almost all of the tests in this module passing except one which I think is failing because we haven't updated the activities code yet.
This gets us down to 23 total failures across the test suite, which is still lots to do (obviously there are some things that need fixing which don't correspond to a failing test) but that does feel like a manageable number of problems. I'll try and start driving more of a wegde into categories of problem we need to fix.


amercader (approved):
Great work @chris48s, getting there!

Issue #530, opened 2021-04-20
CKAN 2.9: Check/fix all `url_for()` and `link_for()` calls


chris48s:
There are a whole bunch of these calls using the `controller=` `action=` syntax which will need fixing

So far I've spotted

- On `/dashboard/organizations` the "Add Data Container" link is `/data-container/new?action=new&controller=organization` which is wrong
- [This test](https://github.com/okfn/ckanext-unhcr/blob/8fd48d7ad316993e8861e250175bc31149c01c83/ckanext/unhcr/tests/pt/test_mailer.py#L433) is failing by coincidence because `& != &amp;`

but there will be loads of these.

I think it makes sense to finish migrating all the pylons controllers to flask blueprints (issue #525 ) before we do this.

chris48s:
also `redirect_to()`

chris48s:
Related issue:
There are a bunch of placed where we are using legacy style routes which do work but are deprecated e.g:
`Route name "dataset_read" is deprecated and will be removed. Please update calls to use "dataset.read" instead`
While we are reviewing all these, it would make sense to look at this too.

chris48s:
The other thing I should do while I'm going through all these is make sure all the links to dataset pages use the dataset_type correctly.
i.e: `dataset_type ~ '.read'` rather than `'dataset.read'` so that links in the deposit don't link out to `/dataset`

If we can do this comprehensively, would this allow us to get rid of https://github.com/okfn/ckanext-unhcr/blob/19352420935ade9083f215bde0a287339255b2ee/ckanext/unhcr/plugin.py#L87-L102 ?
If so, that would be nice...

chris48s:
and `build_nav_icon()`

Issue #529, opened 2021-04-20
CKAN 2.9: Update Activity Streams code


chris48s:
CKAN 2.9 makes substantial changes to activities that we need to account for.
Some notes on this:

- From the CKAN 2.9 Changelog [Removals and deprecations](https://docs.ckan.org/en/2.9/changelog.html#removals-and-deprecations) notes - The following logic functions have been removed:
  - dashboard_activity_list_html
  - organization_activity_list_html
  - user_activity_list_html
  - package_activity_list_html
  - group_activity_list_html
  - organization_activity_list_html
  - recently_changed_packages_activity_list_html
  - dashboard_activity_list_html
  - activity_detail_list
- I commented out a bunch of stuff to do with activites on the CKAN 2.9 branch to get the container to boot in https://github.com/okfn/ckanext-unhcr/commit/4013543318d562b132eba9afe398f05ab340cec5
- https://github.com/okfn/ckanext-unhcr/blob/14d8564b6bac1782141f800dad09b30b2b954fdc/ckanext/unhcr/helpers.py#L736-L740

chris48s:
We can delete all the activity `.._html` actions and `_renderer` helpers. In CKAN 2.9, action rendering is handled by convention of template locations. For example the 'changed package' renderer will live in `templates/snippets/activities/changed_package.html` and we just include it.

Writing activities has not changed: `log_download_activity` and `create_curation_activity` should continue to work without changes. Only reading and displaying activities has changed.

We have two types of custom activities: Download activities and Curation activities

For download activities, we can replace `download_resource_renderer()` with a `download_resource.html` template.

Curation activities are a special-case of 'changed package' activity. `custom_activity_renderer` becomes a custom `templates/snippets/activities/changed_package.html` (it may still be useful to call a template helper here rather than implementing all the logic as jinja template)

Currently we do not save a full package dict when we save a curation activity: https://github.com/okfn/ckanext-unhcr/blob/19352420935ade9083f215bde0a287339255b2ee/ckanext/unhcr/helpers.py#L714 This creates several problems:
- CKAN 2.9 assumes the package JSON object will contain a 'title' key, but it doesn't. Refs https://github.com/ckan/ckan/issues/5865 TODO: read this fully
- CKAN 2.9 includes "view this version"/"changes" feature which allows you to diff the serialized packages against each other. This is going to produce unexpected results if each activity does not include a full package dict
[...]

PR #543, opened 2021-04-26
CKAN 2.9: Migrate package controller to flask

chris48s:
This is incomplete in the sense that there are a couple of old methods on the package controller to migrate, but I don't  think it does any harm to review/merge what I've done so far. I think what I've done so far is pretty solid.


ckanext/unhcr/blueprints/dataset.py
```diff
@@ -0,0 +1,255 @@
...
+            continue
+        if field['field_name'] in resource:
+            data[field['field_name']] = resource[field['field_name']]
+    data['name'] = '%s (copy)' % resource.get('name')
+
+    view = ResourceCreateView()
+    return view.get(package_type, dataset_id, data=data)
+
+
+def _call_publish_action(context, data_dict):
```
chris48s, line 129: I just did this to make it easier to mock out this whole interaction.
Because we're not calling `package_publish_microdata` we can't just directly mock `ckanext.unhcr.actions.package_publish_microdata`

ckanext/unhcr/tests/pt/controllers/test_dataset.py
```diff
@@ -11,7 +11,7 @@
 from ckanext.unhcr.tests import factories, mocks
 
 
-@pytest.mark.usefixtures('clean_db', 'unhcr_migrate')
+@pytest.mark.usefixtures('clean_db', 'unhcr_migrate', 'with_request_context')
 class TestExtendedPackageController(object):
```
chris48s, line 15: TODO: rename this

ckanext/unhcr/controllers/extended_package.py
```diff
@@ -13,96 +13,6 @@
...
-                continue
-            if field['field_name'] in resource:
-                data[field['field_name']] = resource[field['field_name']]
-        data['name'] = '%s (copy)' % resource.get('name')
-
-        return self.new_resource(id, data=data)
-
-    # Download
-
     def resource_download(self, id, resource_id, filename=None):
```
chris48s, line 16: TODO:

Instead of moving this one to `dataset.py`, move this function and the version for S3 downloads ( https://github.com/okfn/ckanext-unhcr/blob/master/ckanext/unhcr/blueprints/storage.py ) into a `resource.py` which defines 2 blueprints. Then do something like


```py
if plugins.plugin_loaded('s3filestore'):                                              
    bp.append(blueprints.unhcr_s3_resource_blueprint)                                 
else:                                                                                 
    bp.append(blueprints.unhcr_local_resource_blueprint)
```

in `plugins.py` to load the right one. Can they share the same checking logic?

ckanext/unhcr/blueprints/dataset.py
```diff
@@ -0,0 +1,255 @@
...
+    data['private'] = bool(dataset.get('private'))
+    if data.get('owner_org'):
+        data['owner_org'] = data['owner_org'] if data['owner_org'] in org_ids else None
+    data['original_dataset'] = dataset
+
+    view = DatasetCreateView()
+    return view.get(package_type, data=data)
+
+
+def resource_copy(package_type, dataset_id, resource_id):
```
chris48s, line 91: ..and then when you've done that, maybe this also moves to `resource.py` too?

ckanext/unhcr/blueprints/dataset.py
```diff
@@ -0,0 +1,255 @@
...
+    'unhcr_dataset',
+    __name__,
+    url_prefix=u'/dataset',
+    url_defaults={u'package_type': u'dataset'},
+)
+
+
+def publish(package_type, dataset_id):
+    if (not hasattr(toolkit.c, "user") or not toolkit.c.user):
+        return toolkit.abort(403, "Forbidden")
```
amercader, line 28: Seeing as we are using it in many methods perhaps is worth creating a small decorator? No big deal

ckanext/unhcr/blueprints/dataset.py
```diff
@@ -0,0 +1,255 @@
...
+
+    context = {'model': model, 'user': toolkit.c.user}
+    dataset = toolkit.get_action('package_show')(context.copy(), {'id': dataset_id})
+    if len(dataset['resources']) == 0:
+        return toolkit.abort(400, 'Dataset must have one or more resources to publish')
+    dataset = toolkit.get_action('package_patch')(context.copy(), {'id': dataset_id, 'state': 'active'})
+    toolkit.h.flash_success('Dataset "%s" has been published' % dataset['title'])
+    return toolkit.redirect_to('dataset_read', id=dataset['name'])
```
amercader, line 36: Flask routes use dots. Underscores should still work but let's future-proof it whenever we can.
```suggestion
    return toolkit.redirect_to('dataset.read', id=dataset['name'])
```

amercader (commented):
Looking good so far @chris48s

ckanext/unhcr/blueprints/dataset.py
chris48s: Yep. I've got a note to go through all these properly when I review all the `url_for`/`link_for`/`redirect_to` calls: https://github.com/okfn/ckanext-unhcr/issues/530#issuecomment-826727562

ckanext/unhcr/tests/pt/controllers/test_dataset.py
chris48s: d6ae2192

ckanext/unhcr/controllers/extended_package.py
chris48s: 14d689c

ckanext/unhcr/blueprints/dataset.py
chris48s: cc2391e

ckanext/unhcr/blueprints/dataset.py
chris48s: 197f196 and 246e2eb

ckanext/unhcr/controllers/extended_package.py
```diff
@@ -4,130 +4,11 @@
...
-
-        resp = super(ExtendedPackageController, self).resource_download(
-            id, resource_id, filename=filename
-        )
-        log_download_activity(context, resource_id)
-        return resp
-
     # Activity
 
     def activity(self, dataset_id):
```
chris48s, line 14: I'm going to leave this last function here as I think moving it to a blueprint is probably tied up in https://github.com/okfn/ckanext-unhcr/issues/529 We can migrate this last function and delete the pylons controller when we handle that issue

ckanext/unhcr/blueprints/resource.py
```diff
@@ -0,0 +1,97 @@
...
+from ckanext.s3filestore.views.resource import resource_download as base_s3_resource_download
+from ckanext.unhcr.activity import log_download_activity
+from ckanext.unhcr.utils import require_user, resource_is_blocked
+log = logging.getLogger(__name__)
+
+
+unhcr_resource_blueprint = Blueprint(
+    u'unhcr_resource',
+    __name__,
+    url_prefix=u'/dataset',
```
chris48s, line 20: I've only registered these routes for `/dataset/...` I'm not registering them explicitly for `/deposited-dataset/...` I haven't really been able to test this with deposits yet because I haven't fixed stuff enough that I can actually upload a resource to a deposit yet, but I will come back to the route definitions if I need to once I've got that working

ckanext/unhcr/blueprints/resource.py
```diff
@@ -0,0 +1,97 @@
...
+    except toolkit.NotAuthorized:
+        return toolkit.abort(403, toolkit._(u'Not Authorized to download the resource'))
+
+    if resource_is_blocked(context.copy(), resource_id):
+        return toolkit.abort(404, toolkit._(u'Resource not found'))
+
+    if plugins.plugin_loaded('s3filestore'):
+        resp = base_s3_resource_download(package_type, id, resource_id, filename)
+    else:
+        resp = base_local_resource_download(package_type, id, resource_id, filename)
```
chris48s, line 44: Detail on this is in the commit message for 14d689c but I'm pleased that I was able to both improve the behaviour and simplify the code at once here :)

amercader: That's one of the goals of 2.9!

ckanext/unhcr/blueprints/resource.py
amercader: Sounds good. In theory we should only need to register `/deposited-dataset` routes for custom ones that are not in core (eg `copy` and `download`), all the ones previously defined in `plugin.py` should be ow automatically handled by core.

amercader (approved):
Well done, this is a big chunk done @chris48s

PR #545, opened 2021-05-07
wrap each user interaction in a request_context under test

chris48s:
This gets some more tests passing, but I ended up going down a bit of a rabbit hole on this one.

The core issue I hit is basically if I have a test like this - contrived example:

```py
@pytest.mark.usefixtures('with_request_context')
def test_something(self, app):
    app.get(url, extra_environ={'REMOTE_USER': 'user1'}, status=200) # toolkit.c.user == 'user1'
    app.get(url, extra_environ={'REMOTE_USER': 'user2'}, status=200) # toolkit.c.user == 'user1'
```

`toolkit.c.user` is still `'user1'` in the second request even though I am explicitly passing `{'REMOTE_USER': 'user2'}`. This seems to be a behaviour change depending on if the URL we are requesting is served by pylons or flask which is why these tests were passing on CKAN 2.8. The solution (or workaround) seems to be to avoid the `with_request_context` decorator and then wrap each set of interactions for a given user in its own `test_request_context` if we are testing the behaviour for more than one user in one test. So rewriting this like:

```py
# no with_request_context decorator
def test_something(self, app):
    with app.flask_app.test_request_context():
        app.get(url, extra_environ={'REMOTE_USER': 'user1'}, status=200) # toolkit.c.user == 'user1'
    with app.flask_app.test_request_context():
        app.get(url, extra_environ={'REMOTE_USER': 'user2'}, status=200) # toolkit.c.user == 'user2'
```

`toolkit.c.user` is now `'user2'` in the second request.

I'm not really sure if this is a bug or a feature, but it was quite non-obvious and took a while to  figure out this was what I needed to do to get these working.

67e54c3 is a pretty straightforward swap.

c590e49 is a bit more messy. There were some other basic things that needed fixing like replacing

`toolkit.url_for(controller='package', action='resource_download', id=dataset['id'], resource_id=resource['id'])` with
`toolkit.url_for('resource.download', id=dataset['id'], resource_id=resource['id'])`

but also while I was picking this apart I realised that there was some odd naming, some situations that weren't tested and some scope to reduce duplication a bit so I basically ended up rewriting these tests. I usually try to avoid functional changes and refactoring in the same commit but that's just the way it panned out this time. The diff isn't great to read but I think the result makes more sense.

### DIFF
--- a/ckanext/unhcr/tests/pt/controllers/test_dataset.py
+++ b/ckanext/unhcr/tests/pt/controllers/test_dataset.py
@@ -323,7 +323,7 @@ class TestDatasetViews(object):
         )
 
 
-@pytest.mark.usefixtures('clean_db', 'unhcr_migrate', 'with_request_context')
+@pytest.mark.usefixtures('clean_db', 'unhcr_migrate')
 class TestPrivateResources(object):
 
     def setup(self):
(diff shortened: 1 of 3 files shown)


ckanext/unhcr/tests/pt/controllers/test_dataset.py
```diff
@@ -354,144 +354,142 @@ def test_private_can_be_true_if_sysadmin(self):
...
-        url = toolkit.url_for(
-            controller='package',
-            action='resource_download',
-            id=dataset['id'],
-            resource_id=resource['id'])
-
-        # We don't pass authorization (forbidden)
-        environ = {'REMOTE_USER': self.normal_user['name'].encode('ascii')}
-        res = app.get(url, extra_environ=environ, status=403)
+        dataset_read_url = toolkit.url_for('dataset.read', id=dataset['id'])
```
amercader, line 373: I suspect the issue is related to https://github.com/ckan/ckan/issues/5748 which I believe we discussed at some point. If you want to avoid using different `test_request_context`s inside the same test you can build the urls using the dataset `name` instead of the `id`. That should avoid the internal redirect which is causing the issues with the authorized user.

amercader (approved):
@chris48s see comment about `test_request_context`. Otherwise it looks good.

PR #547, opened 2021-05-10
data deposit fixes

### LINKED ISSUE #540: CKAN 2.9: Can't add resources to deposited datasets
Adding resources to vanilla datasets works ok
For deposited datasets we get `UndefinedError: 'pkg_dict' is undefined`

### LINKED ISSUE #544: CKAN 2.9: Review dataset add_url_rule statements for deposits
- Once you've got deposited dataset uploads working test dataset/resource routes on both `/dataset/...` and `/deposited-dataset/...`
- Built-in routes (and overridden built-in routes) should automatically work on both `/dataset` and `/deposited-dataset`
- Any fully custom routes need to be manually registered for `/deposited-dataset`. There will be some missing.
- Are there any routes that don't make sense for deposited datasets?

chris48s:
slightly mixed bag of fixes
closes #540
closes #544


ckanext/unhcr/templates/package/edit_base.html
```diff
@@ -1,13 +1,10 @@
...
 
-{% if pkg and dataset_type == 'deposited-dataset' %}
-  {% set curation = h.get_deposited_dataset_user_curation_status(pkg, c.userobj.id) %}
-{% endif %}
-
 {% block secondary %}
 <aside class="secondary col-sm-3">
   {% block secondary_content %}
     {% if pkg and dataset_type == 'deposited-dataset' %}
+      {% set curation = h.get_deposited_dataset_user_curation_status(pkg, c.userobj.id) %}
```
chris48s, line 7: This is a bit of an odd one

I don't think it is a _scope_ issue as such, but in the change from Jinja 2.8 to 2.10 it seems like there is a change in what order stuff gets processed in certain inheritance scenarios (but not all of them) so sometimes it seems like we're processing the blocks before we set this variable.

I've decided not to get too bogged down in this - moving this `set` into the blocks where we need it seems to do the job for all the templates.

PR #550, opened 2021-05-14
CKAN 2.9: Background job fixes

### LINKED ISSUE #528: CKAN 2.9: Uploading resource throws `NotAuthorized` in background job
To reproduce, upload a resource.

Stacktrace:

```
ERROR [ckan.lib.jobs] Job a7a9be05-7fe2-4ece-b05a-adc14fd0781f on worker rq:worker:94fd3134b60144e6abf817538fbce35e raised an exception: Action package_show requires an authenticated user
  File "/usr/lib/python2.7/site-packages/rq/worker.py", line 812, in perform_job
    rv = job.perform()
  File "/usr/lib/python2.7/site-packages/rq/job.py", line 588, in perform
    self._result = self._execute()
  File "/usr/lib/python2.7/site-packages/rq/job.py", line 594, in _execute
    return self.func(*self.args, **self.kwargs)
  File "/srv/app/src_extensions/ckanext-unhcr/ckanext/unhcr/jobs.py", line 42, in process_dataset_on_update
    _process_dataset_fields(package_id)
  File "/srv/app/src_extensions/ckanext-unhcr/ckanext/unhcr/jobs.py", line 62, in _process_dataset_fields
    package = package_show({'job': True}, {'id': package_id})
  File "/srv/app/src_extensions/ckan/ckan/logic/__init__.py", line 477, in wrapped
  File "/srv/app/src_extensions/ckan/ckan/logic/action/get.py", line 978, in package_show
    _check_access('package_show', context, data_dict)
  File "/srv/app/src_extensions/ckan/ckan/logic/__init__.py", line 313, in check_access
    raise NotAuthorized(msg)
```

chris48s:
Closes #528

There's really three things going on in this PR:

One is fairly simple - https://github.com/okfn/ckanext-unhcr/issues/528
In CKAN 2.8 it seems like we could just call actions from a background job with no user in the context object and they would authenticate.. somehow. In CKAN 2.9 that doesn't work so we need to pass some kind of auth info in the context.

While I was digging into that I found there were some situations where background jobs should have been registering but weren't. These problems are not CKAN 2.9-specific. These bugs exist in 2.8 and we could validly backport the whole of fba634d to the 2.8 branch. I've also improved the testing around this

I'm fairly happy with the solution to those two problems - although having found some more bugs in the duck-typing to work out if `data_dict` is a package or a resource I'm think it is worth considering again if it would be more sensible to just split out the `IPackageController` hooks and `IResourceController` hooks into two separate plugins as I think that might make this both simpler and less error-prone?

~The third category of issue is that our method for 2-way linking of datasets relies on the revisions feature in CKAN <=2.8 and this has been completely removed in CKAN 2.9 so we need to do something else. This is where all the worms come out of the can.. I've pushed a WIP solution to this which sort of looks like it works, but it doesn't really. There are too many cases it doesn't handle and it introduces a bunch of new exciting race conditions. I think having got this far into the problem it might be useful to catch up sometime next week on other possible ways of tackling this problem. I'll continue to chew it over..~


chris48s:
TODO: rebase on to ckan2.9 branch before next CI run

ckanext/unhcr/jobs.py
```diff
@@ -125,51 +126,40 @@ def _modify_weighted_field(package, key, weights):
...
-    # Next revision
-    next = []
-    if len(revisions) >= 1:
-        revision = revisions[-1]
-        if revision.state == 'active':
-            next = utils.normalize_list(revision.value)
-
-    return {'prev': prev, 'next': next}
+    # TODO: fix backlinks https://github.com/okfn/ckanext-unhcr/issues/577
+    return {'prev': [], 'next': []}
```
chris48s, line 165: This implementation doesn't will just mean we don't create or remove anything. I've written up an issue at https://github.com/okfn/ckanext-unhcr/issues/577 based on our discussion earlier in the week

amercader (approved):

PR #552, opened 2021-05-17
CKAN 2.9: remove a bunch of global state vars from templates

chris48s:
not exhaustive but every little helps


amercader (approved):
Nice one

Issue #536, opened 2021-04-20
CKAN 2.9: Review template overrides


chris48s:
There are many upstream templates which we are overriding. In some cases there may be upstream template changes that mean we need to update our overrides

chris48s:
Most of this is now done in #552 #553 and #554

The last thing I need to deal with is backporting the changes from https://github.com/ckan/ckan/commit/de7ba5dde39f1df46c95d0fa87f0cfb5163550bd onto our custom version of `templates/organization/member_new.html` but this is slightly more involved so I will do it as another separate PR.

Then I think once that is done I will handle the rest of the template updates under https://github.com/okfn/ckanext-unhcr/issues/542 as that is the remaining category of thing that needs doing.

chris48s:
#556 wraps this one up

chris48s:
remaining template work is will be handled under #530

PR #553, opened 2021-05-17
CKAN 2.9: Update activity streams

### LINKED ISSUE #525: CKAN 2.9: Migrate remaining pylons controllers to flask blueprints
https://github.com/okfn/ckanext-unhcr/tree/ckan2.9/ckanext/unhcr/controllers

- [ ] package
- [x] user - https://github.com/okfn/ckanext-unhcr/pull/541

### LINKED ISSUE #529: CKAN 2.9: Update Activity Streams code

chris48s:
Closes #529
Closes #525
Refs #536

This PR updates us to use the new activities features in CKAN 2.9.


chris48s:
TODO: rebase on to ckan2.9 branch before next CI run

ckanext/unhcr/actions.py
```diff
@@ -403,202 +402,96 @@ def container_request_list(context, data_dict):
...
+        offset,
+        include_hidden_activity,
+        include_activity_types,
+    )
 
-@toolkit.side_effect_free
-def user_activity_list(context, data_dict):
-    full_list = get_core.user_activity_list(context, data_dict)
+    return model_dictize.activity_list_dictize(
+        activity_objects, context, include_data=True)
```
chris48s, line 457: `package_internal_activity_list` always passes `include_data=True`
The user must have `package_update` to see the curation activites (which contain a stripped down object) and they must be a container admin to see the resource download activites (which include a full resource object)

ckanext/unhcr/actions.py
```diff
@@ -403,202 +402,96 @@ def container_request_list(context, data_dict):
...
-    normal_activities = [
-        a for a in full_list if 'curation_activity' not in a.get('data', {})]
-    return normal_activities
+@toolkit.chained_action
+def dashboard_activity_list(up_func, context, data_dict):
+    return _remove_internal_activities(up_func(context.copy(), data_dict))
 
 
 @toolkit.side_effect_free
-def recently_changed_packages_activity_list(context, data_dict):
```
chris48s, line 499: I binned `recently_changed_packages_activity_list` because it doesn't return any activities we care about filtering https://github.com/ckan/ckan/blob/ckan-2.9.3/ckan/model/activity.py#L435

ckanext/unhcr/actions.py
```diff
@@ -403,202 +402,96 @@ def container_request_list(context, data_dict):
...
+    )
 
-@toolkit.side_effect_free
-def user_activity_list(context, data_dict):
-    full_list = get_core.user_activity_list(context, data_dict)
+    return model_dictize.activity_list_dictize(
+        activity_objects, context, include_data=True)
+
+
+def _remove_internal_activities(activities):
```
chris48s, line 460: I think it makes sense to always exclude 'changed curation state' and 'download resource' from any other feeds than the 'internal activity' feed on the dataset page

ckanext/unhcr/activity.py
```diff
@@ -29,3 +30,28 @@ def log_download_activity(context, resource_id):
...
+    """
+    activity_context = {'ignore_auth': True}
+    data_dict = {
+        'user_id': user_id,
+        'object_id': dataset_id,
+        'activity_type': 'changed curation state',
+        'data': {
+            'curation_activity': activity_type,
+            'package': {'name': dataset_name, 'id': dataset_id},
+        }
```
chris48s, line 49: We discussed this on a call the other day, but I decided that we should make the curation activities into a special 'changed curation state' activity instead of a special case of 'changed package' to make the filtering easier.

I also decided that we don't need to bother storing a full package object on them. We're not going to allow the user to browse old versions or view a diff so we can just leave the data object as-is here.
I think this is the simplest solution.

ckanext/unhcr/plugin.py
```diff
@@ -149,17 +141,6 @@ def update_config_schema(self, schema):
...
-
-    def before_map(self, _map):
-
-        # package
-        controller = 'ckanext.unhcr.controllers.extended_package:ExtendedPackageController'
-        _map.connect('dataset_internal_activity', '/dataset/internal_activity/{dataset_id}', controller=controller, action='activity')
-        _map.connect('deposited-dataset_internal_activity', '/deposited-dataset/internal_activity/{dataset_id}', controller=controller, action='activity')
-
-        return _map
-
```
chris48s, line 162: This PR also completes the controller migration

ckanext/unhcr/migration/unhcr/versions/003_058f761ed45f_update_curation_activities.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+# revision identifiers, used by Alembic.
+revision = '058f761ed45f'
+down_revision = '222e309d2220'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
```
chris48s, line 19: When it comes to applying https://github.com/ckan/ckan/blob/master/ckan/migration/migrate_package_activity.py I think the sensible thing to do here is to run this migration first and then run `migrate_package_activity.py` later. Once we've applied this, all the remaining 'changed package' activities will just be an actual 'changed package' - we'll have turned all our custom activities into something else so we should be able to just run it with no changes.

There's a bit of follow up here which is that I want to test running `migrate_package_activity.py` on dump of a real CKAN 2.8 DB that has been upgrade to test this. I haven't done it yet.

ckanext/unhcr/activity.py
amercader: Sounds good

ckanext/unhcr/fanstatic/theme.css
```diff
@@ -2279,6 +2279,10 @@ i {
   color: #000;
 }
 
+.activity .item.changed-curation-state .icon {
+  background-color: #8c76ce;
```
amercader, line 2283: I took me a while to notice this (I think core activities can have different background colors as well). Perhaps something more obvious with a lock icon somewhere or a different background color for the whole div would be better.
Not for this PR of course

amercader (approved):
Great work @chris48s, code looks good and I tested it locally and all seems to work as expected!

PR #554, opened 2021-05-17
CKAN 2.9: mixed bag of template updates

chris48s:
Refs #536

This takes a bunch of minor changes to formatting, terminology and template logic which have changed upstream in core between CKAN 2.8 and CKAN 2.9 and applies them to our custom templates.


chris48s:
There is one category of template change I haven't bothered including here which is the changes to `ckanext/unhcr/templates/user/edit_user_form.html`  and `ckanext/unhcr/templates/user/new_user_form.html` allowing avatar uploads from https://github.com/ckan/ckan/commit/6a1a220910ce737d9be9f108712c79f08285bfed / https://github.com/ckan/ckan/pull/5426

The forms seem to work fine without and I suspect we probably don't really want to enable this (especially as this form is only usable by users creating an external/partner account). Any downside to just leaving these out?

amercader (approved):

amercader:
@chris48s yes, it's fine to leave the user image fields out of the form

PR #560, opened 2021-05-20
remove redundant core patch

chris48s:
this is fixed in CKAN 2.8.8


chris48s:
TOD: pick to 2.9 branch

PR #570, opened 2021-05-26
Scripts to py3

avdata99:
Update and run scripts with Python3

## Notes:
`west-and-central-africa` is failing for the `update_container_visibility` script at the `update_container_visibility_aug2020.csv` file

image


README.md
```diff
@@ -10,21 +10,23 @@ CKAN extension for the UNHCR RIDL project
...
+  - [Building static assets](#building-static-assets)
+  - [Working with i18n](#working-with-i18n)
+  - [Loging into container](#loging-into-container)
+  - [Updating readme](#updating-readme)
+  - [Managing docker](#managing-docker)
+  - [Prepare a local environment for running scripts](#prepare-a-local-environment-for-running-scripts)
+  - [Generate deposited-dataset schema](#generate-deposited-dataset-schema)
+  - [Create data containers and data deposit](#create-data-containers-and-data-deposit)
+  - [Create development users](#create-development-users)
+  - [Testing email notifications](#testing-email-notifications)
```
chris48s, line 29: The README table of contents is generated
Did you do this by hand or with `npx doctoc README.md`?
If I add a section and run `npx doctoc README.md` I don't get an indented list like this.

test.ini
```diff
@@ -9,7 +9,7 @@ host = 0.0.0.0
 port = 5000
 
 [app:main]
-use = config:../../src/ckan/test-core.ini
+use = config:/srv/app/src/ckan/test-core.ini
```
chris48s: This hasn't really bothered me because I never `git commit -a` but I know some peole like that. I think if you want to clean up the working copy you can probably do something like
https://github.com/ckan/ckanext-harvest/blob/e2e8acbb6d87d44663e92c5ff5cbd67401588c2d/.github/workflows/test.yml#L55-L56 in the CI build?
Maybe do that in another PR if you want.
Either way, when you update this one can you squash 7c89de4 into efe1c62 to clean up the history.

chris48s (commented):
> `west-and-central-africa` is failing for the `update_container_visibility` script at the 

We double up these scripts for setting up the dev environment with realistic data but we've also used them to perform tasks on the live env at various points, so
- `initial_data_container_list_feb_2018.csv` (which doesn't contain `west-and-central-africa`) was used to create the initial set of containers (orgs) in production when the site first launched
- then the platform got used for a few years
- then `update_container_visibility_aug2020.csv` was used to set visibility when we launched the External Data Deposit feaure (we'll get into this at some point) byt which point `west-and-central-africa` existed and needed this setting on it.

In terms of using the scripts to set up a dev environment with some realistic-ish test data in it, it makes no difference really :)

README.md
avdata99: I didn't change this TOC manually @chris48s. I didn't run the npx command either

README.md
```diff
@@ -156,6 +158,16 @@ $ docker system prune -a --volumes # CAUTION: it will purge all docker projects
...
 ```
 
+## Prepare a local environment for running scripts
+
+Create a local Python 3 environment, activate it and install requirements.  
+
+```
+python3 -m venv /path/to/your/new/env
+source /path/to/your/new/env/bin/activate
+pip install -r scripts/requirements.txt
```
chris48s, line 168: This file isn't in the PR. Can you add it?

test.ini
avdata99: I just squash it

README.md
avdata99: The only np* command I remember running is the `npm install` from the README.
I just undo this change @chris48s

chris48s: I feel like we're both failing to communicate here :( but I rebuilt the TOC with doctoc, squashed and merged..

PR #572, opened 2021-05-26
Hotfix empty users list

### LINKED ISSUE #521: Requesting password reset for unknown email throws `IndexError: list index out of range`
Need to handle the case here
https://github.com/okfn/ckanext-unhcr/blob/d822476146071517b29fce0cd22258821a6230d9/ckanext/unhcr/actions.py#L1266
where `user_list` returns 0 users

avdata99:
Fixes #521 for CKAN 2.9


ckanext/unhcr/actions.py
```diff
@@ -1284,6 +1284,9 @@ def user_autocomplete(context, data_dict):
 @toolkit.chained_action
 def user_list(up_func, context, data_dict):
     users = up_func(context, data_dict)
+    if len(users) == 0:
+        return []
```
chris48s, line 1288: I tried running this locally and it causes `/user` to throw `TypeError: object of type 'Query' has no len()` for me.

I think we need to deal with the `return_query` case:
https://github.com/ckan/ckan/blob/ckan-2.9.3/ckan/logic/action/get.py#L883-L885

Given this is a bit fiddly can we add some test cases for our `user_list` in https://github.com/okfn/ckanext-unhcr/blob/master/ckanext/unhcr/tests/pt/actions/test_user_actions.py

Ta.

avdata99: Multiple test cases were added @chris48s 
I think the PR is ready to go.

ckanext/unhcr/tests/pt/actions/test_user_actions.py
```diff
@@ -60,6 +61,43 @@ def test_user_list(self):
...
+
+    def test_user_list_query(self):
+        sysadmin = core_factories.Sysadmin()
+        external_user = factories.ExternalUser()
+        internal_user = core_factories.User()
+        default_user = toolkit.get_action('get_site_user')({ 'ignore_auth': True })
+
+        action = toolkit.get_action('user_list')
+        context = {'user': sysadmin['name'], 'return_query': True}
+        # add a filter to get 0 results
```
chris48s, line 85: I don't understand this comment - is it a copy & paste error?

avdata99: Yes, it is. I just fix and rebase.

Issue #511, opened 2021-04-07
Use toolkit.get_action in data_container views


chris48s:
https://github.com/okfn/ckanext-unhcr/blob/14d8564b6bac1782141f800dad09b30b2b954fdc/ckanext/unhcr/blueprints/data_container.py#L8-L10

There are 3 reasons why we might be doing this:

1. We explicitly want to call the core action (and not a version that has been provided by a plugin higher up the stack)
2. To bypass the auth (if it is this, we could use `ignore_auth: True`)
3. There was some problem with chained action/auth

Does it look like any of those were an issue? If not, convert these to use `toolkit.get_action()`

chris48s:
done

Issue #558, opened 2021-05-18
fully adopt chained actions


chris48s:
We've converted quite a few of them, but there are still some actions left which are directly importing from `ckan.logic.action`. Where possible, the rest of them should be moved to chained actions.

PR #574, opened 2021-05-27
Use toolkit actions

avdata99:
Related to #511. 
Probably a final fix


ckanext/unhcr/auth.py
```diff
@@ -166,7 +163,7 @@ def package_create(context, data_dict):
 
     # Data container
     context['model'] = context.get('model') or model
-    return auth_create_core.package_create(context, data_dict)
+    return toolkit.get_action('package_create')(context, data_dict)
```
chris48s, line 166: None of these changes to the auth file will work because you're replacing calls to auth functions with calls to actions.
Tbh, I'm not 100% sure what the best practice is for auth (/cc @amercader ?) but if you want to replace the calls to core auth functions with `toolkit` calls, one option is to use `@chained_auth_function` / `next_auth` if we need to call the parent.

Another is to do something like

https://github.com/okfn/ckanext-unhcr/blob/b1152c47119202ee140108e5e43de9cbaa3f1181/ckanext/unhcr/auth.py#L194-L199

if we need to call another auth function. Remember in CKAN the auth functions themselves return `{'success': False}` if the action is not allowed, but `toolkit.check_access()` raises a `NotAuthorized` so we can't `return toolkit.check_access(...)` in auth functions.

chris48s (changes requested):
I know the CKAN 2.9 branch is a bit of a pain to work on right now as there are still some outstanding PRs for the failing tests, but this PR introduces a lot of new build failures.
We will get the build for this branch green next week to make this more obvious and easier to work on.


I think if you also want to take on https://github.com/okfn/ckanext-unhcr/issues/558 as well as https://github.com/okfn/ckanext-unhcr/issues/511 that's fine, but I don't think we should do it all in one PR. Lets try and address these changes in smaller chunks. A PR that changes so much in one go is really hard and time-consuming to manually test/review. There's so much change going on here to validate.

ckanext/unhcr/auth.py
avdata99: You are right @chris48s. I disn't realize those are auth and not action calls.
I fixed it [here](https://github.com/okfn/ckanext-unhcr/pull/574/commits/66d72103b0b3526327b22eef75247d5acedf201b)

avdata99: I'm seeing [some heavy conflicts](https://github.com/okfn/ckanext-unhcr/pull/574/conflicts) with [your work](https://github.com/okfn/ckanext-unhcr/commit/37454dec702c3f25b966b5ef40104e3344102882) @chris48s  
My plan is to preserve your changes and validate the chained functions still works.
Is this the right approach?

chris48s: Yes. One of the reasons I didn't pick https://github.com/okfn/ckanext-unhcr/issues/558 as an issue to work on is because I knew there would be big conflicts with https://github.com/okfn/ckanext-unhcr/pull/553 . That PR removes quite a lot of actions we don't need any more in CKAN 2.9 and also covers a lot of the conversions to chained actions.

Your approach sounds right. Accept those changes and see what is left. After that PR I think only
- `package_update`
- `organization_create`
- `organization_member_create`
- `organization_member_delete` and
- `organization_list_all_fields`

will need any changes in the actions file.

ckanext/unhcr/auth.py
```diff
@@ -111,7 +111,7 @@ def organization_create(context, data_dict):
                 return {'success': True}
 
     # Base access check
-    result = toolkit.get_action('organization_create')(context, data_dict)
+    result = toolkit.check_access('organization_create')(context, data_dict)
```
chris48s, line 114: If you're going to call `check_access` in an auth function I think the auth function should still return a value if `check_access` throws. See the example in https://github.com/okfn/ckanext-unhcr/pull/574#discussion_r641670157

ckanext/unhcr/auth.py
avdata99: The rebase is done. Also updated tests @chris48s

ckanext/unhcr/auth.py
```diff
@@ -113,7 +114,11 @@ def organization_create(context, data_dict):
                 return {'success': True}
 
     # Base access check
-    result = auth_create_core.organization_create(context, data_dict)
+    try: 
+        result = next_auth(context, data_dict)
```
chris48s, line 118: If we're just calling `next_auth()` in a chained function it returns a dict.
It is only if we call `toolkit.check_access()` that `{'success': False}` is wrapped as an exception instead of returning: https://github.com/ckan/ckan/blob/ckan-2.9.3/ckan/logic/__init__.py#L311-L313
(I know this is a bit confusing  )

ckanext/unhcr/tests/pt/test_schema.py
```diff
@@ -95,7 +95,7 @@ def test_both_types_data_fields_missing(self):
         with pytest.raises(toolkit.ValidationError) as e:
             call_action('package_update', {'user': self.sysadmin['name']}, **dataset)
 
-        errors = e.value.error_dict['resources'][1]
+        errors = e.value.error_dict['resources'][0]
```
chris48s, line 98: Unfortunately the fix for this is not so simple - the test is not failing because the assertion is wrong. The test is correct, but CKAN 2.9 introduces a bug in the validation which needs to be fixed upstream.
@amercader wants to pick it up so I submitted a failing test case at https://github.com/ckan/ckan/pull/6119 and it is with @amercader to work on. My proposed solution is that we skip the test for now to get the build green (PR https://github.com/okfn/ckanext-unhcr/pull/587 ) and then re-enable it once we've got a serviceable core patch we can apply. Can we back this commit out please.

avdata99: Changed [here](https://github.com/okfn/ckanext-unhcr/pull/574/commits/02108ffaf101a6eef8ce486fb6f79c74c2320597)
I rebased this branch and now the test is green

ckanext/unhcr/auth.py
avdata99: I just [make a change](https://github.com/okfn/ckanext-unhcr/pull/574/commits/ffae220236938b5be56a8f29af479d84cb963d63) that fixed this error

chris48s: This is another situation where we are clearly struggling to communicate. Sorry if the previous explanation is unclear.

Calling `next_auth` returns a dict if auth fails (just like this function does) because we're invoking the auth function directly. There is no exception to catch.

The block should be:

```py
    result = next_auth(context, data_dict)
    if not result['success']:
        return result
```

The case where we need to catch `toolkit.NotAuthorized` would be if we were calling `toolkit.check_access()` but That isn't what we are doing here.

avdata99: My bad @chris48s . 
Fixed [here](https://github.com/okfn/ckanext-unhcr/pull/574/commits/97f490ca7f1a4b49e916735d4a6c838352b66579). 
I misunderstood your message.

ckanext/unhcr/tests/pt/test_schema.py
```diff
@@ -80,6 +80,13 @@ def test_both_types_data_fields_missing(self):
             'description': 'Some description',
             'type': 'attachment',
             'file_type': 'report',
+            # TODO add 'visibility' after merge #583 
+            # 'visibility': 'public',
+            'date_range_start': '2017-01-01',
+            'date_range_end': '2017-01-03',
+            'version': '1',
+            'identifiability': 'personally_identifiable',
+            'process_status': 'anonymized',
```
chris48s, line 89: Adding these fields to this resource where the resource is `'type': 'attachment'` doesn't really make sense. They are not part of the schema for attachment resources:

Screenshot at 2021-06-15 14-15-48


Aside from that, this PR is good to go and it needed rebasing on to master to be mergeable anyway. So what I've done is I backed out this change, squashed this all into one commit: 03f0275 and I've pushed that to master. I will mark the PR "closed" but it is really merged.

PR #579, opened 2021-05-28
Fix copy tags

### LINKED ISSUE #549: copying a dataset with tags does not copy the tags to the new dataset
- create a dataset with one or more tags
- press copy
- all the fields are populated, except tags

avdata99:
Fixes #549 

Created a dataset with 3 tags and press copy.
image


amercader (approved):
Looks good @avdata99

Issue #581, opened 2021-05-31
Resource-level visibility


amercader:
### Context

In a standard CKAN install, a dataset visibility is defined by the `private` metadata field (a boolean). A **private** dataset can only be accessed by members of its organization, collaborators or sysadmins. This includes seeing the metadata and downloading the data (for members) and editing the metadata (for editors and admins). A **public** dataset can be accessed and its resources downloaded by any user.

RIDL had a requirement to introduce a third state: metadata public but private data. That means that all users should be able to see all datasets metadata (ie browse and access dataset pages), but in some cases *downloading* the files should be only available to data container members or collaborators. In RIDL this are called "Private" datasets, and when visiting the dataset page as a user without permissions to download you get a message (and the download links are disabled):

![Screenshot_2021-05-31 United Republic of Tanzania Livelihoods Programme Monitoring Beneficiary Survey - 2020 - RIDL](https://user-images.githubusercontent.com/200230/120177518-d12f5880-c208-11eb-9c3e-201f7eefd2dc.png)

As the builtin `private` field didn't allow a new state, visibility is handled by a custom `visibility` field, defined [here](https://github.com/okfn/ckanext-unhcr/blob/master/ckanext/unhcr/schemas/dataset.json#L61-L78). The builtin [`private`](https://github.com/okfn/ckanext-unhcr/blob/master/ckanext/unhcr/schemas/dataset.json#L54-L60) field is always set to `False` (via the `always_false_if_not_sysadmin` validator).

Note that while there are three theoretically possible cases, only two are available via the frontend:

|  Internal Access Level | `visibility` value | `private` value | notes |
|---|---|---|---|
| Internally visible | `public` | `True` | Metadata visible / Downloads public |
| Private | `restricted` | `True` | Metadata visible / Downloads restricted |
| *{Not show}* | `private` | `False` | Metadata restricted / Downloads restricted (*Not available to normal users*) |

The third one was added in case true private datasets were ever needed but to all intents and purposes, all RIDL datasets have public metadata and some of them restricted downloads.

### Current implementation

[...]

Issue #509, opened 2021-03-26
CKAN 2.9: Adopt package_revise


chris48s:
https://docs.ckan.org/en/2.9/api/index.html#ckan.logic.action.update.package_update
https://github.com/ckan/ckan/pull/4618

First place to look at this is in the background jobs as we have discovered another exciting race condition there (hopefully this can help us there) but there may be other places where we are relying on `package_patch` / `package_update` operations which can be improved by switching to a concurrency-safe operation.

PR #583, opened 2021-06-01
Resources visibility

avdata99:
Related to #581

**Work in progress**

 - Copy field definition from dateset to resource (required only at resource level)
 - Remove `visibility` field from dataset form
 - Add `visibility` field to resources form
 - Adds a migration script to copy `visibility` values from datasets to resources
 - Include [ckanextt-ddi PR](https://github.com/okfn/ckanext-ddi/pull/8) to import and take care of visibility for new resources

### Screens

Allow define visibility for resources

image

Show visibility status at resources list

image

Show _request access_ at for all resources

image

Show if request was already requested

image


ckanext/unhcr/jobs.py
```diff
@@ -85,7 +85,11 @@ def _modify_package(package):
     # visibility
     # TODO: clarify what level of anonymization required
     if package['identifiability'] == 'personally_identifiable':
-        package['visibility'] = 'restricted'
+        for resource in package['resources']:
+            if resource['visibility'] != 'restricted':
+                # require a resource_update call
+                resource['visibility'] = 'restricted'
+                toolkit.get_action('resource_update')({'job': True}, resource)
```
chris48s, line 92: At the moment, we've got several known race condition bugs caused by perfoming writes in background jobs so I'm really keen to not make that problem worse than we have to.

We are currently pinning our hopes on moving from `package_update` to the new `package_revise`. See
- https://docs.ckan.org/en/2.9/api/index.html#ckan.logic.action.update.package_revise
- https://github.com/ckan/ckan/pull/4618

which should solve this problem (one of the many things on my list - https://github.com/okfn/ckanext-unhcr/issues/509 ) but there is no `resource_revise`.

I have not tested, but can we accomplish the same thing here by modifying the resource dicts in place in `package['resources']` so these changes get applied when we pass `package` back and call `package_update` in `_process_dataset_fields()`? If so that would help.

ckanext/unhcr/schemas/dataset.json
```diff
@@ -778,6 +778,24 @@
...
+              {
+                "value": "restricted",
+                "label": "Private"
+              },
+              {
+                "value": "public",
+                "label": "Internally Visible"
+              }
+            ]
+          },
```
chris48s, line 798: I think as well as adding "Internal Access Level" to the resource schema we also want to remove it from the dataset schema, don't we?

ckanext/unhcr/templates/package/resource_read.html
```diff
@@ -2,6 +2,9 @@
 
 {% block bodytag %} {{ super() }} class="resource-read"{% endblock %}
 
+{% set access_already_requested = h.get_existing_access_request(c.userobj.id, res.id, 'requested') %}
```
chris48s, line 5: The access request is still going to be a request to access the whole dataset/package so this will want to be `res.package_id` not `res.id`. Same comment applies in some other places where we are doing the same thing

ckanext/unhcr/templates/package/resource_read.html
```diff
@@ -57,6 +69,6 @@
 {% block footer %}
   {{ super() }}
   {% if not h.can_download(pkg) %}
-    {% snippet "package/snippets/collaboration_modals.html", pkg=pkg %}
+    {% snippet "package/snippets/collaboration_modals.html", res=res %}
```
chris48s, line 72: Similarly, the access request is still going to be a request to access the whole dataset/package. By the looks of it the only thing from `pkg` we use in `collaboration_modals.html` is the ID so maybe it makes sense to refactor that so that we pass only the ID then we can call
`{% snippet "package/snippets/collaboration_modals.html", dataset_id=pkg.id %}`
or
`{% snippet "package/snippets/collaboration_modals.html", dataset_id=res.package_id %}`
depending on the context.

ckanext/unhcr/jobs.py
avdata99: If we are going to continue using `visibility` at packages level maybe we don't need this change.
I back to the previous version and redefine visibility for packagers [here](https://github.com/okfn/ckanext-unhcr/pull/583/commits/092dabf72162460d799f5b3f6cda49eb2b97b00f)
Look god for you?
The changes for resources `visibility` will be defined in a migration script later

ckanext/unhcr/schemas/dataset.json
avdata99: I start this PR thinking that we want to delete `visibility` from dataset level.
Then I realized that we use `process_status` and `identifiability` firlds at both levels
In fact, we define datasets levels considering the max level from all resources with the `_modify_weighted_field` function.
I think we need to talk about this @chris48s.

chris48s: Yeah I think if we want to keep it as a search facet, maybe what we actually want to do is keep "Internally Visible" in the schema at the dataset level, but hide it on the form (`form_snippet: null`) so it is a calculated field only.

> I think we need to talk about this

We can catch up on it later :+1:

ckanext/unhcr/templates/package/resource_read.html
avdata99: I move back to dataset ID but I think that maybe it won't be clear for the user to see a _Request Access_ button for each resource. If one of them is requested then all resources will be marked as requested.
I will order my ideas a little bit and ask you tomorrow to have a chat about this.

ckanext/unhcr/actions.py
```diff
@@ -990,6 +990,7 @@ def access_request_create(context, data_dict):
     existing_request = model.Session.query(AccessRequest).filter(
         AccessRequest.user_id==user.id,
         AccessRequest.object_id==object_id,
+        AccessRequest.object_type==object_type,
```
avdata99, line 993: This seems an error. Maybe not required. We are assuming no ID collision between packages users and other `object_types`

ckanext/unhcr/jobs.py
```diff
@@ -83,6 +83,11 @@ def _modify_package(package):
     weights = {'personally_identifiable' : 4, 'anonymized_enclave': 3, 'anonymized_scientific': 2,  'anonymized_public': 1}
     package = _modify_weighted_field(package, 'identifiability', weights)
 
+    # visibility
+    # if some of the resources is restricted then the package will be restricted
+    weights = {'restricted' : 2, 'public': 1}
+    package = _modify_weighted_field(package, 'visibility', weights, reset_value=False)
+
```
avdata99, line 90: We now consider `visibility` for packages as the more restricted of its resources. We do the same for other fields like `identifiability`

ckanext/unhcr/jobs.py
```diff
@@ -108,10 +113,12 @@ def _modify_date_range(package, key_start, key_end):
...
 
 
-def _modify_weighted_field(package, key, weights):
+def _modify_weighted_field(package, key, weights, reset_value=True):
+    """ Set a Package level field considering the MAX value for the same field at all resources """
 
     # Reset for generated
-    package[key] = None
+    if reset_value:
+        package[key] = None
```
avdata99, line 121: To be discussed. In `test_jobs.py` we define datasets (without resources) as `public` and we expect them to remain `public`.
Maybe we want to change this behavior or maybe we are not going to define visibility for datasets without resources

chris48s: A dataset is private if it has one or more private resources. A dataset that has zero resources has zero private resources, so I think that makes the dataset public. If at some future point we add a private resource to it, then it becomes private.

ckanext/unhcr/migration/unhcr/versions/004_0a7c08639f19_visibility_for_resources.py
```diff
@@ -0,0 +1,45 @@
...
+def upgrade():
+    """ Select all resources without visibility defined and copy it from package extras """
+    conn = op.get_bind()
+    rows = conn.execute(
+        """
+        SELECT r.id, pe.value as visibility
+        FROM resource as r
+        JOIN package p on r.package_id = p.id
+        JOIN package_extra pe on pe.package_id = p.id and pe.key='visibility'
+         -- NOT REQUIRED -- WHERE (jsonb(r.extras) ->> 'visibility') IS NULL
```
chris48s, line 28: I am confused by the intent of this.
Is this supposed to be

```sql
SELECT r.id, pe.value as visibility
FROM resource as r
JOIN package p on r.package_id = p.id
JOIN package_extra pe on pe.package_id = p.id and pe.key='visibility'
WHERE (jsonb(r.extras) ->> 'visibility') IS NULL;
```

or

```sql
SELECT r.id, pe.value as visibility
FROM resource as r
JOIN package p on r.package_id = p.id
JOIN package_extra pe on pe.package_id = p.id and pe.key='visibility';
```
 ?

ckanext/unhcr/migration/unhcr/versions/004_0a7c08639f19_visibility_for_resources.py
```diff
@@ -0,0 +1,45 @@
...
+            UPDATE resource 
+                SET extras = jsonb_set(cast(extras as jsonb), '{visibility}', '"%s"', true)
+            WHERE id=%s
+            """,
+            row.visibility, row.id
+        )
+
+
+def downgrade():
+    """ Should we remove the 'visibility' key from resource.extras ? """
```
chris48s, line 44: I think we can leave `downgrade()` empty. If we ever need to actually roll back the change we can write an appropriate `down()` migration.

ckanext/unhcr/schemas/deposited_dataset.json
```diff
@@ -93,11 +93,12 @@
...
-            "validators": "not_missing visibility_validator",
-            "preset": "select",
-            "display_snippet": null,
-            "required": true,
+            "display_snippet": null, 
+            "form_snippet": null,
+            "required": false, 
+            "label": "Internal Access Level", 
+            "preset": "select", 
+            "validators": "ignore_missing visibility_validator",
```
chris48s, line 101: If I run `python scripts/generate_deposited_dataset_schema.py` it changes `deposited_dataset.json`.
`deposited_dataset.json` should not be modified manually - it should be the output of running `python scripts/generate_deposited_dataset_schema.py`

This PR should update `generate_deposited_dataset_schema.py` (if needed) and the committed `deposited_dataset.json` should be the output of running that script.

ckanext/unhcr/templates/package/snippets/resource_item.html
```diff
@@ -1,5 +1,20 @@
...
 
+{% block resource_item_title %}
+
+  {{ super() }}
+
+  {% if res.visibility == 'restricted' %}
+
+  <span class="dataset-private label label-inverse">
+      <i class="fa fa-lock"></i>
+      {{ _('Private') }}
```
chris48s, line 11: If we've got several private resources with different length titles this looks a bit messy. I think it would look cleaner if they could align vertically. Can you try either right-aligning them or putting them before the resource title (like we had them before the dataset titles previously) or something to see if we can get this to look a little bit cleaner?

ckanext/unhcr/templates/package/read.html
```diff
@@ -106,17 +100,33 @@ <h1>
...
+  {% if can_request_access %}
+    <a
+      href="#collaboration-dataset-request-access"
+      role="button"
+      class="btn"
+      data-toggle="modal"
+      title="Request access to {{ pkg.title }}"
+      data-target="#collaboration-dataset-request-access"
+    >
+      <i class="fa fa-key"></i>Request Access for private resources in this dataset
```
chris48s, line 125: I think the call to action is a bit too wordy for a button now. I think the best thing to do would be to modify the text in `snippets/request_access.html` from
"You are not authorized to download the resources from this dataset."
to
"You are not authorized to download the **private** resources from this dataset."
and then use `snippets/request_access.html` on both the `dataset.read` and `resource.read` pages. We can ditch the button. That text works equally well in both places.

ckanext/unhcr/templates/package/read.html
```diff
@@ -106,17 +100,33 @@ <h1>
...
-
   {# FIXME why is this here? seems wrong #}
   <span class="insert-comment-thread"></span>
 
 {% endblock %}
 
 {% block package_resources %}
+
+  {% set access_already_requested = h.get_existing_access_request(c.userobj.id, pkg_dict.id, 'requested') %}
+  {% set can_request_access = pkg_dict.resources|length > 0 and not h.can_download(pkg_dict) and not access_already_requested %}
```
chris48s, line 111: We have this same logic block in 3 different templates.
How about we DRY this up by extracting this to a template helper. Then we can call something like `h.can_request_access(pkg_dict, c.userobj)`

chris48s (commented):
Sorry it has taken quite a while to review this. The requirements introduce some complexity and there is a lot going on here. In general we are going in a good direction with this and most of the comments are not big problems. I have left some comments inline. There are also some comments here that don't neatly fit in on the diff:

1. I tried running a DDI import and it failed with

```
Could not upload file: None - {u'visibility': [u'Missing value']}
```

Can you have a look at this please. I will send you some files you can use to test the DDI import on Slack

2. I was really confused about why the `test_both_types_data_fields_missing` is now suddenly passing in this PR. I think the reason is because it is now throwing an error on `resource1` too. Visibility is now a required field for resource but we're not passing it, so `errors[1]` is now set. The intent of that test is that `resource1` should be _valid_.

3. One thing I haven't quite got my head round is how this PR impacts the dataset `private=true/false` field. It seems like now datasets are always `private=False` and we're just relying only on the visibility field at the resource level. @amercader . I think that's ok.. does that meet with your understanding of what the requirements are supposed to be here?

4. We can check this with the client tomorrow, but as I understand it: One of the requirements from the client was that
`Identifiability=Personally Identifiable` implies `Visibility=Private`
see https://github.com/okfn/ckanext-unhcr/issues/581#issuecomment-853834025
i.e: A dataset can be both (`Anonymised` **and** `Private`), but not `both (Personally Identifiable` **and** `Internally Visible`)
Using the code on this branch, I can create a resource which is `Personally Identifiable` and `Internally Visible`. I guess this should be a `ValidationError` on the resource create/edit form?

ckanext/unhcr/migration/unhcr/versions/004_0a7c08639f19_visibility_for_resources.py
avdata99: Both works.
If we want an idempotent query, then is the first one.
If it's a one time query, the second one will change all resources no matter is they already have some value.
My code supposed to run the second one.
I assume this migration runs just one time.

chris48s: Alembic migrations only run once and then it is marked done (similar model to django migrations) the migration doesn't need to handle being run many times

avdata99:
@chris48s About `identifiability` and `visibility`.[Here](https://github.com/okfn/ckanext-unhcr/blob/581_resources_visibility/ckanext/unhcr/jobs.py#L91-L94) we mark dataset as `restricted` if `identifiability=personally_identifiable`
I think I need to move this to a resource level

ckanext/unhcr/templates/package/read.html
avdata99: Added helper [here](https://github.com/okfn/ckanext-unhcr/pull/583/commits/2fd3d939213d44ec34096c6e00486b4159a4f689)

ckanext/unhcr/schemas/dataset.json
```diff
@@ -797,6 +787,16 @@
...
+          {
+            "field_name": "format",
+            "label": "File Format",
+            "preset": "resource_format_autocomplete",
+            "form_attrs": {
+              "class": "",
+              "data-module": "autocomplete",
+              "data-module-source": "/api/2/util/resource/format_autocomplete?incomplete=?"
+            }
+        },
```
avdata99: Just changing places to preserve the previous order. Noting really changed

scripts/generate_deposited_dataset_schema.py
```diff
@@ -15,15 +15,16 @@ def generate_deposited_dataset_schema():
...
     # Remove dataset required flags
-    for field in schema['dataset_fields'] + schema['resource_fields']:
+    for field in schema['dataset_fields']:
         if field['field_name'] not in ['title']:
             field['required'] = False
 
     # Remove resource required flags
-    for field in schema['resource_fields'] + schema['resource_fields']:
-        if field['field_name'] not in ['type']:
+    for field in schema['resource_fields']:
```
avdata99, line 26: this seems a mistake from previous version

scripts/generate_deposited_dataset_schema.py
```diff
@@ -15,15 +15,16 @@ def generate_deposited_dataset_schema():
 
     # Update dataset type
     schema['dataset_type'] = 'deposited-dataset'
+    schema['warning'] = 'This is an automatically generated file. Do not change it manually.'
 
     # Remove dataset required flags
-    for field in schema['dataset_fields'] + schema['resource_fields']:
+    for field in schema['dataset_fields']:
```
avdata99, line 21: I think we want to change only datasets here. Am I ok?

ckanext/unhcr/schemas/deposited_dataset.json
avdata99: I found a couple errors in the script @chris48s 
Changed and applied [here](https://github.com/okfn/ckanext-unhcr/pull/583/commits/f99d1320f65cd7fd4c7a7e048de1c186a9c8f2ca#)

ckanext/unhcr/templates/package/snippets/resource_item.html
avdata99: Changed [here](https://github.com/okfn/ckanext-unhcr/pull/583/commits/3d406980a2eb0aa46de497f3ec15fbf6ba65046a)

image

ckanext/unhcr/templates/package/read.html
avdata99: Changed [here](https://github.com/okfn/ckanext-unhcr/pull/583/commits/4ebf183722bea1eaf8fb9b970e687dbfeaafde90).

image

avdata99:
About DDI import here is my proposed fix: https://github.com/okfn/ckanext-ddi/pull/8

ckanext/unhcr/templates/package/import_package_form.html
```diff
@@ -38,16 +37,6 @@
...
-  {% if show_visibility_selector %}
-    {% block package_metadata_fields_visibility %}
-      {% set schema = h.scheming_get_schema('dataset', 'dataset') %}
-      {% set field = h.scheming_field_by_name(schema.dataset_fields, 'visibility') %}
-      {%- snippet 'scheming/snippets/form_field.html',
-      field=field, data={'visibility': 'public'}, errors=errors, licenses=[],
-      entity_type='dataset', object_type='package' -%}
-    {% endblock %}
-  {% endif %}
-
```
avdata99, line 50: This removes the visibility field from the import form so DDI extension will mark all resources as `restricted` (the default)

chris48s:
Sorry I haven't had time to give this another pass yet.
Just coming back to the `test_both_types_data_fields_missing` test, we've now pushed a core patch to the ckan2.9 branch on the docker repo https://github.com/okfn/docker-ckan-unhcr-aws/pull/35/commits/385cbc9dcf56de799b24a20fed35e281c1ea4f61 back-porting the fix to CKAN core and I just merged a PR to the ckan2.9 branch on this repo applying the same patch in CI and un-skipping the test so you can rebase on to that if you wanted.

avdata99:
> can rebase on to that if you wanted.

Perfect, I just rebased

ckanext/unhcr/templates/package/read.html
```diff
@@ -107,13 +107,10 @@ <h1>
...
 {% block package_resources %}
 
-  {% set access_already_requested = h.get_existing_access_request(c.userobj.id, pkg_dict.id, 'requested') %}
-  {% set can_request_access = pkg_dict.resources|length > 0 and not h.can_download(pkg_dict) and not access_already_requested %}
-
   {% if access_already_requested %}
   <i class="fa fa-key"></i>Access to private resources has already been requested for this dataset
   {% endif %}
-  {% if can_request_access %}
+  {% if h.can_request_access(c.userobj, pkg_dict) %}
```
chris48s: Having a function with a name that implies a boolean return like `can_do_thing()` or `should_do_thing()` return a tuple of booleans is quite confusing - you're even tripping yourself up here by trying to call `{% if h.can_request_access(c.userobj, pkg_dict) %}` in the same commit it is introduced for that exact reason :) I can see you fix it later in 024e2c8 but how about we just make this two helpers? Its slightly less efficient in the case where you need both values but it will make it much easier for our future selves to "fall into the pit of success", especially given any tuple with length >0 is `True` in python, even `(False,)`

ckanext/unhcr/schemas/deposited_dataset.json
```diff
@@ -822,6 +824,24 @@
             "label": "Description",
             "required": false
         },
+        {
+            "field_name": "visibility",
+            "label": "Internal Access Level",
+            "validators": "ignore_empty not_missing visibility_validator",
+            "preset": "select",
+            "display_snippet": null,
+            "required": true,
```
chris48s, line 833: If we are making "Internal Access Level" required for a deposited dataset resource (which I think is correct), should we be adding `ignore_empty` here?

scripts/generate_deposited_dataset_schema.py
chris48s: Yes these look like bugs. What you've done here looks right :+1:

chris48s (commented):
This is really close to being done although I need to have a proper look at the PR on the DDI repo too, which I have not done yet..

chris48s:
I'm not sure what is causing that test to fail with `MailerException` but just to check, I made a branch off the ckan2.9 branch with no changes called `2.9/runthetests` and pushed it to trigger a test run in case it is something related to the CI environment. That build passes https://github.com/okfn/ckanext-unhcr/runs/2803275509 so I think it must be something on this branch causing it. Happy hunting :bug:

ckanext/unhcr/migration/unhcr/versions/004_0a7c08639f19_visibility_for_resources.py
```diff
@@ -0,0 +1,43 @@
...
+        JOIN package p on r.package_id = p.id
+        JOIN package_extra pe on pe.package_id = p.id and pe.key='visibility'
+        """
+    ).fetchall()
+
+    for row in rows:
+        conn.execute(
+            """
+            UPDATE resource 
+                SET extras = jsonb_set(cast(extras as jsonb), '{visibility}', '"%s"', true)
```
chris48s, line 35: Hmm. I tried to apply this migration and it threw

```
(psycopg2.errors.SyntaxError) syntax error at or near "restricted"
LINE 3: ...onb_set(cast(extras as jsonb), '{visibility}', '"'restricted...
```

for me. Can you have a look and see if you can reproduce?

avdata99: I run both commands locally and worked. Maybe your DB has a real database with `visibility` at a dataset level.
I don't understand these quotes at the beginning  of`'"'restricted`. I mean: single quote + double quote + single quote again.
Maybe you DB have some unexpected data.
I propose this change: https://github.com/okfn/ckanext-unhcr/pull/583/commits/deed097431b3b6ff4262a0d28aa097674f31cd1a
What do you think of it?

Can you show me more on the error log if it fails again?

ckanext/unhcr/schemas/deposited_dataset.json
avdata99: I changed the script [here](https://github.com/okfn/ckanext-unhcr/pull/583/commits/3cb2223f3fc39268b83b52cc6f71cdcf16213fc8). Also run it and change base dataset schema

ckanext/unhcr/migration/unhcr/versions/004_0a7c08639f19_visibility_for_resources.py
chris48s: > Maybe your DB has a real database with `visibility` at a dataset level.

Yes it does. So will the production environment, and yours should too in order to test this properly :)

I've not tested this in any detail, but I think the extra params you pass to `conn.execute()` get single-quoted if they are string params (i.e: this is not a string format operation - its SQL aware) so when you write:

`'"%s"'` (with two sets of quotes)
that expands to
`'"'restricted'"'` (with three sets of quotes)
after the substitution.

Can you check the code against a DB with one or more datasets that have visibility at the dataset level before pushing a fix to review please.

avdata99: I downgrade to CKAN2.9 base branch and create some datasets and resources
I make this changes: https://github.com/okfn/ckanext-unhcr/pull/583/commits/bef976898e4870ca2c03822c491f197d6cef3d50 (It's just was some string formatting problems)

Asking the DB I can see
```
ckan=# SELECT r.id, pe.value as visibility
ckan-#         FROM resource as r
ckan-#         JOIN package p on r.package_id = p.id
ckan-#         JOIN package_extra pe on pe.package_id = p.id and pe.key='visibility';
                  id                  | visibility 
--------------------------------------+------------
 722c40a7-f230-490e-b936-ae36e33d2433 | restricted
 993087d9-8b5a-47bb-bd83-683ef33e8891 | restricted
 d6dbbf55-1686-4c1d-b57c-9060be5a0324 | public
 1134389d-6e0c-4caf-b052-61cf3a4c7a72 | restricted
(4 rows)
```

The move to this branch and start the container again

I can see:
```
postgresql    | 2021-06-15 21:23:56.732 UTC [80] ERROR:  relation "migrate_version" does not exist at character 21
postgresql    | 2021-06-15 21:23:56.732 UTC [80] STATEMENT:  select version from migrate_version
ckan-dev      | Upgrading DB: SUCCESS
```

And ask the database again for resource's extras:

```
select id, extras from resource;
 id  | extras 
-----+-----------
 722c40a7-f230-490e-b936-ae36e33d2433 | {"type": "data", "version": "2", "file_type": "microdata", "visibility": "restricted", "date_range_end": "2021-06-08", "process_status": "raw", "identifiability": "personally_identifiable", "datastore_active": false, "date_range_start": "2021-06-01"}
 993087d9-8b5a-47bb-bd83-683ef33e8891 | {"type": "attachment", "file_type": "report", "visibility": "restricted", "datastore_active": false}
 d6dbbf55-1686-4c1d-b57c-9060be5a0324 | {"type": "attachment", "file_type": "questionnaire", "visibility": "public", "datastore_active": false}
 1134389d-6e0c-4caf-b052-61cf3a4c7a72 | {"type": "attachment", "file_type": "infographics", "visibility": "restricted", "datastore_active": false}
(4 rows)
```

I thinks is done

avdata99: For some reason I can't see the _Private_ label after the migration

image

I will check what's happening

chris48s:
I was doing a bit more testing on this branch and I found that there is something preventing me from creating new resources. Works fine on master. If I check out this branch, I'm just getting booted back to

Screenshot at 2021-06-16 10-16-44

with no obvious validation message. I guess it is a schema issue. Can you see if you can reproduce this and have a look.

ckanext/unhcr/migration/unhcr/versions/004_0a7c08639f19_visibility_for_resources.py
chris48s: This seems to be working ok for me.. Did you need to rebuild your search index after applying the migration?

avdata99: Yes, that fixed the problem

avdata99:
> I guess it is a schema issue. Can you see if you can reproduce this and have a look.

I tried to reproduce the error but can't do it.
I tested inside deposit and regular datasets but everything it's working fine here. With both new and existent datasets.
Did you remember the context @chris48s?

chris48s:
Here's a screen capture of me reproducing the issue on this branch - works fine on master

GIFrecord_2021-06-16_124808

Nothing in the docker logs :/

avdata99:
I can't reproduce the error in this branch @chris48s. I tried in multiple ways but it's always working

chris48s:
Right I've done a bit of digging into this to find out what is going on..

Basically, what I've done is:
- Applied your migration (004) with `581_resources_visibility` checked out
- Switched back to master, done a bunch of work on other branches, etc
- Created some resources (with different branches checked out) that don't have a `visibility` key set
- When I switch back to `581_resources_visibility` and try to add a new resource to one of my test datasets, I get an error like `{u'resources': [{u'visibility': [u'Missing value']}, {}]` when we call `package_update` from `resource_create` here: https://github.com/ckan/ckan/blob/ckan-2.9.3/ckan/logic/action/create.py#L319 but the error is on another resource in the dataset, not the one I'm trying to add, so no error shows up on the page.
- If I add a new resource to a dataset where all the existing resource have `visibility` set then it works.
- the reason you're not hitting this is because you have a `visibility` field set on every resource in your DB.

So I guess the long and the short of it is: As long as we are sure that our migration will result in all existing resources in the DB having a visibility field set, we're ok - this error shouldn't affect us when we deploy.

ckanext/unhcr/jobs.py
```diff
@@ -84,9 +89,9 @@ def _modify_package(package):
...
 
     # visibility
-    # TODO: clarify what level of anonymization required
-    if package['identifiability'] == 'personally_identifiable':
-        package['visibility'] = 'restricted'
+    # if some of the resources is restricted then the package will be restricted
+    weights = {'restricted' : 2, 'public': 1}
+    package = _modify_weighted_field(package, 'visibility', weights, reset_value=False)
 
     return package
```
chris48s, line 96: This isn't a merge conflict as such, but before we retarget this at master and merge this we will need to account for #600 .
#600 moves us from using `package_revise` instead of `package_update`. It is a concurrency safe operation (which is important for background jobs) but it has a different API than package_update.

Related reading for package_revise:
- https://docs.ckan.org/en/2.9/api/index.html#ckan.logic.action.update.package_update
- https://github.com/ckan/ckan/pull/4618

One thing I'm not 100% sure on is how package_revise handles resources so there is a bit of research to do there. I guess we'll need to change `_process_dataset_fields` to either explicitly pass the resource objects or just the specific resource fields we want to update.

avdata99:
@chris48s let me know if you think [this change](https://github.com/okfn/ckanext-unhcr/pull/583/commits/e3bf0b7bb8284e5d64b2b70fda807afa4adb6424) is correct
I'm struggling to hide a field for regular dataset creation and show it for the import form

ckanext/unhcr/templates/package/read.html
avdata99: I created #614 @chris48s connect it to _CKAN 2.9 Upgrade_ milestone.
Do you think it is enough?

chris48s:
> I'm struggling to hide a field for regular dataset creation and show it for the import form

I'm not sure I have a better suggestion for this (maybe @amercader has a suggestion?), but having tested this solution I notice it doesn't maintain its value if there is a form validation error like everything else on the form.

When I was testing this I found if I add or edit a resource I am getting

```
File "/srv/app/src_extensions/ckanext-unhcr/ckanext/unhcr/jobs.py", line 94, in _modify_package
  package = _modify_weighted_field(package, 'visibility', weights, reset_value=False)
File "/srv/app/src_extensions/ckanext-unhcr/ckanext/unhcr/jobs.py", line 127, in _modify_weighted_field
  package_weight = weights.get(package[key], 0)
KeyError: 'visibility'
```

thrown in the background job.

ckanext/unhcr/jobs.py
```diff
@@ -108,17 +113,18 @@ def _modify_date_range(package, key_start, key_end):
...
-    package[key] = None
+    package[key] = default_value
 
     # Iterate resources
     for resource in package['resources']:
         if resource.get(key) is None:
             continue
-        package_weight = weights.get(package[key], 0)
-        resource_weight = weights.get(resource[key], 0)
+        package_weight = weights.get(package.get(key), 0)
```
avdata99, line 126: "Visibility" is no longer a required field, so it may not be present.

ckanext/unhcr/helpers.py
```diff
@@ -972,6 +972,18 @@ def can_download(package_dict):
...
+def can_request_access(user, pkg):
+    """ Define if user is able to request access for private resources in dataset """
+    resources = pkg['resources'] if type(pkg) == dict else pkg.resources
+    if len(resources) == 0:
+        return False
+    if can_download(pkg):
+        return False
+    pkg_id = pkg['id'] if type(pkg) == dict else pkg.id
+    access_already_requested = get_existing_access_request(user.id, pkg_id, 'requested')
+    return not access_already_requested
```
avdata99, line 984: Helper to use in the UI to define whether the access request form should be available

ckanext/unhcr/tests/pt/controllers/test_dataset.py
```diff
@@ -456,6 +456,7 @@ def test_visibility_private(self, app):
         resource = factories.Resource(
             package_id=dataset['id'],
             url_type='upload',
+            visibility='restricted',
```
avdata99, line 459: It is not clear how `visibility='private',` should work now that `visibility` is for resources and just a replica within datasets

ckanext/unhcr/validators.py
```diff
@@ -192,6 +192,16 @@ def visibility_validator(key, data, error, context):
...
 
 
+def visibility_validator_resource(key, data, error, context):
+    ''' Validates visibility has a correct value.
+    Visibility only has two values in the schema, 'restricted' and 'public'. 
+    '''
+    value = data.get(key)
+
+    if value not in ('restricted', 'public'):
+        raise Invalid('Invalid value for the visibility field')
```
avdata99, line 202: @chris48s It is not clear how `visibility='private',` should work now that `visibility` is for resources and just a replica within datasets.
We still have the `visibility_validator` for datasets but this behavior will be overridden by the resources visibility

avdata99:
> > I'm struggling to hide a field for regular dataset creation and show it for the import form
> 
> I'm not sure I have a better suggestion for this (maybe @amercader has a suggestion?), but having tested this solution I notice it doesn't maintain its value if there is a form validation error like everything else on the form.

I think this should do the job (to define the visibility for resources included in the import process)

> When I was testing this I found if I add or edit a resource I am getting
> 
> ```
> File "/srv/app/src_extensions/ckanext-unhcr/ckanext/unhcr/jobs.py", line 94, in _modify_package
>   package = _modify_weighted_field(package, 'visibility', weights, reset_value=False)
> File "/srv/app/src_extensions/ckanext-unhcr/ckanext/unhcr/jobs.py", line 127, in _modify_weighted_field
>   package_weight = weights.get(package[key], 0)
> KeyError: 'visibility'
> ```
> 
> thrown in the background job.

I just fixed it @chris48s. It's related with the fact that now `visibility` is not required for datasets.

chris48s:
> > > I'm struggling to hide a field for regular dataset creation and show it for the import form
> > 
> > I'm not sure I have a better suggestion for this (maybe @amercader has a suggestion?), but having tested this solution I notice it doesn't maintain its value if there is a form validation error like everything else on the form.
> 
> I think this should do the job (to define the visibility for resources included in the import process)

I pulled down the latest changes and the behaviour is the same. Perhaps I didn't explain this well.

Steps to reproduce:

- https://ckan-dev:5000/dataset/import
- Set "Internal Access Level" to "Internally Visible"
- Upload some files that cause a validation error (for example, uploading a file with an existing ID which throws "Dataset already exists and duplicates are not allowed.")

Observed:
- Value of "Internal Access Level" is now "Private" on the form

Expected:
- Value of "Internal Access Level" should still be "Internally Visible" on the form

ckanext/unhcr/jobs.py
avdata99: I created #614 @chris48s connect it to CKAN 2.9 Upgrade milestone.
Do you think it is enough?

avdata99:
> > > > I'm struggling to hide a field for regular dataset creation and show it for the import form
> > > 
> > > 
> > > I'm not sure I have a better suggestion for this (maybe @amercader has a suggestion?), but having tested this solution I notice it doesn't maintain its value if there is a form validation error like everything else on the form.
> > 
> > 
> > I think this should do the job (to define the visibility for resources included in the import process)
> 
> I pulled down the latest changes and the behaviour is the same. Perhaps I didn't explain this well.
> 
> Steps to reproduce:
> 
> * https://ckan-dev:5000/dataset/import
> * Set "Internal Access Level" to "Internally Visible"
> * Upload some files that cause a validation error (for example, uploading a file with an existing ID which throws "Dataset already exists and duplicates are not allowed.")
> 
> Observed:
> 
> * Value of "Internal Access Level" is now "Private" on the form
> 
> Expected:
> 
> * Value of "Internal Access Level" should still be "Internally Visible" on the form

This fixed it @chris48s: https://github.com/okfn/ckanext-unhcr/pull/583/commits/0bf7732bf8fe9621009833cd708e23ac1e92eea4

avdata99:
@chris48s, this PR started from `ckan-2.9` which is not merged to master.
I change the base on this PR to `master` and fixed conflict. I think it will be better now than after merging py3 branch.
If this is a problem I can cherry pick the right commits and create a new PR to `master` after merging py3.

chris48s:
> I change the base on this PR to master and fixed conflict.

I don't think this has gone well - this doesn't look right: https://github.com/okfn/ckanext-unhcr/pull/583/commits
If you need to start a clean branch off master and cherry-pick to it, go for it..

Also, one of the commits that is now on master is: https://github.com/okfn/ckanext-unhcr/commit/6e4cabd017ac61ab0ac64b2f30152ad1f522f400
This changes us from using `package_update` to using `package_revise` to modify the package object in background jobs for concurrency-safety.

In order to modify resources, you will need to update the code on this branch to pass the resource objects (or the resource fields we want to update) to `package_revise`. As it stands, changes to resource will be discarded.
Relevant docs are linked in https://github.com/okfn/ckanext-unhcr/pull/583/files#r653581140

avdata99:
Rebased to `master` and continue at #628

Issue #564, opened 2021-05-24
CKAN 2.9: Resolve remaining test failures


chris48s:
You have still got a small number of test failures outstanding on the CKAN 2.9 branch. Review these and fix the bugs/update the tests as applicable. Getting the build green will give us more confidence and make development/review easier as we work through the rest of the migration tasks.

chris48s:
I think the only thing left that isn't covered by an open PR is the
`IndexError: list index out of range` fail
on `test_both_types_data_fields_missing`
in `test_schema.py`

chris48s:
The final test failure is down to an issue in core.
Reverting this commit https://github.com/ckan/ckan/commit/7b6702b6c17efbe0b2b7eb46ab46da86e387c55c takes us back to the CKAN 2.8 behaviour, which gets the test to pass
@amercader to pick up..

chris48s:
refs https://github.com/ckan/ckan/pull/6119
For now I'm going to mark it skipped
When you circle back to this, also replace `'url': 'http://example.com/data.csv'` with a `FakeFileStorage`

amercader:
I've created https://github.com/okfn/docker-ckan-unhcr-aws/pull/40 which contains the patch I submitted to CKAN core addressing this https://github.com/ckan/ckan/pull/6149

PR #587, opened 2021-06-02
CKAN 2.9: skip failing test (for now)

chris48s:
Refs https://github.com/okfn/ckanext-unhcr/issues/564
Refs https://github.com/ckan/ckan/pull/6119

We've got 1 failing test left, but it isn't failing because the code in ckanext-unhcr is broken. It is failing because of a bug in core.

I think at this point, it is more useful to get the build status green on the 2.9 branch. This will make it more obvious if we are breaking the build in PRs. We can un-skip it once the upstream bug is fixed and we can backport it with a core patch.


chris48s:
oh, and we get coverage reports back :bar_chart:

amercader (approved):

Issue #522, opened 2021-04-19
Poor UX on password reset form


chris48s:
If you request a password reset, it dumps you back to `/` with no confirmation that anything has happened because the template doesn't contain anywhere to display the notification saying "A reset link has been emailed to you"

There is a slightly wider issue in this theme where sometimes error or confirmation notifications don't appear on the right page because not all of the templates (especially the templates for errors) don't have anywhere to display the notification so they just get stored in the session and shown on the next page that does have the right template code. Review this.

PR #589, opened 2021-06-03
CKAN 2.9: prevent SAML2 users from requesting password reset

### LINKED ISSUE #467: Support Saml2 login in CKAN 2.9
Part of #399 

Currently assessing [ckanext-saml2auth](https://github.com/keitaroinc/ckanext-saml2auth), in these steps:

1. Make it work on a standard CKAN 2.9 install with RIDL credentials
2. Investigate what would be needed to make it run on Python 2
3. Potential migration path from our current models based on okfn/ckanext-saml2

chris48s:
Closes #467

This PR prevents SAML2 users from requesting a password reset but allows them for native users (if enabled).

Note the user experience for requesting a password reset as an external user is not great, but this is noted in https://github.com/okfn/ckanext-unhcr/issues/522 I've not done anything about it here. Moving to CKAN 2.9 (the main focus for now) doesn't make this any worse than it was in 2.8 so we can deal with it another time.


ckanext/unhcr/auth.py
```diff
@@ -441,3 +440,51 @@ def user_show(next_auth, context, data_dict):
...
+
+
+def _get_user(username_or_email):
+    userobj = model.User.get(username_or_email)
+    if userobj:
+        return userobj
+    users = model.User.by_email(username_or_email)
+    if len(users) == 1:
+        return users[0]
+    return None
```
chris48s, line 464: There is an edge case here if there are >1 user accounts with the same email.
I think it makes sense to just `return None` (and allow the reset) in this case because it is super-unlikely they are both SAML2 accounts. I'm not sure there can be 2 accounts with the same email in one Active Directory tree. Also we now prevent multiple signups with the same email so this is probably a legacy problem we can eliminate. I'll check this when I have a look at applying migrations to the existing DBs. I think CKAN 2.9 comes with a script for identifying duplicate emails.

ckanext/unhcr/auth.py
```diff
@@ -32,8 +33,6 @@ def restrict_access_to_get_auth_functions():
         'organization_list_for_user',  # Because of #4097
         'get_site_user',
         'user_reset',  # saml2
-        'user_create',  # saml2
```
chris48s, line 35: I don't _think_ we need to allow anon access for user_create any more. The external signup form calls `user_create` with `ignore_auth: True` and the signup process is working correctly. with this removed. Is there something else this might be required for?

ckanext/unhcr/auth.py
```diff
@@ -32,8 +33,6 @@ def restrict_access_to_get_auth_functions():
...
-        'user_delete',  # saml2
```
chris48s, line 36: I'm not really 100% sure what we want to do about `user_delete`.. For the moment I haven't defined any custom auth for it. Do you have any thoughts on the custom behaviour we ideally want here (if any)?

ckanext/unhcr/auth.py
amercader: I can't remember if we allow to send user invites from the data container Members page. I'm almost sure this was removed but it's worth checking as this might require `user_create`

ckanext/unhcr/auth.py
amercader: I think the default behaviour, allowing sysadmins and the user themselves to delete users makes sense.

Would be good to test at some point what happens when:

1. Someone with a UNHCR.org account deletes their user and they login again
2. An external user deletes their user and they try to reapply again

I'm pretty sure that on 1 the deleted user account is reactivated by ckanext-saml2auth, but I'm  not sure about 2

amercader (commented):
This looks good so far @chris48s. I didn't see any related changes to `user_update`. I'll double check on Monday because I think the old ckanext-saml2 extension prevented edits to daml users with an exception for API keys

chris48s:
> I didn't see any related changes to user_update. I'll double check on Monday because I think the old `ckanext-saml2` extension prevented edits to daml users with an exception for API keys

You're right. I missed it because it is an action not an auth function. I'll have a look at porting this over too

chris48s:
I've pushed another commit porting over that custom `user_update` action (but simplifying because the keitaro plugin does not have an `allow_user_changes` setting). We don't need to do anything special about API keys because there is a separate `user_generate_apikey` action. Cycling the key isn't a `user_update` so I think we just leave that alone.

amercader:
I checked and clicking the "Regenerate API Key" button does throw a `ValidationError`

```
  File "/srv/app/src_extensions/ckan/ckan/views/user.py", line 531, in generate_apikey
    result = logic.get_action(u'user_generate_apikey')(context, data_dict)
  File "/srv/app/src_extensions/ckan/ckan/logic/__init__.py", line 477, in wrapped
  File "/srv/app/src_extensions/ckan/ckan/logic/action/update.py", line 904, in user_generate_apikey
    return _get_action('user_update')(context, data_dict)
  File "/srv/app/src_extensions/ckan/ckan/logic/__init__.py", line 477, in wrapped
  File "/srv/app/src_extensions/ckanext-unhcr/ckanext/unhcr/actions.py", line 1273, in user_update
    "User accounts managed by Single Sign-On can't be modified"
ValidationError: None - {'error': ["User accounts managed by Single Sign-On can't be modified"]}
```

Screenshot 2021-06-09 at 12-43-14 Manage - Adria Mercader - Users - RIDL

Maybe to make this easy (as we eventually want to phase out API keys) you can just check if the incoming `data_dict['apikey']` and the `user_obj['apikey']` don't match and allow this update (but only on this field, eg setting the new api key value in the existing dict coming from `user_show` and passing this to `user_update`)

chris48s:
Yeah you're totally right. I noticed there  was special action for `user_generate_apikey` but didn't realise it calls `package_update` internally. I think I must have only tested this as a native user. I've fixed the `user_update` action and added a test case.

ckanext/unhcr/actions.py
```diff
@@ -1268,6 +1268,8 @@ def user_update(up_func, context, data_dict):
     user_id = toolkit.get_or_bust(data_dict, 'id')
     m = context.get('model', model)
     user_obj = m.User.get(user_id)
+    if user_obj is not None and user_obj.apikey != data_dict.get('apikey'):
+        return up_func(context, data_dict)
```
amercader, line 1272: This would allow a SAML user to bypass the limitation to edit the user details by calling `user_update` with a data_dict containing a new `apikey` plus extra params, eg 
```
{
    "id": "some_saml_user",
    "apikey": "xxxxx",
    "email": "some new email",
}
```
To avoid it we can pass the result of `user_show` with the updated api key to `up_func()`

amercader (approved):

Issue #559, opened 2021-05-18
CKAN 2.9: `AttributeError: 'NoneType' object has no attribute 'transaction'` editing resource


chris48s:
Steps to reproduce

1. Create a dataset
2. Add a resource to it
3. Edit the resource
4. Replace the uploaded file with a different one (can have the same name or different name - throws either way)

Stacktrace:

```
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/flask/app.py", line 2449, in wsgi_app
    response = self.handle_exception(e)
  File "/usr/lib/python2.7/site-packages/flask/app.py", line 1866, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/lib/python2.7/site-packages/flask/app.py", line 2446, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/lib/python2.7/site-packages/flask/app.py", line 1951, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/lib/python2.7/site-packages/flask/app.py", line 1820, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/lib/python2.7/site-packages/flask/app.py", line 1949, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/lib/python2.7/site-packages/flask_debugtoolbar/__init__.py", line 125, in dispatch_request
    return view_func(**req.view_args)
  File "/usr/lib/python2.7/site-packages/flask/views.py", line 89, in view
    return self.dispatch_request(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/flask/views.py", line 163, in dispatch_request
    return meth(*args, **kwargs)
  File "/srv/app/src_extensions/ckan/ckan/views/resource.py", line 371, in post
    get_action(u'resource_update')(context, data)
  File "/srv/app/src_extensions/ckan/ckan/logic/__init__.py", line 477, in wrapped
    result = _action(context, data_dict, **kw)
  File "/srv/app/src_extensions/ckanext-unhcr/ckanext/unhcr/actions.py", line 863, in resource_update
    resource = up_func(context, data_dict)
  File "/srv/app/src_extensions/ckan/ckan/logic/action/update.py", line 104, in resource_update
    updated_pkg_dict = _get_action('package_update')(context, pkg_dict)
  File "/srv/app/src_extensions/ckan/ckan/logic/__init__.py", line 477, in wrapped
[...]

PR #591, opened 2021-06-04
CKAN 2.9: use local session in `task_status_update`

chris48s:
Might be useful for us to catch up on this one next week @amercader - Background reading:

- #559
- https://github.com/okfn/docker-ckan-unhcr-aws/pull/35#pullrequestreview-670304673
- https://github.com/ckan/ckan/issues/5597#issuecomment-719512398


chris48s:
TODO: look at https://github.com/ckan/ckan/pull/6137

amercader:
@chris48s I think the ckan/ckan#6137 approach is slightly cleaner, I've merged that upstream and backported to 2.9.4b, here's the cleaned up patch:

https://github.com/ckan/ckan/commit/cf4a537436260ffef3de1478ff82873e00ea91b9

chris48s:
OK :+1:  I've replaced the patch file on the ckan2.9 branch of the docker repo with https://github.com/ckan/ckan/commit/cf4a537436260ffef3de1478ff82873e00ea91b9 We can bin this PR then..I was thinking I would need to modify the call to `task_status_update` in `scan_hook` to pass a local session in the context too, but it seems we don't so :shrug:

PR #595, opened 2021-06-08
CKAN 2.9: ensure webasset cache owned by ckan after test run

### LINKED ISSUE #562: CKAN 2.9: `Permission denied: u'/var/lib/ckan/webassets/...`
To reproduce, run the tests.
Then you next interaction though the web UI will throw something like
```
IOError: [Errno 13] Permission denied: u'/var/lib/ckan/webassets/.webassets-cache/4e32c268e935039c45a92f30b043293b'
```
You can fix by SSHing in and running `chown -R ckan:ckan webassets/` but it is an unnecessary faff
This might need fixing on the docker repo rather than here..

chris48s:
crude but effective solution to fix #562
open to other ideas


amercader:
I think this should be fixed in the Docker images. Probably by explicitly creating the `/var/lib/ckan/webassets` path [here](https://github.com/okfn/docker-ckan/blob/bf1ac27d2699ef490d68a09e90b86bdef980e2e0/ckan-base/2.9/Dockerfile.py2#L75) and chowning it to the ckan user.

amercader:
How do you run the tests to reproduce this @chris48s . I thought the default user in the image was the `ckan` one (ie `USER ckan` was set up in the Dockerfile) but looking at it now this might not be the case.

chris48s:
So I might have slightly clouded the issue here by chowning the whole `/var/lib/ckan/webassets` dir at the end of the test run. I think it is only actually `/var/lib/ckan/webassets/.webassets-cache` that we really need to worry about. When I was trying to reproduce this the other day I was only seeing it on files in the `.webassets-cache` dir but I couldn't remember if that was the case 100% of the time.

We could probably just chown (or delete) `/var/lib/ckan/webassets/.webassets-cache` at the end of the test run.

As far as I can tell, all the other files under webassets do get created on boot and are owned by ckan.

Reproducing is a bit less trivial than originally suggested in #562 . I guess it is because it depends if any of the cache objects under `.webassets-cache` expire during your test run or not but you can `rm -rf /var/lib/ckan/webassets/.webassets-cache` and then run any controller test to reliably throw the error.

AFAIK everything in the CKAN container runs as root. See also https://github.com/okfn/docker-ckan-unhcr-aws/pull/17 which is semi-related

PR #596, opened 2021-06-08
CKAN 2.9: remove use of c.controller and c.action

chris48s:
clsoes #590


amercader (approved):
Maybe we embrace the new times and replace "controller, action" by "blueprint, view"?

chris48s:
the future is now :rocket:

PR #600, opened 2021-06-10
CKAN 2.9: Use package_revise in background job

chris48s:
Refs  #509

w00p :tada: 
Unsurprisingly, switching to a concurrency-safe operation here eliminates at least one known class of race condition which I have a reliable repro case for. Probably others too. Unfortunately this conflicts with PR #583 so one of us is going to have to update our PR but it is useful to prove the concept in any case.


amercader (approved):

chris48s:
I'm going to go ahead and merge this one. We'll sort out rebasing #583 when we get to it..

PR #601, opened 2021-06-10
Move to chained actions

avdata99:
**To be merged after #574**
Related to #558 
 

Finish migration for top chained actions and auths


ckanext/unhcr/actions.py
```diff
@@ -899,6 +899,10 @@ def access_request_update(context, data_dict):
     if not request:
         raise toolkit.ObjectNotFound("Access Request not found")
 
+    if request.object_type not in ['organization', 'package', 'user']:
+        raise toolkit.Invalid("Unknown Object Type")
```
avdata99, line 903: Moving from `auth` file

ckanext/unhcr/auth.py
```diff
@@ -92,7 +93,7 @@ def organization_show(next_auth, context, data_dict):
     return next_auth(context, data_dict)
 
 
-def organization_list_all_fields(next_auth, context, data_dict):
```
avdata99, line 95: probably a mistake

ckanext/unhcr/actions.py
```diff
@@ -899,6 +899,10 @@ def access_request_update(context, data_dict):
...
+
+    context['request'] = request
```
avdata99, line 905: Skip load the request again in `auth` function

chris48s:
I haven't started reviewing this one yet. Before I start, can we make a PR targetting master (including https://github.com/okfn/ckanext-unhcr/commit/03f0275eae482e10fc85b8824c1d4a9b02aef106 ) and I'll pick up the review on that.
This will get easier now we have a single working branch

avdata99:
Should I move this to PR to `master` @chris48s?

ckanext/unhcr/actions.py
amercader: In general, the less we (ab)use the context the better, as storing things there leads to hard to track bugs. I think the overhead of another simple db query is not a big deal, but if you want to cache the request object for the auth function delete it from the context after it's called ( `del context["request"]`) (like in this [core action](https://github.com/ckan/ckan/blob/891e6aa6542edc108bb6ee07a4f023199e9d5b36/ckan/logic/action/update.py#L83))

chris48s:
Yes. Now that https://github.com/okfn/ckanext-unhcr/commit/03f0275eae482e10fc85b8824c1d4a9b02aef106 is on master it doesn't make sense for this PR to use `511_use_toolkit_actions` as the target branch. You can do it by rebase or start a new branch off master and cherry-pick on to it.. whatever works for you

avdata99:
Moved to #613

PR #603, opened 2021-06-11
CKAN 2.9 🚀

chris48s:
Eek. Not much going on here tbh. Just a couple of minor changes :grimacing: 

This branch is the same work as the `ckan2.9` branch but I have:

- rebased on to master
- resolved the merge conflicts
- rolled back the MFA changes

I think when I started this had the lofty goal of trying to squash some of the early FIXME commits into the solutions to clean up the history. I actually did this on the docker repo but I think what we have here is just too complicated so I have not attempted to that: The history pretty much just is what it is..

so this is the one we should actually merge. I probably could have just force-pushed ckan2.9 but I'm conscious Andres is working off it too


ckanext/unhcr/__init__.py
```diff
@@ -1 +1 @@
-__VERSION__ = '2.3.3'
+__VERSION__ = '3.0.0'
```
chris48s, line 1: TODO: tag this on Monday

Issue #598, opened 2021-06-10
Expiry dates for partner users


avdata99:
When a user signs up with a partner account, we [set some extra fields on the user object](https://github.com/okfn/ckanext-unhcr/blob/0fa381bc90aa0733278b3a0d0c222196a5ddd070/ckanext/unhcr/actions.py#L1231-L1241)  
These are returned when we do a [user_show](https://github.com/okfn/ckanext-unhcr/blob/0fa381bc90aa0733278b3a0d0c222196a5ddd070/ckanext/unhcr/actions.py#L1210-L1214) 

One of the things we're going to need is a CLI job that we can run once a day to find users whose expiry date has passed. We currently have two of these ([[1]](https://github.com/okfn/ckanext-unhcr/blob/0fa381bc90aa0733278b3a0d0c222196a5ddd070/ckanext/unhcr/click_commands.py#L34) [[2]](https://github.com/okfn/ckanext-unhcr/blob/0fa381bc90aa0733278b3a0d0c222196a5ddd070/ckanext/unhcr/click_commands.py#L67)) and they are scheduled with [yacron in the docker repo](https://github.com/okfn/docker-ckan-unhcr-aws/blob/ckan2.9/ckan/yacrontab.yaml). 

So as a first step we can write a click CLI command that finds any user whose expiry date is in the past. Then one of the requirements we are certain of is that if the user has not made any activity since signing up we just want to deactivate them. 
We can check the user's activity stream - if the most recent activity was "account create" or something, we can set them to `state=deleted`.
I think we can do that bit without too much worry that it will change.

The case where they do have activity is slightly less clear at the moment what the interaction will be but one component is that we will want the ability for some internal user to approve or reject extending the expiry date. If they approve, we add another 180 days to `expiry_date`. 
If they reject we set `state=deleted`. 
We have this concept in RIDL of the AccessRequest object. Users can see a list of requests they can approve/reject at /dashboard/requests and the endpoints for approving/rejecting them are [here](https://github.com/okfn/ckanext-unhcr/blob/ckan2.9/ckanext/unhcr/blueprints/access_requests.py).

Although it is still ambiguous exactly how we will create the AccessRequest object, I think we can assume that a "account extension request" will be a type of AccessRequest, decide what that looks like and implement the approve/reject actions.

PR #604, opened 2021-06-14
Expire users

avdata99:
Related to #598 

Adding new CLI command to expire users 

Test added for:
 - The new CLI command
 - The internal function to detect expired users
 - The new request access to renew users


avdata99:
@chris48s this is not finished but it will be a good idea to know if I'm in the right path.

ckanext/unhcr/click_commands.py
```diff
@@ -88,3 +89,60 @@ def send_summary_emails(verbose):
...
+            }
+
+            message = 'User %(user_name)s expired and "expire-user" command automatically ask ' \
+                      'for renewal. Last activity: %(last_activity)s registered on %(last_date)s' % data
+
+            context = {'user': expired_user['id'], 'ignore_auth': True}
+            toolkit.get_action('access_request_create')(
+                context, {
+                    'object_id': expired_user['id'],
+                    'object_type': 'user',
```
chris48s, line 131: Just to flag this: When it comes to this AccessRequest object, an AccessRequest of type "user" will get listed on the dashboard page under "Requests for new User Accounts" as "External user xxx requested access to deposit datasets" which isn't really right so when we get to that we're going to need to differentiate requests for new user accounts from requests to extend a user account.
I think there's a few different ways round this:

- One would be to modify the `object_type` enum so that 'new user' and 'extend user' are two different things and make a migration to convert the existing 'user' requests to 'new user'
- The other option would be to leave `object_type` as user (which is more accurate) and put something in the `data` JSON field to tell us what sort of request it is.

I've not fully thought through the tradeoffs of this, but bear in mind it is something to think through.

chris48s (commented):
Had a quick look over this. Not done a full review but it seems to be going in a sensible direction.

avdata99:
Rebased to `master`

ckanext/unhcr/click_commands.py
avdata99: I add more info to `data` @chris48s. Also added a test case to validate it's working

ckanext/unhcr/click_commands.py
```diff
@@ -88,3 +89,64 @@ def send_summary_emails(verbose):
...
+                    'message': message,
+                    'role': 'member',  # TODO this is only required to fit the schema
+                    'data': {
+                        'user_request_type': USER_REQUEST_TYPE_RENEWAL
+                    }
+                }
+            )
+            print(' - Renewal access requested')
```
avdata99, line 140: Once the user expires we request for renewal but we didn't delete the user until this request is rejected
Is this behaviour ok @chris48s? 
Or we should delete the user immediately?

ckanext/unhcr/actions.py
```diff
@@ -925,16 +928,25 @@ def access_request_update(context, data_dict):
                 context, _data_dict
             )
     elif request.object_type == 'user':
+        data = request.data
         state = {'approved':  m.State.ACTIVE, 'rejected': m.State.DELETED}[status]
         _data_dict = {'id': request.object_id, 'state': state}
+
+        if data.get('user_request_type', USER_REQUEST_TYPE_NEW) == USER_REQUEST_TYPE_RENEWAL:
+            _data_dict['renew_expiry_date'] = True
```
avdata99, line 936: Mark this user to be updated by adding days to the `expiry_date`

ckanext/unhcr/actions.py
```diff
@@ -925,16 +928,25 @@ def access_request_update(context, data_dict):
...
         if status == 'approved':
             # Notify the user
-            subj = mailer.compose_account_approved_email_subj()
-            body = mailer.compose_account_approved_email_body(user)
+            if data.get('user_request_type', USER_REQUEST_TYPE_NEW) == USER_REQUEST_TYPE_RENEWAL:
+                subj = mailer.compose_account_renewed_email_subj()
+                body = mailer.compose_account_renewed_email_body(user)
+            else:
+                subj = mailer.compose_account_approved_email_subj()
+                body = mailer.compose_account_approved_email_body(user)
```
avdata99, line 949: Notify user about the change in their account

ckanext/unhcr/actions.py
```diff
@@ -977,6 +989,17 @@ def access_request_create(context, data_dict):
     )
     data = data_dict.get('data', {})
 
+    # Users could ask for new account but also to renewal and expired account
+    if object_type == 'user':
+        user_request_type = data.get('user_request_type')
```
avdata99, line 994: The access request for the user can refer to a new user being accepted or a renewal of their account

ckanext/unhcr/actions.py
```diff
@@ -1041,6 +1065,16 @@ def external_user_update_state(context, data_dict):
...
     if not user_obj:
         raise toolkit.ObjectNotFound("User not found")
+
+    if state == m.State.ACTIVE:
+
+        if renew_expiry_date:
+            plugin_extras = _init_plugin_extras(user_obj.plugin_extras)
+            days = toolkit.config.get('ckanext.unhcr.external_accounts_expiry_delta', 180)
+            new_expiry_date = datetime.date.today() + datetime.timedelta(days)
+            plugin_extras['unhcr']['expiry_date'] = new_expiry_date.isoformat()
```
avdata99, line 1075: If an access request is approved and is about a renewal then the `expiry_date` should be updated.

ckanext/unhcr/actions.py
```diff
@@ -1227,13 +1329,17 @@ def user_create(up_func, context, data_dict):
...
+    if not data_dict.get('expiry_date'):
+        expiry_date = datetime.date.today() + datetime.timedelta(
+            days=toolkit.config.get(
+                'ckanext.unhcr.external_accounts_expiry_delta',
+                180  # six months-ish
+            )
         )
-    )
+    else:
+        expiry_date = data_dict.get('expiry_date')
```
avdata99, line 1340: This allow users to be created whit an specific `expiry_date`. It is used for testing purposes.

ckanext/unhcr/blueprints/user.py
```diff
@@ -118,7 +119,10 @@ def post(self):
                 'object_type': 'user',
                 'message': data_dict['message'],
                 'role': 'member',
-                'data': {'default_containers': containers}
+                'data': {
+                    'default_containers': containers,
+                    'user_request_type': USER_REQUEST_TYPE_NEW
```
avdata99, line 124: Every new access requests for users preserves the default type as new user asking for permission to log in.

ckanext/unhcr/mailer.py
```diff
@@ -350,6 +350,19 @@ def compose_account_approved_email_body(recipient):
     return render_jinja2('emails/user/account_approved.html', context)
 
 
+def compose_account_renewed_email_subj():
+    return '[UNHCR RIDL] - User account renewed'
```
avdata99, line 354: A different notification email is sent for renewed users

ckanext/unhcr/plugin.py
```diff
@@ -113,6 +113,7 @@ class UnhcrPlugin(
     plugins.implements(plugins.IValidators)
     plugins.implements(plugins.IPermissionLabels)
     plugins.implements(plugins.IBlueprint)
+    plugins.implements(plugins.IAuthenticator, inherit=True)
```
avdata99, line 116: Pending to define how expired/inactive users should be redirected to register again if try to access after being expired

ckanext/unhcr/click_commands.py
chris48s: Reading over the PR, I think this question is outdated now, but this is a multi-stage process.
When we create the `AccessRequest` object, the account is yet to expire. It is "One month before account expiry".
If the request is rejected, we delete the account.
If we reach the expiry date and the request has still not been approved (extended), we delete the account.
In the gap while the account is not yet expired but the request has not been approved or rejected, we leave the account active.

ckanext/unhcr/actions.py
chris48s: I don't think we need a confirmation email to the external user to tell them about this. The concept of account expiry isn't really exposed to external users.

ckanext/unhcr/actions.py
```diff
@@ -1197,6 +1231,74 @@ def user_list(up_func, context, data_dict):
     return users
 
 
+def expired_users_list(before_expire_days=None, include_activities=False):
```
chris48s, line 1234: `expired_users_list` and `request_renewal` are not actions. We're not invoking them with `toolkit.get_action()` and they don't have the signature `my_action(context, data_dict)` so they don't really fit in `actions.py`. I don't think these are really actions we want to expose via the API, so I think it makes more sense to think of these more as just utility functions used by one management command.  If we don't want to load up `click_commands.py` with helpers, you could turn it into a `/commands` module.

ckanext/unhcr/actions.py
```diff
@@ -1197,6 +1231,74 @@ def user_list(up_func, context, data_dict):
...
+    """ Get as list of expired (or about to expire) users and their activities """
+    ctx = {'ignore_auth': True}
+    users = toolkit.get_action('user_list')(ctx, {})
+    expired_users = []
+
+    # check expiry_date
+    for user in users:
+        full_user = toolkit.get_action('user_show')(ctx,  {'id': user['id']})
+        if full_user['expiry_date']:
```
chris48s, line 1243: Can we check the `'external'` flag here. We _shouldn't_ have internal users who have an `expiry_date` set but we never want to expire an internal user.

ckanext/unhcr/click_commands.py
```diff
@@ -88,3 +89,58 @@ def send_summary_emails(verbose):
...
+
+        if last_activity.get('activity_type', 'new user') == 'new user':
+            if verbose:
+                print(' - No relevant activities: {} ignored'.format(about_to_expire_user['name']))
+            ignored += 1
+            continue
+        renewal += 1
+
+        request_renewal(about_to_expire_user, last_activity)
+        print(' - Renewal access requested for {}'.format(about_to_expire_user['name']))
```
chris48s, line 129: This is going to say we requested renewal even if a request already exists

Shall we split the logic for
- do we need to create a request: yes/no
- create the request

into 2 functions. Then we can show a better message here

ckanext/unhcr/actions.py
avdata99: Removed [here](https://github.com/okfn/ckanext-unhcr/pull/604/commits/c78e11d019dae20cf9150dc2d6fa4e55de2f328a) @chris48s

ckanext/unhcr/actions.py
avdata99: Right. Moved [here](https://github.com/okfn/ckanext-unhcr/pull/604/commits/e620729de2b71358bf8137d41175c4aab8856f39)

ckanext/unhcr/click_commands.py
avdata99: I make [some changes](https://github.com/okfn/ckanext-unhcr/pull/604/commits/7aa1b97ce87b441af32e6dde7b47c4738cd4f54a) that improves the function and the message. I think is simpler than two functions. Let me know is this OK @chris48s.

ckanext/unhcr/actions.py
avdata99: Right. Fixed [here](https://github.com/okfn/ckanext-unhcr/pull/604/commits/efa55dce8818eac626eeb237efa88af2c2a0b6cd)

ckanext/unhcr/commands.py
```diff
@@ -0,0 +1,88 @@
...
+        Returns tuple:
+         - created: bool. Was the request created?
+         - reason: None or string. If created=False is the reason why the request was not created
+    """
+    # check if the request was already exists
+    previous_requests = model.Session.query(AccessRequest).filter(
+        AccessRequest.object_id == user_dict['id'],
+        AccessRequest.object_type == 'user',
+        AccessRequest.status == 'requested'
+    ).all()
```
chris48s, line 60: Someone needs to get an email notification about this AccessRequest so they can approve/reject the request. I have now lost access the the requirements document so I can't look it up but we need to identify the relevant user who should approve/reject it, send them an email and make sure that user is authorized to approve/reject the request.

ckanext/unhcr/commands.py
```diff
@@ -0,0 +1,88 @@
...
+
+    data = {
+        'user_name': user_dict['name'],
+        'last_activity': last_activity.get('activity_type', 'No activities found'),
+        'last_date': last_activity.get('timestamp', 'No date found'),
+        'expire_date': user_dict['expiry_date']
+    }
+
+    message = 'User %(user_name)s will expire on %(expire_date)s. Should this user be renewed? ' \
+              'Last activity: "%(last_activity)s" registered on %(last_date)s' % data
```
chris48s, line 73: if we're going to use the message field here can we format the timestamps in a human-readable way "registered on 2021-06-07T08:53:54.986580" is not that helpful

chris48s:
Some comments that don't go on the diff

1. On https://ckan-dev:5000/dashboard/requests renewal requests are shown as "External user Joe Bloggs requested access to deposit datasets" which isn't really right. In ckanext/unhcr/templates/user/snippets/access_request_list.html we should define a proper message for account renewal based on `user_request_type`.

2.  The `get_access_request_for_user()` helper throws on user detail page if there is a renewal request
That helper needs to filter on `user_request_type == USER_REQUEST_TYPE_NEW` now because there is more than one type of 'user' request.

3.  Can you review/update the logic in the `access_request_update` auth function please. I think at the moment nobody can approve these requests (except sysadmins because they skip auth checks).

4. Can you review/update the logic in the `access_request_list_for_user` action to ensure the right users can view the renewal requests that they can approve. Again, I've lost access to the requirements doc so I'm not 100% sure what the logic should be but if we need to ask questions to work out who should be able to see them, we can.

ckanext/unhcr/commands.py
avdata99: I propose notify curators and admins [here](https://github.com/okfn/ckanext-unhcr/pull/604/commits/277b636b1ffa84c3c307dc51036b2775e56b9d9d)
I will prepare some test for that

ckanext/unhcr/commands.py
avdata99: Fixed

avdata99:
> 1. On https://ckan-dev:5000/dashboard/requests renewal requests are shown as "External user Joe Bloggs requested access to deposit datasets" which isn't really right. In ckanext/unhcr/templates/user/snippets/access_request_list.html we should define a proper message for account renewal based on user_request_type.

@chris48s I make [these changes](https://github.com/okfn/ckanext-unhcr/commit/cc75aacedf9c95a7c44f13b8b1d63c619bfc4bf4) and then fake an about to expire user (also a new user to ckeck):

Selection_251

avdata99:
> The get_access_request_for_user() helper throws on user detail page if there is a renewal request
That helper needs to filter on user_request_type == USER_REQUEST_TYPE_NEW now because there is more than one type of 'user' request.

Updated [here](https://github.com/okfn/ckanext-unhcr/commit/1c15303984571f87e9af69bc78cbc64278d56a52) @chris48s

avdata99:
> 3. Can you review/update the logic in the access_request_update auth function please. I think at the moment nobody can approve these requests (except sysadmins because they skip auth checks).

@chris48s [this change](https://github.com/okfn/ckanext-unhcr/commit/d027b61b2b8ff4eae81bd91aa56b4257c9a570be) is related
I will create test for all these changes to validate.

avdata99:
> 4. Can you review/update the logic in the access_request_list_for_user action to ensure the right users can view the renewal requests that they can approve. Again, I've lost access to the requirements doc so I'm not 100% sure what the logic should be but if we need to ask questions to work out who should be able to see them, we can.

@chris48s, [here](https://github.com/okfn/ckanext-unhcr/pull/604/commits/b67a0b323159dbbb969fb2097a340ec083f783bb) I add curators (editors) to the access request list as the document says:

_Curation Team members and system administrators should have access to accounts about to expire and be able to perform the extension, but do not receive notification emails. _

ckanext/unhcr/actions.py
```diff
@@ -834,9 +841,11 @@ def access_request_list_for_user(context, data_dict):
     if user.sysadmin:
         return [dictize_access_request(req) for req in m.Session.execute(sql).fetchall()]
 
+    # create_dataset: We need a permission for admin and editors because both roles should have access to Access Requests
     organizations = toolkit.get_action("organization_list_for_user")(
-        context, {"id": user_id, "permission": "admin"}
+        context, {"id": user_id, "permission": "create_dataset"}
     )
+
```
chris48s, line 848: OK so what this code does is just shows all container editors all access requests (including requests for access to packages, requests for organisation membership and user account approval requests). This isn't what we're trying to achieve. It also doesn't solve the problem we are trying to solve because the check on `user` access requests is looking at the `data["default_containers"]` but a request where `data["user_request_type"] == USER_REQUEST_TYPE_RENEWAL` won't have a `default_containers` key in the data object.
https://github.com/okfn/ckanext-unhcr/pull/604/files#diff-9875b8b8f05c1e4eb2097f3c650144bfbe24b6f37e9cf5ab5f106340a3a80b00R81-R91

I think one of the issues here is we are trying do quite an obscure thing - we want to determine who is allowed to do something based whether that user has performed an activity in the past rather than whether they have a role or permission, so here is my suggeestion:

When we call `access_request_create` in our management command, we run the logic to work out who performed the last relevant `changed curation state` activity for this user (if applicable) and store a list of `user_id`s in the `data` key on the `AccessRequest` object. If there is no relevant `changed curation state` activity, then its an empty list. Lets call it `users_who_can_approve` (naming things is hard). If it is an empty list, it doesn't matter: sysadmins can still see and approve all access requests.

Then we can work this out once when we create the request and use `data["users_who_can_approve"]` in multiple places. The code here can then look something like:

```py
organizations = toolkit.get_action("organization_list_for_user")(
    context, {"id": user_id, "permission": "admin"}
)

containers = [o["id"] for o in organizations]
if not containers:
    return []

sql = sql.where(
    or_(
        and_(
            access_requests_table.c.object_type == "package",
            package_table.c.owner_org.in_(containers),
        ),
        and_(
            access_requests_table.c.object_type == "organization",
            access_requests_table.c.object_id.in_(containers),
        ),
        and_(
            access_requests_table.c.object_type == "user",
            access_requests_table.c.data["user_request_type"] == USER_REQUEST_TYPE_NEW,
            access_requests_table.c.data["default_containers"].has_any(array(containers)),
        ),
        and_(
            access_requests_table.c.object_type == "user",
            access_requests_table.c.data["user_request_type"] == USER_REQUEST_TYPE_RENEWAL,
            access_requests_table.c.data["users_who_can_approve"].contains(user_id),
            # ^^ I have not tested this line - check `.contains()` is defintely the right syntax
        ),
    )
)
```

and we can use this `users_who_can_approve` key in the auth function and mailer too.

ckanext/unhcr/auth.py
```diff
@@ -392,6 +402,21 @@ def external_user_update_state(context, data_dict):
 
     if not target_userobj.external:
         return {'success': False, 'msg': "Can only perform this action on an external user"}
+
+    if renew_expiry_date:
+        # Just a renewal, editor and admins are allowed
+        # create_dataset: We need a permission for admin and editors because both roles should have access to this request
+        user = toolkit.get_action('user_show')(context, {'id': target_user_id})
+        containers = user['default_containers']
```
chris48s, line 410: Similar comment to the one on `access_request_list_for_user` When you are calling `access_request_create` we are not populating the `data` key with a `default_containers` property (indeed it would make no sense to do this), so we can't rely on this property here. Lets use the same solution I've suggested above of:
- work out who can approve the request at the point we create it
- embed the user id in the `data` object on the `AccessRequest`
- when we get here, check if the user trying to change the state is in `users_who_can_approve`

ckanext/unhcr/mailer.py
```diff
@@ -350,6 +350,27 @@ def compose_account_approved_email_body(recipient):
...
+
+    user = model.User.get(user_id)
+    subj = 'Renewal request for {}'.format(user.name)
+    context = {
+        'requesting_user': user,
+        'message': message,
+        'h': toolkit.h
+    }
+
+    recipient_ids = helpers.get_user_admins(user_id) + helpers.get_user_curators(user_id)
```
chris48s, line 365: Again, lets work out who we're going to allow to approve the request once in the click command based on `changed curation state` activity, store it in `data["users_who_can_approve"]` and then here just send to those user IDs. If the list is empty, we can fall back to sysadmins.

ckanext/unhcr/auth.py
```diff
@@ -369,14 +371,22 @@ def access_request_update(context, data_dict):
             )
         }
     elif request.object_type == 'user':
-        return external_user_update_state(context, {'id': request.object_id})
+        from ckanext.unhcr.actions import USER_REQUEST_TYPE_NEW, USER_REQUEST_TYPE_RENEWAL
```
chris48s, line 374: What's with importing these constants inline in functions instead of at the top of the files? There are quite a lot of places we are doing this.
Also, I think we can move these consts to `models.py`. They don't really belong in `actions.py`

ckanext/unhcr/actions.py
avdata99: I implemented `users_who_can_approve` [here](https://github.com/okfn/ckanext-unhcr/pull/604/commits/b564cf0b4443f17674d23be24f9475717060b1fa) @chris48s

ckanext/unhcr/auth.py
avdata99: Implemented [here]()https://github.com/okfn/ckanext-unhcr/pull/604/commits/5f54b91101d1e917e62374241900108573657a43) @chris48s

ckanext/unhcr/auth.py
avdata99: Relocated [here](https://github.com/okfn/ckanext-unhcr/pull/604/commits/fb2a3fe24806f4a06f02cc9e1d529d3ef59808cb)

ckanext/unhcr/mailer.py
avdata99: Updated [here](https://github.com/okfn/ckanext-unhcr/pull/604/commits/c504d8a87efcbd5cea27efedd1de5cb7e221d229) @chris48s

ckanext/unhcr/actions.py
```diff
@@ -853,7 +859,17 @@ def access_request_list_for_user(context, data_dict):
...
+                    not_(access_requests_table.c.data.has_key("user_request_type")),
+                    access_requests_table.c.data["user_request_type"].astext == USER_REQUEST_TYPE_NEW,
+                ),
                 access_requests_table.c.data["default_containers"].has_any(array(containers)),
+            ),
+            and_(
+                access_requests_table.c.object_type == "user",
+                access_requests_table.c.data["user_request_type"].astext == USER_REQUEST_TYPE_RENEWAL,
+                access_requests_table.c.data.has_key("users_who_can_approve"),
+                access_requests_table.c.data["users_who_can_approve"].contains([user_id]),
```
chris48s, line 872:
```suggestion
                access_requests_table.c.data["users_who_can_approve"].contains([user.id]),
```

This wasn't working when I tested locally because we invoke `access_request_list_for_user` with `'user': toolkit.c.user` which makes `user_id` the user _name_, not the id. Switching to `user.id` fixes it.

ckanext/unhcr/auth.py
```diff
@@ -335,10 +335,12 @@ def package_collaborator_create(next_auth, context, data_dict):
 
 def access_request_list_for_user(context, data_dict):
     user = context.get('user')
+    # create_dataset: We need a permission for admin and editors because both roles should have access to Access Requests
     orgs = toolkit.get_action("organization_list_for_user")(
         {"user": user},
-        {"id": user, "permission": "admin"}
+        {"id": user, "permission": "create_dataset"}
```
chris48s, line 341: This should be `{"id": user, "permission": "admin"}` (to match the corresponding action) and the comment can be deleted

ckanext/unhcr/actions.py
```diff
@@ -834,9 +838,11 @@ def access_request_list_for_user(context, data_dict):
     if user.sysadmin:
         return [dictize_access_request(req) for req in m.Session.execute(sql).fetchall()]
 
+    # create_dataset: We need a permission for admin and editors because both roles should have access to Access Requests
     organizations = toolkit.get_action("organization_list_for_user")(
         context, {"id": user_id, "permission": "admin"}
     )
```
chris48s, line 844: This comment no longer makes sense

ckanext/unhcr/click_commands.py
```diff
@@ -5,7 +5,8 @@
 from ckan.plugins import toolkit
 import ckan.model as model
 
-from ckanext.unhcr.models import create_tables, TimeSeriesMetric
+from ckanext.unhcr.commands import expired_users_list, request_renewal
+from ckanext.unhcr.models import AccessRequest, create_tables, TimeSeriesMetric
```
chris48s, line 9:
```suggestion
from ckanext.unhcr.models import create_tables, TimeSeriesMetric
```

`AccessRequest` is an unused import here

ckanext/unhcr/click_commands.py
```diff
@@ -88,3 +89,62 @@ def send_summary_emails(verbose):
...
+        if created:
+            print(' - Renewal access requested for {}'.format(about_to_expire_user['name']))
+            renewal += 1
+        else:
+            print(' - Renewal access not created for user {}: {}'.format(about_to_expire_user['name'], reason))
+            ignored += 1
+
+    expired_users = expired_users_list()
+    if len(expired_users) == 0:
+        print('There is no expired users')
```
chris48s, line 137:
```suggestion
        print('There are no expired users')
```

ckanext/unhcr/click_commands.py
```diff
@@ -88,3 +89,62 @@ def send_summary_emails(verbose):
...
+           - notify admins about the users about to expire to allow to renewal those accounts """
+
+    ignored = 0  # User is about to expire but doesn't have any relevant activity or already requested
+    renewal = 0  # User is about to expire and have some relevant activity. Renewal is requested
+    deleted = 0  # User expired and was inactivated/deleted
+
+    before_expire_days = toolkit.config.get('ckanext.unhcr.external_accounts_notify_delta', 30)
+    about_to_expire_users = expired_users_list(before_expire_days=before_expire_days, include_activities=True)
+    if len(about_to_expire_users) == 0:
+        print('There is no users about to expire')
```
chris48s, line 111:
```suggestion
        print('There are no users about to expire')
```

ckanext/unhcr/templates/emails/access_requests/access_renewal_request.html
```diff
@@ -0,0 +1,33 @@
...
+{% block email_body %}
+
+  {% call email.paragraph() %}
+    Dear <b>{{ recipient.fullname or recipient.name }}</b>,
+  {% endcall %}
+
+  {% call email.paragraph() %}
+    The user 
+    {{ h.link_to(requesting_user.fullname or requesting_user.name, h.url_for('user.read', id=requesting_user.id, qualified=True)) }}
+    is about to expire and it is necessary to define if his account should be renewed.
```
chris48s, line 13:
```suggestion
    is about to expire and it is necessary to decide if their account should be renewed.
```

chris48s (commented):
I have left a few inline comments but we're down to very minor stuff now. All the core functionality is now working well.

One final thing that doesn't fit on the diff:
While I was testing, I found that if you reject the renewal, the user gets an email saying

> Dear Joe Bloggs,
>
> Your request for a RIDL user account has been rejected.
> ...

which isn't really right.

I think we need to either send a slightly different email or just not send it at all here:
https://github.com/okfn/ckanext-unhcr/blob/d4b9d9aad31bcd757c0b9e297a936d96e84be3c0/ckanext/unhcr/blueprints/access_requests.py#L53
if the request type is `USER_REQUEST_TYPE_RENEWAL`

We might need to confirm with the client what the call-to-action is for that notification so maybe to start with we just disable it.
Can you have a look at that as well as the inline comments please.

ckanext/unhcr/auth.py
avdata99: Fixed [here](https://github.com/okfn/ckanext-unhcr/pull/604/commits/f53d2e4c37a59b4342b484c8d42288d7f61dcbb9)

ckanext/unhcr/actions.py
avdata99: Fixed [here](https://github.com/okfn/ckanext-unhcr/pull/604/commits/f53d2e4c37a59b4342b484c8d42288d7f61dcbb9)

ckanext/unhcr/mailer.py
```diff
@@ -373,6 +376,22 @@ def notify_renewal_request(user_id, message, recipient_ids):
...
 
+def notify_rejection(request, message):
+    recipient = model.User.get(request['user_id'])
+    obj = toolkit.get_action('{}_show'.format(request['object_type']))(
+        {'user': toolkit.c.user}, {'id': request['object_id']}
+    )
+    if request['object_type'] == 'user':
+        if request['data'].get('user_request_type') == USER_REQUEST_TYPE_RENEWAL:
+            # we don't notify renewal rejections
+            return
```
avdata99, line 387: Same code than before, just moved to `mailer.py` and added this 3 lines

avdata99:
> I have left a few inline comments but we're down to very minor stuff now. All the core functionality is now working well.
> 
> One final thing that doesn't fit on the diff:
> While I was testing, I found that if you reject the renewal, the user gets an email saying
> 
> > Dear Joe Bloggs,
> > Your request for a RIDL user account has been rejected.
> > ...
> 
> which isn't really right.
> 
> I think we need to either send a slightly different email or just not send it at all here:
> https://github.com/okfn/ckanext-unhcr/blob/d4b9d9aad31bcd757c0b9e297a936d96e84be3c0/ckanext/unhcr/blueprints/access_requests.py#L53
> 
> 
> if the request type is `USER_REQUEST_TYPE_RENEWAL`
> We might need to confirm with the client what the call-to-action is for that notification so maybe to start with we just disable it.
> Can you have a look at that as well as the inline comments please.

Fixed [here](https://github.com/okfn/ckanext-unhcr/pull/604/commits/5858f5dbd6d49118500cf3b371b1e6e7a7dd5eed#diff-1cabe4cb9bf3ecd49db870fb2eaaa5f4138b611c4114351460637d4202e798c3R385-R387)

chris48s:
Nice :+1:  One failing test left to sort out. I think you've still got a test which is checking the old auth functionality (before f53d2e4 was applied) which needs updating then I think this is good to go.

avdata99:
> Nice  One failing test left to sort out. I think you've still got a test which is checking the old auth functionality (before [f53d2e4](https://github.com/okfn/ckanext-unhcr/commit/f53d2e4c37a59b4342b484c8d42288d7f61dcbb9) was applied) which needs updating then I think this is good to go.

This was fixed and the branch was rebased to master with py3 @chris48s

chris48s (approved):
Cool. Thanks.
I think this PR is finished but we should hold off on merging to master until we want to make a release and deploy it.
If we pick up any problems with the python3 migration that need to be fixed we will want to apply those fixes without deploying this yet.
We'll also want to coordinate the merge with https://github.com/okfn/docker-ckan-unhcr-aws/pull/43

PR #610, opened 2021-06-16
run the tests on python3

chris48s:
Woo. We just got the tests green. Lets break them again :hammer: 
This PR sets up a second CI job that runs the tests in a CKAN 2.9/py3 docker environment. I've got it as far as setting up all the plugins then failing to set up the DB because of a python2-only import error.
If we merge this, that will allow us to pick away at moving from a python2 only codebase to a python2/3 codebase with six, future, etc until both CI builds pass.
Probably the next job is to get it to a point where the "Initialize DB" step will run and `pytest` can actually run (and fail a whole bunch of tests). I suspect given that running the DB init script requires loading `plugin.py` (which imports _all the things_ ) that might actually be quite big. I'm not sure if it makes most sense to do that first before merging this so we can hopefully pick off smaller chunks on an ad-hoc basis.


amercader:
Can we set the py3 jobs to "ignore if failed" (I can't remember the Github Actions equivalent). I think it's important to not interfere with the "supported" tests while py3 support is not there yet and  we work on other stuff

amercader:
Looking at it it looks like it might not be supported, and it will be like this just for a while so perhaps we just merge it

https://github.com/actions/toolkit/issues/399

PR #613, opened 2021-06-17
558 chained actions

avdata99:
Continue from #574
Related to #558 
Replace for #601
 
Finish migration for top chained actions and auths

### DIFF
--- a/ckanext/unhcr/auth.py
+++ b/ckanext/unhcr/auth.py
@@ -348,6 +348,8 @@ def access_request_update(context, data_dict):
     request = model.Session.query(AccessRequest).get(request_id)
     if not request:
         raise toolkit.ObjectNotFound("Access Request not found")
+    if request.object_type not in ['organization', 'package', 'user']:
+        raise toolkit.Invalid("Unknown Object Type")
 
     if request.object_type == 'package':
         package = toolkit.get_action('package_show')(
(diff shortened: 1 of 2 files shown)


ckanext/unhcr/auth.py
```diff
@@ -370,8 +369,6 @@ def access_request_update(context, data_dict):
     elif request.object_type == 'user':
         return external_user_update_state(context, {'id': request.object_id})
 
-    raise toolkit.Invalid("Unknown Object Type")
```
chris48s, line 373: Everything else in this PR is fine, but I think we need to do _something_ here.

We sometimes invoke auth functions from places other than the correspondingly named action.
For example
- https://github.com/okfn/ckanext-unhcr/blob/53dbd720cd65c2cf25adc157b2eaa95b05227d31/ckanext/unhcr/auth.py#L308-L313
- https://github.com/okfn/ckanext-unhcr/blob/53dbd720cd65c2cf25adc157b2eaa95b05227d31/ckanext/unhcr/auth.py#L191-L196
- https://github.com/okfn/ckanext-unhcr/blob/53dbd720cd65c2cf25adc157b2eaa95b05227d31/ckanext/unhcr/auth.py#L235-L240
etc

If we completely remove the fallback case here and invoke this auth function somewhere else with an invalid `object_type` then this function will (implicitly) return `None`, which is not right. I don't think there is any issue with an auth function throwing `Invalid` tbh, but if you think it is wrong for the auth function to throw `Invalid` we should at least `return {'success': False}` in the fallback case.

avdata99: Ok, fixed [here](https://github.com/okfn/ckanext-unhcr/pull/613/commits/fd799f0334e30ca36598e29f33389355821f923d)

chris48s:
Does this mean we've now fully addressed #558 or is there still more to do?
If we're done with #558 lets close it..

avdata99:
We are done @chris48s. I'm closing #558

PR #616, opened 2021-06-21
WIP: fix linked dataset backlinks

### LINKED ISSUE #577: CKAN 2.9: Fix related datasets backlinks
Our old approach to this relied on revisions, a feature which has been removed in CKAN 2.9.

I tried a fairly naive approach to this in PR #550

```diff
diff --git a/ckanext/unhcr/jobs.py b/ckanext/unhcr/jobs.py
index 09add51..8af32ae 100644
--- a/ckanext/unhcr/jobs.py
+++ b/ckanext/unhcr/jobs.py
@@ -1,6 +1,8 @@
 import logging
 import time
 
+from sqlalchemy import desc
+
 from ckan import model
 from ckanext.unhcr import utils
 import ckan.plugins.toolkit as toolkit
@@ -34,15 +36,12 @@ def process_dataset_on_delete(package_id):
 
 
 def process_dataset_on_update(package_id):
-
     # Pause execution
     time.sleep(3)
 
-    # Process dataset_fields
-    _process_dataset_fields(package_id)
-
     # Prepare back references
-    link_package_ids = _get_link_package_ids_from_revisions(package_id)
+    # do this before _process_dataset_fields so we don't write another activity
+    link_package_ids = _get_link_package_ids_from_activities(package_id)
 
     # Create back references
     created_link_package_ids = set(link_package_ids['next']).difference(link_package_ids['prev'])
@@ -52,6 +51,9 @@ def process_dataset_on_update(package_id):
     removed_link_package_ids = set(link_package_ids['prev']).difference(link_package_ids['next'])
     _delete_link_package_back_references(package_id, removed_link_package_ids)
 
+    # Process dataset_fields
+    _process_dataset_fields(package_id)
+
 
 # Internal
 
@@ -160,5 +162,32 @@ def _delete_link_package_back_references(package_id, link_package_ids):
             )
 
 
-def _get_link_package_ids_from_revisions(package_id):
-    return {'prev': [], 'next': []}
+def _get_link_package_ids_from_activities(package_id):
+    activities = model.Session.query(
+        model.Activity
+    ).filter(
+        model.Activity.activity_type == 'changed package',
+        model.Activity.object_id == package_id,
+    ).order_by(
+        desc(model.Activity.timestamp)
+    ).limit(
+        2
+    ).all()
+
+    prev = []
+    if len(activities) > 1:
+        if activities[1].data.get('package'):
+            if activities[1].data['package'].get('linked_datasets'):
+                prev = utils.normalize_list(
+                    activities[1].data['package']['linked_datasets']
+                )
+
+    next_ = []
+    if len(activities) > 0:
+        if activities[0].data.get('package'):
+            if activities[0].data['package'].get('linked_datasets'):
+                next_ = utils.normalize_list(
+                    activities[0].data['package']['linked_datasets']
+                )
+
+    return {'prev': prev, 'next': next_}

```

The issue with this approach is that depending on the operations we are performing, it is not always a valid assumption that the last two 'changed package' activities are the ones to look at. This also creates some race conditions (e.g: if the datapusher service performs a `package_update` before the background job runs, etc etc).

Having discussed a few different methods, I think we've decided that the best thing to do is:

- We continue to store a `linked_datasets` field on each package
- We create a new link table which stores pairs of dataset IDs which are related to each other `(source, dest)`. Links worth both ways so a link between ds1 and ds2 can be stored either as
  - `(ds1, ds2)`
  - `(ds2, ds1)`
  and duplicates are OK
- In the `process_dataset_on_update` hook, we just take the contents of the `linked_datasets` field and write it to the dataset link table as `(source, dest)`. This way we don't have to care whether the latest update adds removes or doesn't change, as long as jobs execute in order and the end result is the entries for source match `linked_datasets`, it will be right
- We implement a `before_index()` hook which queries the link table `SELECT * FROM dataset_link WHERE source=%s or dest=%s`, normalises the order, de-dupes and writes a flat array back to the `linked_datasets` field on the dataset
- Write a data migration to retrospectively populate this table for existing datasets

chris48s:
Closes  #577

Gah. This got more complicated than I wanted it to :(

I haven't written any tests for this yet as I think I want to be confident that this is the right approach before I formalise this with tests, but if we're happy with this I will write a bunch of tests for this code. It is a sufficiently tangled web that I think it definitely needs tests but I don't want to lock them in until we're settled on the approach.

Some design decisions:

- I was hitting all sorts of problems trying to do this in a background job. Every time a whack-a-mole-d one problem another one sprung up. I know we were doing this in a background job for performance reasons but I think moving this operation inline has acceptable performance characteristics and solves a lot of other performance issues. I will comment on this more on the diff.
- I've decided to leverage the built-in concept for `package_relationships`. In some cases, I've used the core actions. In some cases I've used SQL. I'll leave more comments about this on the diff.

@amercader  - do you want to have a quick look over this WIP PR and make sure you're happy with the approach. If so I will:
- Add tests for all this stuff
- Write a migration to convert existing `linked_datasets` to `package_relationship`s


ckanext/unhcr/plugin.py
```diff
@@ -324,8 +326,60 @@ def before_index(self, pkg_dict):
...
+        if linked_datasets is None:
+            pkg_dict.pop('linked_datasets', None)
+        else:
+            pkg_dict['linked_datasets'] = linked_datasets
+
         return pkg_dict
 
+    def before_view(self, data_dict):
+        if 'owner_org' not in data_dict:
+            return data_dict
```
chris48s, line 339: I'm trying to duck-type `data_dict` here because lots of plugin interfaces have a before_view hook:
- `ITagController`
- `IGroupController`
- `IOrganizationController`
- `IPackageController`

ckanext/unhcr/plugin.py
```diff
@@ -324,8 +326,60 @@ def before_index(self, pkg_dict):
...
 
+        linked_datasets = self._get_linked_datasets(pkg_dict['id'])
+        if linked_datasets is None:
+            pkg_dict.pop('linked_datasets', None)
+        else:
+            pkg_dict['linked_datasets'] = linked_datasets
+
         return pkg_dict
 
+    def before_view(self, data_dict):
```
chris48s, line 337: I've implemented `before_search` and `before_view` here but there is no `IPackageController.before_show` hook for packages (only `IResourceController` has this) so API responses will only return what is saved in `linked_datasets` directly on the package object. It does work bi-directionally through the front-end though.

ckanext/unhcr/plugin.py
```diff
@@ -324,8 +326,60 @@ def before_index(self, pkg_dict):
...
+                    if row.object_package_id != pkg_id
+                ]
+            )
+        )
+
+        if len(linked_datasets) == 0:
+            return None
+        if len(linked_datasets) == 1:
+            return linked_datasets[0]
+        return '{' + ','.join(linked_datasets) + '}'
```
chris48s, line 381: I've manually implemented this serialization format here. It is the inverse of

https://github.com/okfn/ckanext-unhcr/blob/f62d710bc2326d764179eb5da27f21e38d0b2da0/ckanext/unhcr/utils.py#L16-L22

but I'm not really 100% sure how data gets into this format. If I do

```py
from ckan import model
rec = model.PackageExtra(
    package_id=data_dict['id'],
    key='linked_datasets',
    value=['34cb35d1-dd60-4cf5-b948-b4bbd4627eaa','ad22d394-6389-4e7a-821b-ddb9c207c6a1'],
    state='active',
)
model.Session.add(rec)
model.Session.commit()
```

then I end up with `{34cb35d1-dd60-4cf5-b948-b4bbd4627eaa,ad22d394-6389-4e7a-821b-ddb9c207c6a1}` in my DB so I guess maybe it is an internal behaviour of SQLAlchemy or Postgres? Let me know if there is a better way of doing this.

This can't just be a list because Solr throws
`multiple values encountered for non multiValued field linked_datasets`

ckanext/unhcr/actions.py
```diff
@@ -37,6 +37,100 @@ def _get_user_obj(context):
...
+    ).filter(
+        m.PackageRelationship.subject_package_id==package_id,
+    ).update({
+        m.PackageRelationship.state: m.State.DELETED
+    })
+
+    default_user = toolkit.get_action('get_site_user')({ 'ignore_auth': True })
+    results = []
+    for id_ in linked_datasets:
+        result = toolkit.get_action('package_relationship_create')(
```
chris48s, line 69: This is potentially inefficient. If our dataset was related to a thousand other datasets or something this would give us a problem here, but I'm not worried about it for a couple of reasons.
- I had a look at the production database. Out of >500 datasets, the dataset with the largest number of `linked_datasets` is related to two other datasets so I think we're unlikely to encounter practical examples of datasets with piles of links.
- The way I'm using the action here (calling the actions with `defer_commit: True` and then committing the transaction later) minimises the number of round-trips to the database.

If we're super-worried about it or we hit performance issues in future, we could stop using `package_relationship_create` here and switch to something like this https://stackoverflow.com/questions/3659142/bulk-insert-with-sqlalchemy-orm to bulk-insert the records.

ckanext/unhcr/actions.py
```diff
@@ -37,6 +37,100 @@ def _get_user_obj(context):
...
+            {
+                'user': default_user['name'],
```
chris48s, line 71: I'm passing a user instead of using `ignore_auth: True` because it is required here https://github.com/ckan/ckan/blob/0ceebd17bcbd435b156e776df20593ed41af3ffd/ckan/logic/action/create.py#L518 but `user` is not used anywhere so I think it is probably a pointless requirement.

ckanext/unhcr/actions.py
```diff
@@ -37,6 +37,100 @@ def _get_user_obj(context):
...
+                'defer_commit': True,
+                'job': True,
+                'model': m,
+                'session': session,
+            },
+            {
+                'subject': package_id,
+                'object': id_,
+                'type': 'links_to',
+                'comment': '',
```
chris48s, line 81: I think if `package_relationship_create` accepted a `state` parameter, what I probably would have done here is pass `state: model.State.PENDING` if either dataset is still pending and then make the relationship active when both datasets are active. It doesn't though, so we just create them as active. This creates a situation where a draft dataset can be related to an active one. I think this probably existed using the old (revisions-based) code too though so I'm not going to get really bogged down in this.

Issue #612, opened 2021-06-17
migrate to python3


chris48s:
I have now merged #610 which runs the tests on both python2 and python3 in CI.
The python3 build is failing at this stage. Note Github Actions does not have the concept of "allowed failure" so we're back to a red build on master :( but we should be keeping the "CKAN 2.9-py2" build green at least
For the moment we are still deploying and developing on python2
The next stage is to get to a stage where we've got a python2/3 compatible codebase (using future/six/etc) and get the test suite passing on python3. Then we can start doing some manual tests.

Currently the build is failing at the "Initialize DB" step of the build - we can't even apply the DB migrations so step 1 is to get to a stage where that passes and we can actually invoke pytest to get some specific test failures. That might unavoidably end up being a fairly big PR as bootstrapping the plugin actually loads quite a lot of code but lets scope that as tightly as possible.

Once we've done that we can start tacking test failures in manageable chunks. Ideally what I'd like to do is attack this in small chunks. A series of small PRs like "get the access request controller tests passing under python 3" is going to be much easier to review and there will be less friction to getting stuff reviewed/merged quickly.

PR #617, opened 2021-06-22
python3, part 1

chris48s:
Refs #612
This is probably the next minimum useful unit of work toward python3. This gets us to a stage where the tests can run and we've got
121 failed, 198 passed
on the python3 build. After this we can start chipping away at those failure in smaller PRs


chris48s:
121 failed, 198 passed

bin/install-extensions-2.9.bash
```diff
@@ -7,7 +7,7 @@ pip install setuptools-rust
 git clone --depth 1 --branch release-2.1.0 https://github.com/ckan/ckanext-scheming
 (cd ckanext-scheming && python3 setup.py develop)
 
-git clone --depth 1 --branch 0.2.3 https://github.com/okfn/ckanext-hierarchy
+git clone --depth 1 --branch v0.2.4 https://github.com/okfn/ckanext-hierarchy
```
chris48s, line 10: I just went ahead and did this on the hierarchy repo.
https://github.com/okfn/ckanext-hierarchy/compare/okfn:0.2.3...v0.2.4
This was throwing in the `before_search` hook which was failing basically every test because it gets called by one of the factories.
Basically running on python3 switches us to a more recent version of werkzeug (via CKAN's `requirements.txt` rather than `requirements-py2.txt`) which rethrows this as `RuntimeError`  https://github.com/pallets/werkzeug/blob/1.0.0/src/werkzeug/local.py#L310

ckanext/unhcr/actions.py
```diff
@@ -1,9 +1,15 @@
+try:
```
amercader, line 1: So just to be clear on the strategy here, are you planning on merging this in master and then try to get the python3 tests green with PRs targeting master? ie essentially implementing support for both py2 and py3 in master

The issue with that is that you will end up with a lot of code like this where you need to maintain compatibility with both versions. Wouldn't it we cleaner to do an approach like the `ckan2.9` branch and work on the `py3` branch directly with   py3 only support (and in `docker-ckan-unhcr-aws`), and then we merge / deploy we feel we are ready? We will have to keep up to date with other features/fixes merged in master but as long as there are tests it should be easy to incorporate those.

What do you think?

chris48s: > So just to be clear on the strategy here, are you planning on merging this in master and then try to get the python3 tests green with PRs targeting master? ie essentially implementing support for both py2 and py3 in master

Yes exactly. I was thinking the way to do this would be to get the plugin to a stage where it will work on both python2/3 (with six/future/etc). Then once we've got it working on both runtimes we look at switching the deploy environment on to python3. Then once we've done that we can drop out all the python2/3 compatibility stuff. The advantage of that approach is it would allow us to pick at this gradually with one default branch while we continue to ship features from the roadmap.

That's how we dealt with the last big python2 --> 3 migration I worked on. That said, I will concede it was a much larger codebase. That project was much bigger and it probably took about 2 years from first starting to run tests on python3 to actually deploying on py3 in production (although this was long before py2 was EOL - there wasn't the same sense of urgency around it and we probably had to wait for packages we depended on and stuff). RIDL is a smaller project so maybe I'm envisaging a bigger job than it really is and we can get it done much quicker here?

> Wouldn't it we cleaner to do an approach like the ckan2.9 branch and work on the py3 branch directly with py3 only support

I think if we want to do more of a "big bang" like you suggest, we probably need to basically set a point where we block the other feature work on the python3 upgrade (I very consciously merged almost nothing to master while I was working on 2.9) and really go all-in the py3 branch (or start using that as the target for new work). I'm not sure where we find that point though. We still haven't got approval to ship 2.9 to prod (and not really much sign of that happening soon) and as soon as we ship that we've got more features that are pretty much ready to ship on py2.

I'm guessing that the second approach is how you've done things on GDX? My gut instinct is that migrating from python2 to 3 will be useful to break down into smaller managable chunks but maybe I'm over-estimating it. How long did it take you on GDX?

If we want to take that approach, we can revert https://github.com/okfn/ckanext-unhcr/commit/a6f14beeeb8336557623e7ed6f4cac19a66cb03e on master as there is not really any point having a failing python3 build on master if we don't plan to apply fixes there until we drop the py2 tests.

Issue #340, opened 2020-06-19
Offboarding users


chris48s:
Goal: make sure that when a user is deleted they no longer have permissions on the site and they are replaced if necessary.

Deleted users stay on the system with a “deleted” state to ensure no activity history is lost. All their memberships are removed though, including data containers they belong

Tasks could involve:
* When a user is deleted, check if they are the sole administrator for a Data Container, and if so trigger an action, eg assign a sysadmin, notify sysadmins etc

Right now users need to be manually deleted by sysadmins. We could investigate if there is a way to automate the deletion of users based on deletions in Azure AD, either:
* Configure somehow Azure AD to send a notification to RIDL (push), this will be the preferred way
* Periodically check the status of all RIDL users in Azure AD to see if some were deleted

chris48s:
Requirements on this one are now more clear.

- We will pull data from Azure AD. @amercader to follow up with request for API endpoint we can query
- We will have a daily job that runs using yacron to:
    - Loop over every user in the DB where state=active
    - Query Azure AD: is this user still active: Yes/No?

If the user is active, do nothing. Move on to the next one.
If Active Directory reports the user is inactive, the requirements are:

- Set the user's state=deleted
- Find any containers where the user is a member, editor or admin and remove the role
- If the user is/was the only admin of a container, add a sysadmin to that container as admin (every container should have at least one admin)
- Email sysadmins with a notification explaining
    - user A has left the org
    - they've been removed from container B
    - user C has been added as an admin to container B as a fallback position

While we don't yet have the API endpoint we need to query Azure AD, we could write the logic needed to do all those other jobs, so we could start working on this but just stub the call to AD for now i.e:

```py
def user_is_active(user):
    # TODO - call Azure AD
    return True/False
```

and develop the rest of it around that.

avdata99:
@amercader @chris48s 
I'm using the [Microsoft Graph code-try](https://docs.microsoft.com/en-us/graph/api/user-list?view=graph-rest-1.0&tabs=http#code-try-1) to check the Azure AD API

I can see thousands of users and groups.
The plan is to get the filtered list for users using RIDL CKAN

I check my own user to find out which groups he is a part of
https://graph.microsoft.com/v1.0/me/appRoleAssignments

I can see I'm in a lot of groups but maybe `RIDL-Dev1` is the one we want

I'm also see my `companyName` is DENCO. This is another possible filter
[...]

PR #620, opened 2021-06-28
WIP: Check Azure Active Directory users

avdata99:
Related to #340

In progress


ckanext/unhcr/actions.py
```diff
@@ -9,15 +9,15 @@
 from sqlalchemy.dialects.postgresql import array
 from sqlalchemy.orm import aliased
 from ckan import model
-from ckan.authz import has_user_permission_for_group_or_org
+from ckan.authz import has_user_permission_for_group_or_org, get_group_or_org_admin_ids
 from ckan.plugins import toolkit
 from ckan.lib import mailer as core_mailer
-from ckan.lib.mailer import MailerException
```
avdata99, line 15: Just making simpler the confusing multiples `mailers` imports

ckanext/unhcr/actions.py
```diff
@@ -270,8 +270,9 @@ def organization_member_create(up_func, context, data_dict):
         user = toolkit.get_action('user_show')(context, {'id': data_dict['username']})
 
         # Notify the user
+        extra_mail_msg = data_dict.get('extra_mail_msg')
```
avdata99, line 273: The fallback admin for an orphan container will receive an email and we need top clarify why is being set him/her as a new admin.

ckanext/unhcr/actions.py
```diff
@@ -1203,16 +1204,52 @@ def user_show(up_func, context, data_dict):
...
+    user['focal_point'] = unhcr_extras['focal_point']
+    user['expiry_date'] = unhcr_extras['expiry_date']
+    user['default_containers'] = unhcr_extras['default_containers']
 
-    user['focal_point'] = extras['focal_point']
-    user['expiry_date'] = extras['expiry_date']
-    user['default_containers'] = extras['default_containers']
+    if data_dict.get('include_plugin_extras', False):
+        plugin_extras.update({'unhcr': unhcr_extras})
+        user['plugin_extras'] = plugin_extras
```
avdata99, line 1216: Allow `include_plugin_extras` to work as expected and clarify `plugin_extras` (besides `unhcr` ones)

ckanext/unhcr/azure.py
```diff
@@ -0,0 +1,77 @@
...
+def internal_azure_users_list():
+    """ get a list of azure users"""
+    log.info('Getting Azure AD users')
+    azure_users = []
+    ctx = {'ignore_auth': True}
+    users = toolkit.get_action('user_list')(ctx, {})
+
+    for user in users:
+        userobj = model.User.get(user['id'])
+        if is_saml2_user(userobj):
```
avdata99, line 20: We already have a `is_saml2_user` function to filter the users list

avdata99, line 20: We already have a `is_saml2_user` function to filter the users list

ckanext/unhcr/azure.py
```diff
@@ -0,0 +1,77 @@
...
+
+    headers = {
+        'Authorization': 'Bearer {}'.format(azure_token)
+    }
+
+    # Get the full list of users.
+    # This URL could be improved with a filtered list with only RIDL CKAN users
+    azure_users_url = toolkit.config.get(
+        'ckanext.unhcr.azure_users_list_url',
+        'https://graph.microsoft.com/v1.0/users/?$top=999&$select=id,mail'
```
avdata99, line 46: Max page size is 999
Still pending to define a better endpoint for a only RIDL CKAN users. Maybe [using Azure AD groups](https://github.com/okfn/docker-ckan-unhcr-aws/pull/49/files#diff-5b57400c5ceb6cddcf7a66c676b048be688653c3762dd4b379a369caadd73b3fR29).

If not possible, this should work

ckanext/unhcr/jobs.py
```diff
@@ -146,3 +171,10 @@ def _delete_link_package_back_references(package_id, link_package_ids):
 def _get_link_package_ids_from_revisions(package_id):
     # TODO: fix backlinks https://github.com/okfn/ckanext-unhcr/issues/577
     return {'prev': [], 'next': []}
+
+
+def _define_new_admin_member(group_id):
+    """ A data-container require a fallback admin.
+        Pick one from sysadmins.
+        TODO: check if this group has a parent and define its admin """
+    return get_sysadmins()[0]
```
avdata99, line 180: Should we try to assign this orphan container to a closer admin?
Maybe an admin from parent container

ckanext/unhcr/jobs.py
```diff
@@ -146,3 +174,27 @@ def _delete_link_package_back_references(package_id, link_package_ids):
...
+                parent = toolkit.get_action('organization_show')(
+                    {'ignore_auth': True},
+                    {'id': parent['name']}
+                )
+            admins = get_group_or_org_admin_ids(parent['id'])
+            if len(admins) > 0:
+                admin = model.User.get(admins[0])
+                return admin
+
+    return get_sysadmins()[0]
```
chris48s, line 200: I've just realised looking at this that the `get_sysadmins()` helper should be filtering on `model.Member.state == 'active'` as well as `model.User.sysadmin==True`

chris48s, line 200: This probably raises some things to query with the client:
- Do we want to assign an admin from a parent container if available instead of a sysadmin? (basically the question you've asked in https://github.com/okfn/ckanext-unhcr/pull/620/#discussion_r661771200 )
- If we've got >1 possible new admins to pick from, should we just pick one (semi)-randomly (like you've done here with `[0]`) or all of them?

chris48s (commented):
I haven't really looked at any of the azure stuff in this PR but I've had a look over the logic in the `user_delete` action and `process_last_admin_on_delete` background job. It looks sensible. I've left a couple of comments, but lets not do loads more work on this one until we have clarified with the client how we are going to detect inactive container admins.

ckanext/unhcr/jobs.py
avdata99: Fixed and randomized [here](https://github.com/okfn/ckanext-unhcr/pull/620/commits/2be68fb2491b54453809e761b5f21ed7f2caa0a9)

avdata99: We can leave this random until the client define what we want here

chris48s: OK. We should still fix the original function though. It is used in other places
https://github.com/okfn/ckanext-unhcr/blob/0f9f025528215237ee805f722a80af43c6a5d977/ckanext/unhcr/mailer.py#L249
https://github.com/okfn/ckanext-unhcr/blob/f62d710bc2326d764179eb5da27f21e38d0b2da0/ckanext/unhcr/actions.py#L237
and we only want active there.

We could do that in its own PR tbh. Then we can just merge it rather than blocking the fix on this

avdata99: Updated [here](https://github.com/okfn/ckanext-unhcr/pull/620/commits/8c1ed30042b52ac1844bf1c18ccef56d6b0b071b) @chris48s

avdata99:
old

PR #621, opened 2021-06-30
python3

### LINKED ISSUE #612: migrate to python3

chris48s:
Closes #612
Refs https://github.com/okfn/docker-ckan-unhcr-aws/pull/48

- I rolled back the commit which run the python3 tests on master in https://github.com/okfn/ckanext-unhcr/commit/d4b9d9aad31bcd757c0b9e297a936d96e84be3c0 Merging this PR will drop python2 compatibility and testing completely

- This PR gets us to the stage where all the tests pass on python3 I also did a load of manual testing and couldn't find any further issues

- There are various other bits of python2 cruft which I have chosen not to remove yet. For example, we can change all the `import mock` statements to `from unittest import mock`. We can remove all the `# -*- coding: utf-8 -*-` statements. We can also drop any leading `u`s from strings (i.e: `u'things'` --> `'things'`) etc. I have decided not to do these yet. Since we may end up "maintaining" this diff alongside master for a while I want to keep the changeset as small as possible: minimum viable python3-compatibility, if you will :) We can do this kind of "cleanup" in a follow-on PR

### DIFF
--- a/bin/install-extensions.bash
+++ b/bin/install-extensions.bash
@@ -1,22 +1,17 @@
 #!/bin/bash
 set -euo pipefail
 
+pip install setuptools-rust
+
 
 git clone --depth 1 --branch release-2.1.0 https://github.com/ckan/ckanext-scheming
-(cd ckanext-scheming && python setup.py develop)
+(cd ckanext-scheming && python3 setup.py develop)
 
 git clone --depth 1 --branch 0.2.3 https://github.com/okfn/ckanext-hierarchy
-(cd ckanext-hierarchy && python setup.py develop && pip install -r requirements.txt)
+(cd ckanext-hierarchy && python3 setup.py develop && pip install -r requirements.txt)
 
 git clone --depth 1 --branch v1.0.0 https://github.com/keitaroinc/ckanext-s3filestore
-(cd ckanext-s3filestore && python setup.py develop && pip install -r requirements.txt)
+(cd ckanext-s3filestore && python3 setup.py develop && pip install -r requirements.txt)
 
 git clone --depth 1 --branch v1.1.1 https://github.com/keitaroinc/ckanext-saml2auth
-(
-    cd ckanext-saml2auth &&
-    # workaround for Python 2
-    # TODO: remove when we move to Py3
-    sed -i -e "s/install_requires=\['pysaml2>=6.3.0'\],//" setup.py &&
-    pip install pysaml2==4.9.0 python2-secrets==1.0.5 &&
-    python setup.py develop
-)
+(cd ckanext-saml2auth && python3 setup.py develop)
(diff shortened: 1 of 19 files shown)


bin/install-extensions.bash
```diff
@@ -1,22 +1,17 @@
 #!/bin/bash
 set -euo pipefail
 
+pip install setuptools-rust
```
amercader, line 4: Is this need to install the cryptography package?

ckanext/unhcr/helpers.py
```diff
@@ -1028,3 +1028,8 @@ def nl_to_br(text):
     result = u'\n\n'.join(u'<p>%s</p>' % p.replace('\n', Markup('<br>\n'))
                           for p in _paragraph_re.split(escape(text)))
     return Markup(result)
+
+
+def is_plugin_loaded(plugin):
```
amercader, line 1033: Use this instead, it is more reliable across all scenarios:

```python
import ckan.plugins as plugins

def is_plugin_loaded(plugin_name):
     return plugins.plugin_loaded(plugin_name)
```

amercader (approved):
Great stuff @chris48s!

bin/install-extensions.bash
chris48s: Yes exactly.
I've made the same change in the Dockerfiles in the deploy repo. The prod Dockerfile also needed `libffi-dev` for the cryptography install to work (the dev Dockerfile already has it)

chris48s:
Plan for deploying this:

- Merge this PR https://github.com/okfn/ckanext-unhcr/pull/621
- Write a changelog entry
- Bump https://github.com/okfn/ckanext-unhcr/blob/master/ckanext/unhcr/__init__.py#L1
- Push a tag to this repo
- Merge https://github.com/okfn/docker-ckan-unhcr-aws/pull/48
- Bump ckanext-unhcr tag in prod `Dockerfile` e.g: https://github.com/okfn/docker-ckan-unhcr-aws/commit/b3e180f5c7ba2e6536e4ef89928b23ae0b7df1ac
- Deploy

Issue #619, opened 2021-06-23
Allow user to tag datasets with geographic metadata


chris48s:
Once we have completed #618 we will also need the ability to tag datasets with geographies.
The specifics of this part of the job are currently less clear in my mind, but I'm zeroing in on them as I think about it..

Some requirements and assumptions we are already clear on:

- Geographic tags apply to datasets not resources
- Relationship is one-to-many: One dataset can be tagged with many geographies (including at the same level. One dataset can be apply to multiple countries, for example)
- Geography is not constrained by data container. Any dataset can be tagged with any geography. It is up to humans to make good decisions
- At any given time the user can tag datasets with whatever geographies are 'current' at the point of upload but as time passes we will end up with datasets that are associated with geographies which are no longer current
  **TODO**: more notes on how we deal with versioning/change over time TBC
- Datasets are related to units of geography by `pcode`

Probably the hard part of this is the interface we give users to pick geographies with. These lists have lots of things in them:

- Countries: ~200
- Populations of Concern: ~13,000
- Admin 1: ~3,000
- Admin 2: ~30,000
- Reference places: ~200,000

How do we allow users to meaningfully pick from a list of >10,000 possible options?

I think in the first instance there are a few techniques we could use to create a serviceable solution. A good solution probably uses several things from this list, but maybe not all of them...

1. Text search: Searching for places by name can help us but it is not a complete solution. Place names are often ambiguous. https://en.wikipedia.org/wiki/List_of_popular_place_names
2. Hierarchy: UN pcodes are hierarchical. For example, the code for admin2 area Nadirshahkot is `20AFG026004`. The AFG part tells us it is in Afghanistan and the `026` part tells us this admin2 area is inside admin1 area `20AFG026`. See the diagram in #618 . This means if we know some higher levels of goegraphy we can constrain our full text search by the higher level areas. For example the search query `name LIKE '%nadir%'` might match lots of admin2 areas globally but if we know the dataset is tagged with `AFG` for Afghanistan and `PAK` for Pakistan, we could also constrain that search to `AND (pcode LIKE '%AFG%' OR pcode LIKE '%PAK%')` to narrow things down.
3. Additional data: We can use the hierarchy or additional fields from the API responses to show the user more than just the name we're searching for. For example, if I'm searching for an admin area that has multiple matches, showing the list
    - San Jose
    - San Jose
    - San Jose
    - San Jose
    - ...
      is not helpful, but if we presented this as:
[...]

PR #622, opened 2021-07-01
import data from Geoportal API

### LINKED ISSUE #618: Harvest and store geographies from the UN Geoportal
## 🌍 Units of Geography

- There are 5 units of geography we are interested in: Countries, Populations of Concern (PoC), Admin1, Admin2 and Reference Places.
- Countries are an outlier. They are not versioned and are coded by ISO 3-letter code.
- The other 4 layers (PoC, Admin1+2, Reference Places) are versioned and are coded using the UN's hierarchical pcode system:

pcode

The geography types we are most interested to start with are countries and PoCs, but we will need to work with all of them at some point.

## 🔗 The APIs

We can query these layers as follows:

### Countries
- https://gis.unhcr.org/arcgis/rest/services/core_v2/wrl_polbnd_int_1m_a_unhcr/FeatureServer/0
- example: https://gis.unhcr.org/arcgis/rest/services/core_v2/wrl_polbnd_int_1m_a_unhcr/FeatureServer/0/query?where=1%3D1&objectIds=&time=&geometry=&geometryType=esriGeometryEnvelope&inSR=&spatialRel=esriSpatialRelIntersects&distance=&units=esriSRUnit_Foot&relationParam=&outFields=*&returnGeometry=false&maxAllowableOffset=&geometryPrecision=&outSR=&gdbVersion=&historicMoment=&returnDistinctValues=false&returnIdsOnly=false&returnCountOnly=false&returnExtentOnly=false&orderByFields=objectid&groupByFieldsForStatistics=&outStatistics=&returnZ=false&returnM=false&multipatchOption=&resultOffset=&resultRecordCount=&returnTrueCurves=false&sqlFormat=none&f=pjson

### Populations of Concern
- https://gis.unhcr.org/arcgis/rest/services/core_v2/wrl_prp_p_unhcr_PoC/FeatureServer/0
- example: https://gis.unhcr.org/arcgis/rest/services/core_v2/wrl_prp_p_unhcr_PoC/FeatureServer/0/query?where=1%3D1&objectIds=&time=&geometry=&geometryType=esriGeometryEnvelope&inSR=&spatialRel=esriSpatialRelIntersects&distance=&units=esriSRUnit_Foot&relationParam=&outFields=*&returnGeometry=false&maxAllowableOffset=&geometryPrecision=&outSR=&gdbVersion=&historicMoment=&returnDistinctValues=false&returnIdsOnly=false&returnCountOnly=false&returnExtentOnly=false&orderByFields=objectid&groupByFieldsForStatistics=&outStatistics=&returnZ=false&returnM=false&multipatchOption=&resultOffset=&resultRecordCount=&returnTrueCurves=false&sqlFormat=none&f=pjson

### Admin 1
- https://gis.unhcr.org/arcgis/rest/services/core_v2/wrl_polbnd_adm1_a_unhcr/MapServer/0
- example: https://gis.unhcr.org/arcgis/rest/services/core_v2/wrl_polbnd_adm1_a_unhcr/MapServer/0/query?where=1%3D1&text=&objectIds=&time=&geometry=&geometryType=esriGeometryEnvelope&inSR=&spatialRel=esriSpatialRelIntersects&relationParam=&outFields=*&returnGeometry=false&returnTrueCurves=false&maxAllowableOffset=&geometryPrecision=&outSR=&returnIdsOnly=false&returnCountOnly=false&orderByFields=objectid&groupByFieldsForStatistics=&outStatistics=&returnZ=false&returnM=false&gdbVersion=&returnDistinctValues=false&resultOffset=&resultRecordCount=&queryByDistance=&returnExtentsOnly=false&datumTransformation=&parameterValues=&rangeValues=&f=pjson

### Admin 2
- https://gis.unhcr.org/arcgis/rest/services/core_v2/wrl_polbnd_adm2_a_unhcr/MapServer/0
- example: https://gis.unhcr.org/arcgis/rest/services/core_v2/wrl_polbnd_adm2_a_unhcr/MapServer/0/query?where=1%3D1&text=&objectIds=&time=&geometry=&geometryType=esriGeometryEnvelope&inSR=&spatialRel=esriSpatialRelIntersects&relationParam=&outFields=*&returnGeometry=false&returnTrueCurves=false&maxAllowableOffset=&geometryPrecision=&outSR=&returnIdsOnly=false&returnCountOnly=false&orderByFields=objectid&groupByFieldsForStatistics=&outStatistics=&returnZ=false&returnM=false&gdbVersion=&returnDistinctValues=false&resultOffset=&resultRecordCount=&queryByDistance=&returnExtentsOnly=false&datumTransformation=&parameterValues=&rangeValues=&f=pjson

### Reference places
- https://gis.unhcr.org/arcgis/rest/services/core_v2/wrl_prp_p_unhcr_ALL/FeatureServer/0
- example: https://gis.unhcr.org/arcgis/rest/services/core_v2/wrl_prp_p_unhcr_ALL/FeatureServer/0/query?where=1%3D1&objectIds=&time=&geometry=&geometryType=esriGeometryEnvelope&inSR=&spatialRel=esriSpatialRelIntersects&distance=&units=esriSRUnit_Foot&relationParam=&outFields=*&returnGeometry=false&maxAllowableOffset=&geometryPrecision=&outSR=&gdbVersion=&historicMoment=&returnDistinctValues=false&returnIdsOnly=false&returnCountOnly=false&returnExtentOnly=false&orderByFields=objectid&groupByFieldsForStatistics=&outStatistics=&returnZ=false&returnM=false&multipatchOption=&resultOffset=&resultRecordCount=&returnTrueCurves=false&sqlFormat=none&f=pjson

Those are some quite long URLs. Lets drill down into the queries (and how we build them) a bit:

arcgis

- `?where=1%3D1`: A query that always evaluates to True i.e: give us all records, no filter.
- `&outFields=*`: Return all available metadata fields in the output.
- `&returnGeometry=false`: Don't return geometries: We don't need the actual polygons (at this stage) - just the metadata fields. This will make the API responses much smaller and easier to deal with.
- `&orderByFields=objectid`: Order by ArcGIS's internal identifier. This may be unnecessary but having the response in a consistent order each time we call the API may be a useful property.
- `&f=pjson`: Give us a response using ArcGIS's pjson format. We can request other formats but given all we need is the metadata fields this is probably the easiest thing to work with.

### Pagination

`&resultOffset=&resultRecordCount=` are used for pagination. 

- The Populations of Concern endpoint allows us to fetch 10,000 records in each API request
- The Countries, Admin 1, Admin 2 and Reference Places endpoints only allow us to fetch 1,000 records in each API request

For countries, that means we can just get them all in one request. For everything else, we need to paginate over the responses using `resultOffset` and `resultRecordCount` as limit+offset params.

## ⛏️ Requirements

- Create models and migrations for a table (or tables) to store geographies in
  - Question: Can we just have one table to store them all? Maybe countries are different from everything else because they are not versioned.
  - Question: How much on the UN's data model do we recreate inside our own DB?
- Write a daily click/cron job to pull geographies by paging over the API responses and update our local table(s)

**TODO:** Add notes on how we deal with versioning/change over time TBC

## 🔗 Related links

- https://gis.unhcr.org/arcgis/rest/services/core_v2
- https://data2.unhcr.org/en/geoservices/
- https://docs.google.com/document/d/1R-DG08eQvchsgk_3mpuhUTJJ19yb2PL2
- https://docs.google.com/document/d/1fOLIMYEGb3v9F4gadZ54ASvEe1L2ga8-

chris48s:
Closes #618

This PR gives us an import script which we can use to harvest data from the UN GeoPortal into RIDL and a model to store it in. Unsurprisingly, there are some nuances :D I've covered a lot of these in code comments, but I'll also included some comments on the diff.

Given this doesn't actually make any user-facing changes, we should be able to just merge this when it is ready because even if we deploy this code it won't actually change anything until we schedule the job. That should also unblock building API/frontend that relies on the data.

Running the import start-to-finish takes about 40 mins.


ckanext/unhcr/models.py
```diff
@@ -53,6 +53,27 @@ class AccessRequest(Base):
     actioned_by = Column(UnicodeText, nullable=True)  # user who approved or rejected the request
 
 
+class GisStatus(object):
+    ACTIVE = 'active'
+    INACTIVE = 'inactive'
+
+
+class Geography(Base):
```
chris48s, line 61: Refs https://github.com/okfn/ckanext-unhcr/issues/618#issuecomment-869656640
Having got into this, I think whichever way we end up presenting it in the front-end I think it is probably makes most sense to store all this in one normalized table and do all the pre-processing at import time. It should make the API endpoint more performant if nothing else.

ckanext/unhcr/models.py
```diff
@@ -53,6 +53,27 @@ class AccessRequest(Base):
...
+    __tablename__ = u'geography'
+
+    globalid = Column(UnicodeText, primary_key=True)
```
chris48s, line 64: > globalid = Column(UnicodeText, primary_key=True)

I started off using `pcode` as the primary key here.

Unfortunately I quickly found we can't rely on `pcode` being unique. There are 3 cases that give us problems here.

1. Sometimes the same area appears twice in the data e.g:

- https://gis.unhcr.org/arcgis/rest/services/core_v2/wrl_polbnd_adm2_a_unhcr/MapServer/0/query?where=pcode%3D%2721PAK002031%27&text=&objectIds=&time=&geometry=&geometryType=esriGeometryEnvelope&inSR=&spatialRel=esriSpatialRelIntersects&relationParam=&outFields=pcode%2Cgis_name&returnGeometry=false&returnTrueCurves=false&maxAllowableOffset=&geometryPrecision=&outSR=&returnIdsOnly=false&returnCountOnly=false&orderByFields=&groupByFieldsForStatistics=&outStatistics=&returnZ=false&returnM=false&gdbVersion=&returnDistinctValues=false&resultOffset=&resultRecordCount=&queryByDistance=&returnExtentsOnly=false&datumTransformation=&parameterValues=&rangeValues=&f=geojson
- https://gis.unhcr.org/arcgis/rest/services/core_v2/wrl_polbnd_adm2_a_unhcr/MapServer/0/query?where=pcode%3D%2720CHL012004%27&text=&objectIds=&time=&geometry=&geometryType=esriGeometryEnvelope&inSR=&spatialRel=esriSpatialRelIntersects&relationParam=&outFields=pcode%2Cgis_name&returnGeometry=false&returnTrueCurves=false&maxAllowableOffset=&geometryPrecision=&outSR=&returnIdsOnly=false&returnCountOnly=false&orderByFields=&groupByFieldsForStatistics=&outStatistics=&returnZ=false&returnM=false&gdbVersion=&returnDistinctValues=false&resultOffset=&resultRecordCount=&queryByDistance=&returnExtentsOnly=false&datumTransformation=&parameterValues=&rangeValues=&f=geojson

2. Sometimes two different objects have the same `pcode` e.g:

- https://gis.unhcr.org/arcgis/rest/services/core_v2/wrl_polbnd_adm2_a_unhcr/MapServer/0/query?where=pcode%3D%2721MMR015001%27&text=&objectIds=&time=&geometry=&geometryType=esriGeometryEnvelope&inSR=&spatialRel=esriSpatialRelIntersects&relationParam=&outFields=pcode%2Cgis_name&returnGeometry=false&returnTrueCurves=false&maxAllowableOffset=&geometryPrecision=&outSR=&returnIdsOnly=false&returnCountOnly=false&orderByFields=&groupByFieldsForStatistics=&outStatistics=&returnZ=false&returnM=false&gdbVersion=&returnDistinctValues=false&resultOffset=&resultRecordCount=&queryByDistance=&returnExtentsOnly=false&datumTransformation=&parameterValues=&rangeValues=&f=geojson
- https://gis.unhcr.org/arcgis/rest/services/core_v2/wrl_polbnd_adm2_a_unhcr/MapServer/0/query?where=pcode%3D%2721MMR014001%27&text=&objectIds=&time=&geometry=&geometryType=esriGeometryEnvelope&inSR=&spatialRel=esriSpatialRelIntersects&relationParam=&outFields=pcode%2Cgis_name&returnGeometry=false&returnTrueCurves=false&maxAllowableOffset=&geometryPrecision=&outSR=&returnIdsOnly=false&returnCountOnly=false&orderByFields=&groupByFieldsForStatistics=&outStatistics=&returnZ=false&returnM=false&gdbVersion=&returnDistinctValues=false&resultOffset=&resultRecordCount=&queryByDistance=&returnExtentsOnly=false&datumTransformation=&parameterValues=&rangeValues=&f=geojson

I think these two cases indicate errors in the data. These are uncommon, but they exist and even if we flag them and their GIS team fix them, all there is nothing to stop them happening again in future.

The third problem is not an error.

3. Not everything in the countries layer is actually a country. There are also some "secondary territories". Some of these things have a non-unique pcode so for example the iso3 code for Portugal is `PRT` and the iso3 code for the Azores is also `PRT`. This isn't wrong - just inconvenient. It means we can't rely on the `iso3` (which I'm using as the `pcode` for countries) to be unique for totally valid reasons.

So I switched to using `globalid` (which does appear to be unique/stable) as the primary key to solve all of those issues.


Given that context, I'm not sure whether it will make most sense to attach pcodes to datasets and deal with the fact that a pcode can map to >1 records, or if it would be more sensible to just attach globalids to datasets so it is a strict one-to-one but hide the concept of `globalid` from the user.

ckanext/unhcr/arcgis.py
```diff
@@ -0,0 +1,223 @@
...
+
+
+def merge_dicts(*args):
+    out_dct = {}
+    for arg in args:
+        out_dct.update(arg)
+    return out_dct
+
+
+def upsert_features(session, features):
```
chris48s, line 75: Approach copied from https://gist.github.com/malexer/0647b208b2f1c48ec93e1bd157dc67c0

ckanext/unhcr/arcgis.py
```diff
@@ -0,0 +1,223 @@
...
+        layer=layer,
+        hierarchy_pcode=properties['hierarchy_pcode'],
+    )
+
+
+@retry(RequestException, tries=3, delay=10)
+def request_wrapper(*args, **kwargs):
+    """
+    The Geoportal API is quite flaky and sometimes responds with a error
+    but then the same request works a few seconds later.
```
chris48s, line 59: :angry:

ckanext/unhcr/migration/unhcr/versions/006_65d2d180530c_add_geography_models.py
```diff
@@ -0,0 +1,25 @@
...
+
+"""
+from alembic import op
+import sqlalchemy as sa
+from ckanext.unhcr.models import Geography
+
+
+# revision identifiers, used by Alembic.
+revision = '65d2d180530c'
+down_revision = '058f761ed45f'
```
chris48s, line 15: TODO: There are other PRs open with migrations so depending on which of the open PRs gets merged first I might need to update this ID

chris48s:
Gah! :angry:  There is some issue with the tests that is causing `responses` not to deactivate - its intercepting the HTTP calls for all subsequent tests in the suite which is making everything fail trying to talk to the SOLR container.. I'll have to work it out next week

ckanext/unhcr/models.py
amercader: > Given that context, I'm not sure whether it will make most sense to attach pcodes to datasets and deal with the fact that a pcode can map to >1 records, or if it would be more sensible to just attach globalids to datasets so it is a strict one-to-one but hide the concept of globalid from the user.

It makes sense to use `globalid` as PK, but also thinking in terms of interoperability I think it would be nice if the RIDL API (ie the dataset metadata) exposed pcodes, so it could be integrated more easily with other UN tools. So on #619 we'll need to figure out what gets stored and expose against the dataset metadata. I think it make sense to have two different dataset fields, one for `globalid(s)` and one for `pcode(s)` so users can use the more suitable one (and also us, when displaying regions in the dataset page, rendering the form snippets, etc)

ckanext/unhcr/arcgis.py
```diff
@@ -0,0 +1,233 @@
...
+                str(row.count).rjust(7)
+            )
+        )
+
+
+URL_TEMPLATE = 'https://gis.unhcr.org/arcgis/rest/services/core_v2/{}/MapServer/0/query'
+
+
+LAYERS = [
+    'wrl_polbnd_int_1m_a_unhcr',
```
amercader, line 116: For the benefit of something new coming to this code, can you add a small comment next to each layer with their human readable name (Countries, Admin level1 , etc)

amercader (approved):
@chris48s This looks great. The model is the right approach and the command looks solid. I run it for a while without issues.

Thinking ahead when actually leveraging these geometries for attaching them to datasets we might need to store further fields from the original data. I'm making this up, but things like `loc_type`, `loc_subtype`, `pop_type` or `poc_subtype` might be potentially useful to have for filtering etc. If that ends up happening we can add a JSON field to the Geographies table for "other" fields. But again, this is hypothetical and I guess that if needed we can add a migration and tweak the import script.

Let's get this merged and we'll start thinking about #619

ckanext/unhcr/click_commands.py
```diff
@@ -27,6 +28,15 @@ def init_db():
     print(u'UNHCR tables initialized')
 
 
+@unhcr.command(
+    u'import-geographies',
+    short_help=u'Import geographies from UNHCR GeoPortal'
+)
+@click.option('-v', '--verbose', count=True)
```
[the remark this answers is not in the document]
chris48s: Yes you're right. Removed in 30491e6

ckanext/unhcr/arcgis.py
chris48s: Yes. One thing I qickly realised as I started using this is we need a mapping of layer names to "friendly" names for the user anyway and that can help us as devs too so I've pulled 4768547 forward from that branch. I reckon

```py
COUNTRY['layer_name'],
```

is probably even more desirable than

```py
'wrl_polbnd_int_1m_a_unhcr'  # country
```

becuase we've got a "single source of truth" for these "magic" string constants and the comment rot can't set in.

ckanext/unhcr/models.py
chris48s: Yes. I quickly realised as I started this that even if we ignore the cases where there is a data error (e.g: duplicates) we need to use an unambiguous ID for the relationship because of the secondary territories case: If we just slap an ISO-3 code on a dataset then we would have no way of knowing if this dataset is related to Portugal only, the Azores only, or both (and so on for other countr/secondary territory pairs).

Once I started down that path, I realised another thing which is that the out-of-the box format of the UNHCR globalids is going to give us a headache due to the serialization format CKAN uses for tags. See https://github.com/okfn/ckanext-unhcr/blob/2a06af292768c48fc676a99d990d7d7f41e6771f/ckanext/unhcr/utils.py#L16-L22 I think the path of least resistance here is going to be that we strip the leading `{` and trailing `}` off the ID before we store it see 1a7464e . So when we get `"globalid":"{8F43E6AA-1D96-4419-80FE-D7A72563752E}"` we just store `8F43E6AA-1D96-4419-80FE-D7A72563752E`. I think that's the path of least resitance.

Assuming we go with that, if you have already imported data you will need to clear your geography table and re-import or fiddle the IDs in-place to avoid dupes because I've changed the ID format.

Also good point on data interoperability. For user-facing stuff I'm already working on the assumption we use globalid internally but display pcodes in the front-end but it does need a bit more thought on the API side of things :+1:

chris48s:
just so it doesn't get lost one of the replies to this review is further up in https://github.com/okfn/ckanext-unhcr/pull/622#discussion_r673933541

amercader:
All good here @chris48s , we can merge this. Just so we don't forget can you prepare a small PR on docker-ckan-aws-unhcr with the scheduled command that we can merge whenever is needed?

PR #628, opened 2021-07-13
Move visibility field from dataset to resources

avdata99:
Replaces #583 (now based on master and squashed)
Related to #581

This PR:
 - Copy field definition from dateset to resource (required only at resource level)
 - Remove `visibility` field from dataset form
 - Add `visibility` field to resources form
 - Adds a migration script to copy `visibility` values from datasets to resources
 - Include [ckanextt-ddi PR](https://github.com/okfn/ckanext-ddi/pull/8) to import and take care of visibility for new resources


ckanext/unhcr/helpers.py
```diff
@@ -984,6 +984,18 @@ def can_download(package_dict):
...
+def can_request_access(user, pkg):
+    """ Define if user is able to request access for private resources in dataset """
+    resources = pkg['resources'] if type(pkg) == dict else pkg.resources
+    if len(resources) == 0:
+        return False
+    if can_download(pkg):
+        return False
+    pkg_id = pkg['id'] if type(pkg) == dict else pkg.id
+    access_already_requested = get_existing_access_request(user.id, pkg_id, 'requested')
+    return not access_already_requested
```
avdata99, line 996: Helper to use in the UI to define whether the access request form should be available

ckanext/unhcr/jobs.py
```diff
@@ -118,17 +123,18 @@ def _modify_date_range(package, key_start, key_end):
...
-    package[key] = None
+    package[key] = default_value
 
     # Iterate resources
     for resource in package['resources']:
         if resource.get(key) is None:
             continue
-        package_weight = weights.get(package[key], 0)
-        resource_weight = weights.get(resource[key], 0)
+        package_weight = weights.get(package.get(key), 0)
```
avdata99, line 136: "Visibility" is no longer a required field, so it may not be present.

ckanext/unhcr/validators.py
```diff
@@ -192,6 +192,16 @@ def visibility_validator(key, data, error, context):
...
 
 
+def visibility_validator_resource(key, data, error, context):
+    ''' Validates visibility has a correct value.
+    Visibility only has two values in the schema, 'restricted' and 'public'. 
+    '''
+    value = data.get(key)
+
+    if value not in ('restricted', 'public'):
+        raise Invalid('Invalid value for the visibility field')
```
avdata99, line 202: @chris48s It is not clear how visibility='private', should work now that visibility is for resources and just a replica within datasets.
We still have the visibility_validator for datasets but this behavior will be overridden by the resources visibility

chris48s: I think at the resource level it is ok to do what you've done here and just say that only `'restricted'` and `'public'` are the allowed values as it is a completely new field and there is no established meaning of a `'private'` resource in CKAN.

chris48s:
I had another look over this today. Aside from getting the background job working with package_revise I think this is done (although I think we should just pull that work into this branch - I'll comment on #629

chris48s (approved):
Ready to go, but again we need to get py3 to production or deploy from a different branch before we merge it

chris48s (changes requested):
@avdata99
Now that we've merged #622 to master we need to make a small update to this PR to get the migrations into the right order. Can you:
- Rebase this branch onto master
- Rename `004_0a7c08639f19_visibility_for_resources.py` to `005_0a7c08639f19_visibility_for_resources.py`
- Change the `down_revision` in that migration to `65d2d180530c`

Thanks :+1:

avdata99:
@chris48s: Rebased and updated [here](https://github.com/okfn/ckanext-unhcr/pull/628/commits/f3065f20bcd43264fd9db4da18d06770e155d02e)

chris48s (approved):
nice - thanks

Issue #614, opened 2021-06-18
Revise package and resources before save


avdata99:
Check [this comment](https://github.com/okfn/ckanext-unhcr/pull/583/files#r653581140).
After merging #583 and before going to `master` ensure the `package_revise` function work as we expect.

We are modifying resources and we need to be sure all resources are saved as we expect.

avdata99:
Merged

PR #629, opened 2021-07-13
Resource revise

avdata99:
Related to #614
To merge before #628 

Ensure resources are updated when the package is updated/revisated


chris48s:
I don't think this is the right solution as it will probably re-introduce the race conditions we were trying to avoid.

To give a bit of background on this: We were finding that in some cases when we were processing multiple resource uploads to the same dataset at the same time or in quick succession (via the API) some resources were "disappearing". When a resource first gets uploaded it is created as 'pending' then marked 'active' later. Because we are writing to the package in a background job we were sometimes finding that resources were getting marked 'active' and then immediately set back to 'pending' in the background jobs because we did the initial `package_show` while some of the resources were still 'pending'.

Just writing the whole resource dict for every resource every time we `package_revise` will re-introduce this bug. What we really need to do is update #628 to only update the relevant resource field(s) for the resource(s) we are processing (I think just `visibility` is the only field we actually need to change?). Based on the docs it looks like this is possible but I'm not sure what the exact syntax we need is. The job is to figure that out :)

avdata99:
@chris48s the only relevant change on `resources` is the `visibility` field.
I make [a change](https://github.com/okfn/ckanext-unhcr/pull/629/commits/85163edb3041402f7b63a3ac94b727e5e7b3b9dd) that is probably better but not sure if this the right solution.
Maybe you can take a look @amercader.
It seems that `package_revise` is a bit tricky for resources..
Tests are green but I'm not sure if we're already capturing the race condition where these updates fail.

chris48s:
So I think part of the reason this is hard to test is because you are trying to make sure the resource visibility field gets updated on a branch where there is no code modifying the resource visibility field. It will be much easier to test this work if you do it on a branch where you've got some code in the background job which touches the field you're trying to modify :)

The race condition is not tested because it is really hard to create a condition that relies on concurrent requests timed in a certain way under test, but basically what we want here is code that updates `resource.extras.visibility` and leaves all the other resource fields alone. If that's what is happening we should be good.

avdata99:
[These new tests](https://github.com/okfn/ckanext-unhcr/pull/629/commits/6a84c7ad258249b25cedfe164773b280c920aa75) ensures we update resources and packages with `package_revise` as expected.
If it's ok for you @chris48s, I think we can merge this and continue discussion in the main PR for visibility.

PR #632, opened 2021-07-14
Allow plugins to update the user

avdata99:
Avoid saml2 errors

### DIFF
--- a/ckanext/unhcr/actions.py
+++ b/ckanext/unhcr/actions.py
@@ -1253,20 +1253,23 @@ def user_create(up_func, context, data_dict):
 
 @toolkit.chained_action
 def user_update(up_func, context, data_dict):
-    user_id = toolkit.get_or_bust(data_dict, 'id')
-    m = context.get('model', model)
-    user_obj = m.User.get(user_id)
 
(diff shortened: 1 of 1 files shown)


ckanext/unhcr/actions.py
```diff
@@ -1253,20 +1253,23 @@ def user_create(up_func, context, data_dict):
 
 @toolkit.chained_action
 def user_update(up_func, context, data_dict):
-    user_id = toolkit.get_or_bust(data_dict, 'id')
-    m = context.get('model', model)
-    user_obj = m.User.get(user_id)
 
-    if user_obj is not None and is_saml2_user(user_obj):
+    if not context.get('ignore_auth', False):
```
amercader, line 1257: The default return vale for `.get()` is None
```suggestion
    if not context.get('ignore_auth'):
```

ckanext/unhcr/actions.py
```diff
@@ -1253,20 +1253,23 @@ def user_create(up_func, context, data_dict):
...
+        # allow plugins to update the user
```
amercader, line 1258: This comment here is confusing, better to add it to the up_func() call at the end

amercader (approved):

PR #636, opened 2021-07-22
Tag datasets with geographies

chris48s:
Refs #619

This PR:
 - Index Geographies (and parent ones) for searching
 - [Hardcode* GIS names](https://github.com/okfn/ckanext-unhcr/blob/619-geo-part2/ckanext/unhcr/src/js/geographies.js) at package form start
 - (extra) Drop patches after CKAN 2.9.4 release

\* It seems `autocomplete` CKAN module do not include any option to _name_ elements when the came from an external source. Should we fix this at core?

UI search by parent GIS working

Selection_408

Starting package form, showing custom name instead Geography `globalid`

image


ckanext/unhcr/blueprints/api.py
```diff
@@ -0,0 +1,43 @@
...
+            "error": {
+                "__type": "Authorization Error",
+                "message": "Access denied: geography_autocomplete requires an authenticated user"
+            },
+            "success": False
+        }
+        return _finish(403, error, 'json')
+
+    q = toolkit.request.args.get('q', '')
+    limit = toolkit.request.args.get('limit', 100)
```
chris48s, line 21: I think its reasonable for the default limit to be fairly large here. A drop-down with 100 search results in it is still quite cumbersome, but given the total search space is >200,000 objects and there are lots of placed with similar names I reckon limiting to the first 10-20 results is too restrictive. This seems about right to me.

ckanext/unhcr/auth.py
```diff
@@ -417,6 +417,14 @@ def external_user_update_state(context, data_dict):
     return {'success': False}
 
 
+def geography_autocomplete(context, data_dict):
+    return {'success': True}
+
+
+def geography_show(context, data_dict):
+    return {'success': True}
```
chris48s, line 425: All the data returned by these actions is data we downloaded from the UN's public APIs so I think it is ok for any authenticated user to call them.

ckanext/unhcr/blueprints/api.py
```diff
@@ -0,0 +1,43 @@
+from flask import Blueprint
+import ckan.plugins.toolkit as toolkit
+from ckan.views.api import _finish, _finish_ok, API_REST_DEFAULT_VERSION
```
chris48s, line 3: I didn't see any nice way to avoid importing private methods here. I don't think the extension toolkit has methods for returning JSON API responses, does it?

ckanext/unhcr/plugin.py
```diff
@@ -32,6 +32,8 @@
     'datastore_info',
     'datastore_search',
     'format_autocomplete',
+    'geography_autocomplete',
+    'geography_show',
```
chris48s, line 36: External users will need to be able to call these actions when creating/viewing deposited datasets

ckanext/unhcr/templates/scheming/form_snippets/geographies.html
```diff
@@ -0,0 +1,21 @@
+<div class="form-group control-full">
+  <label class="control-label" for="field-geographies">Geographies</label>
+  <div class="controls">
+    <input
```
chris48s, line 4: This isn't the exact right approach here but lets have a conversation about it as the next job.

ckanext/unhcr/actions.py
```diff
@@ -1174,6 +1186,66 @@ def user_autocomplete(context, data_dict):
...
+    m = context.get('model', model)
+    q = data_dict['q']
+    limit = data_dict.get('limit', 100)
+
+    toolkit.check_access('geography_autocomplete', context, data_dict)
+
+    query = m.Session.query(
+        Geography
+    ).filter(
+        Geography.gis_name.ilike(f'%{q}%')
```
chris48s, line 1221: This provides a functional search and will do for an MVP, but it would be more user-friendly to have a search that ignores accents, diacritics etc. For example, searching for "beaupre" with `ilike` won't match the name "La Côte-de-Beaupré" - you have to explicitly search for "beaupré" to match that.

ckanext/unhcr/actions.py
```diff
@@ -51,6 +52,17 @@ def package_update(up_func, context, data_dict):
...
+    # add any parent geographies
+    if data_dict.get('geographies'):
+        ids = data_dict['geographies'].split(',')
+        for id_ in ids:
+            geog = m.Session.query(Geography).get(id_)
+            if geog:
+                parents = geog.parents
+                ids = ids + [p.globalid for p in parents]
+        ids = list(set(ids))
+        data_dict['geographies'] = ",".join(ids)
```
chris48s, line 64: TODO: All this feature does as it stands is store a list of globalids against a dataset. This is what we need to unambiguously connect datasets with geographies. However, for interoperability purposes it is going to be more useful for data consumers to see UN pcodes so it will be desirable to expose pcode via the API either as well as or instead of the globalid.
One way to achieve this would be to also store the pcodes as well as the globalids here when we save the package. The other would be to use the globalids from the DB to infer the pcodes in `package_show` and add them to the dict there.

chris48s: Also: The way this is currently written there is no distinction between geographies which the user picked themselves and geographies which are inferred from the hierarchy. For example: If I tag a dataset with "Paris" the way this is written now, next time I edit it it will be tagged with both "Paris" and "France", even though I didn't explicitly tag it with France.
Question for the client: Is this ok, or do we want this distinction between tags the user has chosen and additional tags that have been automatically chosen?

chris48s:
In CKAN there is an "autocomplete" module
https://github.com/ckan/ckan/blob/ckan-2.9.3/ckan/public/base/javascript/modules/autocomplete.js
This seems to essentially have two modes of operation:

- A `<select>` HTML element
    - closed (but searchable) option list
    - allows `<optgroup>`s
    - you have to write every option to the page at HTML render time
- An `<input>` HTML element
    - allows free text
    - no `<optgroup>`s
    - works with an `autocomplete` API endpoint

Neither of these does exactly what we want - we really need some features from each of those buckets.

In this PR I have implemented a prototype of a frontend with `<input>`, but it has several problems that need to be solved:

1. The structure is flat, but it should be nested in `<optgroup>`s.

Screenshot at 2021-07-23 16-10-20

I worked round this by encoding the hierarchy in the `display_name` but it would be much nicer if Admin2 was an `<optgroup>` and Reference Place was an `<optgroup>`. In principle, it is possible for select2 to receive a nested data structure and render a `<select>` with `<optgroup>`s. See https://select2.org/data-sources/formats#grouped-data So in an ideal world, our autocomplete endpoint should be able to return a data structure like

```json
[
    {
        "name": "Admin 1",
        "children": [
            {
                "globalid": "D5C44369-FDF9-4B6D-906A-1BA71BDF0C39",
                "name": "Admin 1: Bratislava (20SVK002)"
            }
        ]
    },
    {
        "name": "Admin 2",
        "children": [
            {
                "globalid": "E969C52D-D227-4619-951E-67B5372439FA",
                "name": "Admin 2: Okres Bratislava I (20SVK002002)"
            },
            {
                "globalid": "D3E1D1E1-07CE-4D8E-8FAB-3DB948145DF9",
                "name": "Admin 2: Okres Bratislava II (20SVK002003)"
            }
        ]
    }
]
```

and then select2 should render it. In practice, this does not work. I think the reason for this is that CKAN's `parseCompletions` function in https://github.com/ckan/ckan/blob/ckan-2.9.3/ckan/public/base/javascript/client.js#L155 ignores the `children` key. **Extending CKAN to support/expose this feature of select2 would seem like a reasonable upstream contribution to CKAN core.**

2. You can add arbitrary text as well as results from the autocomplete endpoint (although we'll reject arbitrary text when we try to submit the form)

Screenshot at 2021-07-23 16-02-54

Again, it would seem reasonable for the autocomplete component to have some mode of operation that says you can _only_ add tags which were returned from the autocomplete endpoint (and not arbitrary text). I phrased this as a `<select>` element with autocomplete endpoint when we talked about this, but on reflection maybe I'm wrong and that is not the solution - maybe the same thing could be achieved using an `<input>` with an additional client-side validation on what you can add to it to constrain it to only options from the autocomplete endpoint. **Either way, some solution to this issue would also seem like a reasonable upstream contribution to CKAN core**

3. When you edit a dataset you've already added tags to, they're displayed as `globalid`s not user-friendly display names.

Screenshot at 2021-07-23 16-04-31

**Fundamentally this is an issue we need to solve in our scheming `form_display` widget in RIDL - it is not a concern for CKAN's autocomplete input element.**

---

We probably do need to solve all of these these problems, but maybe getting some feedback from the client first is the next step before jumping in to them.

amercader:
Latest update after discussion with the client:

* Let's try to implement the changes Chris described above in the autocomplete module / our endpoint so we can display the options returned by the geography lookup endpoint with headings (`optgroup`) ordered hierarchically 
* We want to index the name of the geography selected by the user so it becomes available for searching, but also their parent geographies. I don't think we need to store the parents as part of the metadata as these can be computed at index time, but in any case if we need it should be done in a separate field
* We should store the global id (what we are storing now) but also the pcodes in a separate field (these should be automatically filled at validation time)

Let's focus on the UI widget first and then we can look at the search more in depth

avdata99:
@amercader rebased to latest release. 
My first proposed change is to [remove parent geographies](https://github.com/okfn/ckanext-unhcr/pull/636/commits/f8812249be361e6f4088d2140ca82cb09ea707a4) from the geographies field and show them (and allow to search by) using a different field which will be synced from the Geography local table any time geography change

avdata99:
@amercader I just [created read-only parent geographies field](https://github.com/okfn/ckanext-unhcr/commit/4f22d43db44d0d78ef9806394e7f18fc714dac06)
Selection_393.
My plan is to ensure we can search on them 

After that my plan is to fix how we show the field on the edit form

Selection_394

avdata99:
@amercader I will expect `package_search` to find any data added at `before_index` [like this](https://github.com/okfn/ckanext-unhcr/blob/619-geo-part2/ckanext/unhcr/plugin.py#L338-L350)
But test fails ([here](https://github.com/okfn/ckanext-unhcr/blob/619-geo-part2/ckanext/unhcr/tests/pt/controllers/test_admin.py#L124))
So, I decided to [add a new field](https://github.com/okfn/ckanext-unhcr/blob/619-geo-part2/ckanext/unhcr/schemas/dataset.json#L685) and [update its value](https://github.com/okfn/ckanext-unhcr/blob/619-geo-part2/ckanext/unhcr/jobs.py#L135) with the names and pcodes for Geographies. 
This works but I'm not sure if that is what we want.

amercader:
@avdata99 Sorry I hadn't looked at the Solr schema fields in a while and didn't remember how they work.

If we add a `geo_search_info` field it will get indexed using this Solr field:

https://github.com/ckan/ckan/blob/77425a04ff2a956e8dfe04c1364c12f1d342874b/ckan/config/solr/schema.xml#L174

This is a standard single text field so what will get indexed is a single string, eg `Admin 1: Nisporeni (20MDA023),Country: Republic of Moldova (MDA)`.

What we want to use is a multi-valued field like this one:

https://github.com/ckan/ckan/blob/77425a04ff2a956e8dfe04c1364c12f1d342874b/ckan/config/solr/schema.xml#L173

So if we instead add the `vocab_geo_search_info` field and add a list of the geographies we will get each individual value indexed. This seems to work fine for me.

Note that instead of using `str(geog)` [here](https://github.com/okfn/ckanext-unhcr/blob/619-geo-part2/ckanext/unhcr/plugin.py#L346) we should use just the actual name so we don't index things like `Admin 2:` or `Country`. We can also add the pcodes as an item for power users. So what we end up indexing should look like:

```python
pkg_dict['vocab_geographies'] = ['Nisporeni', '20MDA023', 'Republic of Moldova', 'MDA)']
```

At the schema level we don't need a field for this.

amercader:
Regarding the "Parent Geographies" field, is that really needed? I guess if we want to display the parent geographies in the dataset page it makes sense to store them, but we should check that with the client.

I also think that we should have a custom display_snippet for geographies that shows both the user entered ones and the parent ones on the same field, with some styling to differentiate them.

avdata99:
@amercader 
 - The UI now shows geos including all parents information
 - Removed all unnecessary files (parent field and srt geos)
 - Search index works fine (we only index the list of values required to search)
 - Only is pending to fix an ~error. If I search using a pcode (e.g. `20IRQ015`) I get the dataset using this geograpy but also another dataset using another geograpy

avdata99:
Now the display for reading it's better

Selection_410

About the form view, we have thousands of geographies. The select - optgroup option by loading all of them seems a bad idea
I'll try to improve the way we show them

amercader:
@avdata99 thanks, I'll review this on Monday. Regarding this
> About the form view, we have thousands of geographies. The select - optgroup option by loading all of them seems a bad idea
I'll try to improve the way we show them

I think select + optgroup plus the autocomplete API is a good first step for an MVP

ckanext/unhcr/schemas/deposited_dataset.json
```diff
@@ -710,6 +710,13 @@
             "help_text": "Links to other RIDL datasets. It supports multiple selections",
             "required": false
         },
+        {
+            "field_name": "geographies",
+            "preset": "geographies",
+            "label": "Geographies",
+            "help_text": "One or more geographies related to this dataset",
+            "required": false
+        },
```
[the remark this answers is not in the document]
amercader: `json.dumps()` will help here only if we have an `output_validator` that parses it back to a list with `json.loads()`. But then we don't actually need to bother with json and just parse the comma-separated string that we have now. I think we just need this:

```diff
diff --git a/ckanext/unhcr/plugin.py b/ckanext/unhcr/plugin.py
index c4476d4e..74068459 100644
--- a/ckanext/unhcr/plugin.py
+++ b/ckanext/unhcr/plugin.py
@@ -580,6 +580,7 @@ class UnhcrPlugin(
             'upload_not_empty': validators.upload_not_empty,
             'object_id_validator': validators.object_id_validator,
             'activity_type_exists': validators.activity_type_exists,
+            'output_list': validators.output_list
         }
 
     # IPermissionLabels
diff --git a/ckanext/unhcr/schemas/presets.json b/ckanext/unhcr/schemas/presets.json
index 2f064797..1e717eeb 100644
--- a/ckanext/unhcr/schemas/presets.json
+++ b/ckanext/unhcr/schemas/presets.json
@@ -15,6 +15,7 @@
       "preset_name": "geographies",
       "values": {
         "validators": "ignore_missing geographies_validator",
+        "output_validators": "output_list",
         "form_snippet": "geographies.html",
         "display_snippet": "geographies.html",
         "label": "Geographies",
diff --git a/ckanext/unhcr/validators.py b/ckanext/unhcr/validators.py
index 67a3c2c9..1981e3c0 100644
--- a/ckanext/unhcr/validators.py
+++ b/ckanext/unhcr/validators.py
@@ -60,6 +60,13 @@ def geographies(value, context):
     return value
 
 
+def output_list(value):
+
+    if not value:
+        return []
+    return value.split(',')
+
+
```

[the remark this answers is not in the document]
avdata99: I applied [this changes](https://github.com/okfn/ckanext-unhcr/commit/6a3138973e2c9efe809f32287cdbbc377e9fcfea) @amercader 
We already have an output list function so I reuse (and fixed) it.

ckanext/unhcr/plugin.py
```diff
@@ -580,6 +580,7 @@ def get_validators(self):
             'upload_not_empty': validators.upload_not_empty,
             'object_id_validator': validators.object_id_validator,
             'activity_type_exists': validators.activity_type_exists,
+            'output_list': helpers.normalize_list,
```
amercader, line 583: Sorry, I know this is a bit anal, but can you create a validator in the validators module (that can call `helpers.normalize_list()`) and register it here? I like to keep things tidy

amercader:
@avdata99 I added one last annoying comment and then this can be merged!

I think there are things we can polish but these can be done separately

amercader (approved):

ckanext/unhcr/plugin.py
avdata99: Perfect, done [here](https://github.com/okfn/ckanext-unhcr/commit/1de1b5d9de4991e6ec14ce88195749121a5a19db)

jj-crypto (approved):

Issue #659, opened 2021-08-06
Don't allow updating KoBo special resources


avdata99:
The special resources we create automatically after import datasets from KoBo should be protected
The user can change metadata but no the data files or questionnaire.
Define with the client if we want to allow creating new versions or update data periodically from KoBo API

PR #666, opened 2021-08-13
Prevent uploading files to automatically imported KoBo resources

avdata99:
Related to #659


ckanext/unhcr/tests/pt/controllers/test_resource.py
```diff
@@ -183,6 +205,30 @@ def test_resource_download_valid(self, app):
...
+    # def test_get_edit_kobo_resource_works(self, app):
+    #     url = toolkit.url_for(
+    #         'resource.edit',
+    #         id=self.kobo_dataset['id'],
+    #         resource_id=self.kobo_resource['id']
+    #     )
+    #     env = {'REMOTE_USER': self.sysadmin['name'].encode('ascii')}
+    #     resp = app.get(url, extra_environ=env, status=200)
+    #     # TODO it fails probably for the JS running after page is loaded
+    #     assert 'KoBo files cannot be updated' in resp.body
```
avdata99, line 231: Is there an easy way to test response after `kobo-file-upload` JS is executed?

ckanext/unhcr/templates/scheming/package/snippets/resource_form.html
```diff
@@ -53,8 +53,12 @@ <h1>Additional resources</h1>
     <i class="fa fa-info-circle"></i>
     Max upload size: {{ h.get_max_resource_size() }}Mb
   </div>
-  <input type="hidden" value="{{ data.kobo_type }}" name="kobo_type" />
 
+  <input type="hidden" value="{{ data.kobo_type }}" name="kobo_type" id="kobo_type_field"/>
```
avdata99, line 57: This is for JS to detect that it is a KoBo resource and the upload field needs to be hidden.

ckanext/unhcr/fanstatic/theme.js
```diff
@@ -270,3 +270,18 @@ this.ckan.module('resource-type', function ($) {
 
   };
 });
+
```
amercader, line 273: I assume that by having it [here](https://github.com/okfn/ckanext-unhcr/pull/666/files#diff-ed1ca4e4e9f6e1101ba3c466701e3920ea33b4bcd0ce5d8f6e9dc65bb594c168R1) we no longer need it in theme.js? Or vice versa. In any case the code is duplicated.

ckanext/unhcr/fanstatic/theme.js
```diff
@@ -270,3 +270,18 @@ this.ckan.module('resource-type', function ($) {
...
+  /* 
+  Update file resources from KoBo is not allowed since those files are "official" data from KoBo
+  */
+ 
+  if ($('#kobo_type_field').val() != "" ) {
+    msg = "You can change metadata but actual <b>KoBo files cannot be updated</b>.";
+    old_url = $('#field-image-url').val();
+    preserve_url = "<input type='hidden' name='url' value='" + old_url + "'>";
+    $('.image-upload').html('<div class="form-group control-full">' + msg + '</div> ' + preserve_url);
+  }
```
amercader, line 285: Some tweaks to language and style:

Screenshot 2021-08-25 at 13-28-14 Edit - Survey json data - RIDLTest - RIDL


```suggestion

  /* 
  Update file resources from KoBo is not allowed since their files are always directly imported from KoBo
  */
  let kobo_type = $('#kobo_type_field').val();
  if (kobo_type !== undefined && kobo_type != "") {
    // TODO: link to where you can re-import them
    msg = "You can change the resource metadata but the actual data files are imported from Kobo.";
    old_url = $('#field-image-url').val();
    preserve_url = "<input type='hidden' name='url' value='" + old_url + "'>";
    $('.image-upload').html('<div class="alert alert-warning">' + msg + '</div> ' + preserve_url);
  }
```

This is to cover the case where the input field is not present in the page and has a value:

```
if (kobo_type !== undefined && kobo_type != "") 
```

ckanext/unhcr/templates/scheming/package/snippets/resource_form.html
amercader: This is fine, but FYI you can also use `name` attributes to select elements in jquery: 

```js
$('input[name="kobo_type"]')
```

ckanext/unhcr/actions.py
```diff
@@ -766,8 +766,26 @@ def resource_update(up_func, context, data_dict):
...
+
+    # ensure kobo_details are not missing
+    old_resource = toolkit.get_action('resource_show')(context, {'id': data_dict['id']})
+    data_dict['kobo_type'] = old_resource.get('kobo_type')
+    data_dict['kobo_details'] = old_resource.get('kobo_details')
+
+    if data_dict.get('kobo_type'):
+        original_url = data_dict.pop('original_url', None)
+        upload_changed = original_url and data_dict.get('url') != original_url
+        # if the users try to change a "special" (extra) resource with a file,
```
amercader, line 780: The word "special" is from now on completely banned in this code base :slightly_smiling_face: 

Let's use "automatically imported" or "kobo data file"

ckanext/unhcr/actions.py
```diff
@@ -766,8 +766,26 @@ def resource_update(up_func, context, data_dict):
...
+        # we should raise an error
+        if upload_changed:
+            raise toolkit.ValidationError(
+                {'data file': ['You cannot update a special KoBo resource with a new file']}
```
amercader, line 784:
```suggestion
                # TODO: say how
                {'data file': ['You cannot update a KoBo data file directly, please re-import the data instead']}
```

ckanext/unhcr/templates/package/resource_read.html
```diff
@@ -68,7 +68,37 @@
...
+{% if res.kobo_type %}
+<section class="module">
+  
+  <div class="module-content">
+    <h2>{{ _('KoBo information') }}</h2>
+    <table class="table table-striped table-bordered table-condensed" data-module="table-toggle-more">
+      <tbody>
+        <tr>
+          <th scope="row">{{ _('KoBo asset ID') }}</th>
+          <td>{{ c.package.kobo_asset_id }}</td>
```
amercader, line 84: Unless strictly necessary we should no longer use variables in the `c` object in templates. In 2.9 all core properties added to `c` are now passed directly to templates:

```suggestion
          <td>{{ package.kobo_asset_id }}</td>
```

ckanext/unhcr/tests/pt/controllers/test_resource.py
```diff
@@ -249,3 +297,44 @@ def test_edit_resource_must_provide_upload(self, app):
...
+
+            'url': 'different-file.csv',
+            'clear_upload': '',
+            'save': ''
+
+        }
+
+        resp = app.post(url, data=data, extra_environ=env, status=200)
+
+        print(resp.body)
```
amercader, line 338: Remove print

ckanext/unhcr/tests/pt/controllers/test_resource.py
```diff
@@ -183,6 +205,32 @@ def test_resource_download_valid(self, app):
         result = model.Session.execute(sql).fetchall()
         assert 1 == len(result)
 
+    # Resource edit form
+
+    # TODO find a way to test JS after page is loaded
```
amercader, line 210: The only way to test this properly is with frontend tests like Cypress. We don't have these set up so I think that if you are testing the behavior of sending the form we should be fine.
These commented out tests can be removed

ckanext/unhcr/tests/pt/controllers/test_resource.py
```diff
@@ -249,3 +297,44 @@ def test_edit_resource_must_provide_upload(self, app):
...
+
+            # this is manually added to the update form as hidden fields:
+            'original_url': self.kobo_resource['url'],
+            'kobo_type': 'data',
+
+            'url': 'different-file.csv',
+            'clear_upload': '',
+            'save': ''
+
+        }
```
amercader, line 334: Perhaps this is more readable:
```python

form_data = {
    'url_type': 'upload',
    'upload': mocks.FakeFileStorage(filename='different-file.csv'),
    # this is manually added to the update form as hidden fields:
    'original_url': self.kobo_resource['url'],
    'kobo_type': 'data',
    'url': 'different-file.csv',
    'clear_upload': '',
    'save': ''
}

data = dict(self.kobo_resource, **form_data)

```

BTW if `self.kobo_resource` is only used in this test just create it here instead of the class constructor

ckanext/unhcr/kobo/kobo_dataset.py
```diff
@@ -118,6 +118,10 @@ def create_kobo_resources(self, context, pkg_dict, user_obj):
             Uses context from the package_create call and the created dataset """
 
         kobo_asset_id = pkg_dict.get('kobo_asset_id')
+
+        if kobo_asset_id.startswith('test_'):
```
amercader, line 122: This is a bit ugly. It looks like it could be replaced by mocking the `create_kobo_resources()` method in the relevant tests right?

amercader (commented):
@avdata99 Overall the logic looks good, there some changes needed though

ckanext/unhcr/fanstatic/theme.js
avdata99: theme.js is the destination while we run `make assets`. This file always include all code from other js files @amercader

ckanext/unhcr/templates/scheming/package/snippets/resource_form.html
avdata99: I know. I prefer ID because is supposed to be unique
Also [W3C says](https://www.w3.org/TR/html4/interact/forms.html#edef-FORM):
```name = This attribute names the element so that it may be referred to from style sheets or scripts. Note. This attribute has been included for backwards compatibility. Applications should use the id attribute to identify elements.```

ckanext/unhcr/actions.py
avdata99: Done

ckanext/unhcr/actions.py
avdata99: Done

ckanext/unhcr/tests/pt/controllers/test_resource.py
avdata99: Done

ckanext/unhcr/tests/pt/controllers/test_resource.py
avdata99: Done

ckanext/unhcr/tests/pt/controllers/test_resource.py
avdata99: Done all changes [here](https://github.com/okfn/ckanext-unhcr/commit/80b8c286e447c18854375af4d3ab1fe171af1700)

ckanext/unhcr/kobo/kobo_dataset.py
avdata99: Done

Issue #669, opened 2021-08-19
Test deployed but empty KoBo surveys


avdata99:
The client mentioned some 500 errors probably related to empty surveys (deployed but with no data)
Ensure the export creation works fine for every survey or notify the user about this specific error. Avoid non explained 500 error

PR #671, opened 2021-08-20
Use Kobo survey submission count to validate

avdata99:
Related to #669 

This PR:
 - Use submission count to validate before creating KoBo datasets
 - Save the submission count for future reference
 - Shows submission count at the survey list

Selection_341

Selection_342


ckanext/unhcr/kobo/kobo_dataset.py
```diff
@@ -165,9 +170,14 @@ def create_data_resources(self, context, pkg_dict, survey):
...
+            survey.load_asset()  # required to get the dates
             # we can use the inacurate creation and modification dates from the survey
-            date_range_start = survey.asset.get('date_created')
-            date_range_end = survey.asset.get('date_modified')
+            # those are DateTimes and we need just Dates
+            date_created = parse_date(survey.asset.get('date_created'))
+            date_modified = parse_date(survey.asset.get('date_modified'))
+
+            date_range_start = date_created.strftime('%Y-%m-%d')
+            date_range_end = date_modified.strftime('%Y-%m-%d')
```
avdata99, line 180: This fixes the actual 500 error before this PR. We are not able to read the survey metadata and after that we are using `datetimes` instead `dates`.

ckanext/unhcr/actions.py
```diff
@@ -71,14 +72,22 @@ def package_update(up_func, context, data_dict):
 def package_create(up_func, context, data_dict):
     """ Create resources for KoBo assets """
 
+    # Check if the survey it's ready to import
```
amercader, line 75:
```suggestion
    # Check if the survey is ready to import
```

ckanext/unhcr/actions.py
```diff
@@ -71,14 +72,22 @@ def package_update(up_func, context, data_dict):
...
+    kobo_asset_id = data_dict.get('kobo_asset_id')
+    if kobo_asset_id:
+        user_obj = model.User.by_name(context['user'])
+        kd = KoboDataset(kobo_asset_id)
+        kobo_api = kd.get_kobo_api(user_obj)
+        survey = KoBoSurvey(kobo_asset_id, kobo_api)
```
amercader, line 81: This should definitely not be changed now but it seems a bit strange to have: `KoboDataset -> KoboAPI -> KoboSurvey`
Perhaps I would expect an API closer to this:
```python

client = KoboClient(user_obj)    # Raises error if token not ok

client.get_dataset(asset_id)
client.get_survey(asset_id)
# etc...

```

Again, no need to change it now

ckanext/unhcr/kobo/kobo_dataset.py
```diff
@@ -120,6 +123,7 @@ def create_kobo_resources(self, context, pkg_dict, user_obj):
         kobo_asset_id = pkg_dict.get('kobo_asset_id')
 
         if kobo_asset_id.startswith('test_'):
+            logger.warning('Skipping test dataset because test_ prefix')
```
amercader, line 126: See https://github.com/okfn/ckanext-unhcr/pull/666#discussion_r695675471

ckanext/unhcr/templates/package/resource_read.html
```diff
@@ -91,6 +91,10 @@ <h2>{{ _('KoBo information') }}</h2>
           <th scope="row">{{ _('KoBo export ID') }}</th>
           <td>{% if res.kobo_details %}{{ res.kobo_details.kobo_export_id }}{% else %}n/a{% endif %}</td>
         </tr>
+        <tr>
+          <th scope="row">{{ _('Submission count') }}</th>
+          <td>{% if res.kobo_details %}{{ res.kobo_details.kobo_submission_count }}{% else %}n/a{% endif %}</td>
```
amercader, line 96: I don't know if it makes sense to show this value here as it will eventually go out of sync (I assume we are not re-querying this value on each resource page load)

ckanext/unhcr/tests/factories.py
```diff
@@ -73,7 +73,25 @@ def _generate_plugin_extras(user):
...
 class InternalUser(factories.User):
 
     email = factory.LazyAttribute(_generate_internal_email)
     plugin_extras = factory.LazyAttribute(_generate_plugin_extras)
+
+
+class InternalKoBoUser(factories.User):
+
+    email = factory.LazyAttribute(_generate_internal_email)
+    plugin_extras = factory.LazyAttribute(_generate_kobo_extras)
```
amercader, line 97:
```suggestion
class InternalKoBoUser(InternalUser):

    plugin_extras = factory.LazyAttribute(_generate_kobo_extras)
```

ckanext/unhcr/tests/pt/controllers/test_resource.py
```diff
@@ -13,12 +14,18 @@
 @pytest.mark.usefixtures('clean_db', 'unhcr_migrate', 'with_request_context')
 class TestResourceViews(object):
 
-    def setup(self):
+    @mock.patch('ckanext.unhcr.kobo.api.KoBoSurvey.get_total_submissions')
```
amercader, line 17: I don't like mocking things in `setup()` methods as it can affect individual tests in ways that are hard to debug. Better to move whatever needs mocking to their own test and do it there.

amercader (commented):
This looks good @avdata99, only some minor changes needed

ckanext/unhcr/actions.py
avdata99: Moved to a new issue, #674

ckanext/unhcr/tests/pt/controllers/test_resource.py
avdata99: Changed

Issue #668, opened 2021-08-19
Skip ClamAV check for empty files


avdata99:
First step with KoBo resources is to create empty files. 
We are getting email from ClamAV saying `Error performing Clam AV Scan`
This is probable because of these files.
Check if we really need to create empty files (maybe skipping the `upload_not_empty` validator)

avdata99:
2 possible solutions:
 - Stop scanning temporary files: #672 (maybe could be better)
 - Ensure file have content and is not deleted: #673

amercader:
@avdata99 definitely go for option 1. We don't want these resources to even get to the Clam AV service. But instead of passing the `skip_scan` property on the data_dict, pass it on the `context` (so it's not exposed on the API)

PR #672, opened 2021-08-23
Allow skipping ClamAV

avdata99:
Related to #668 
This PR allow using `skip_clamav_scan` context param to skip unnecessary check on fake KoBo resources


amercader:
See https://github.com/okfn/ckanext-unhcr/issues/668#issuecomment-905466191

ckanext/unhcr/kobo/kobo_dataset.py
```diff
@@ -177,8 +177,8 @@ def create_data_resources(self, context, pkg_dict, survey):
             date_range_end = date_modified.strftime('%Y-%m-%d')
 
         resources = []
-        # create empty resources to be downloaded later
-        f = tempfile.NamedTemporaryFile()
+        # create empty resources to be updated later
+        f = tempfile.NamedTemporaryFile(delete=False)
```
amercader, line 181: Why do we need to keep the temp files? Ideally we should avoid this otherwise we need to take care of cleaning up

amercader (commented):
This looks good @avdata99 , just a question regarding temp files

ckanext/unhcr/kobo/kobo_dataset.py
avdata99: Yes, we don't really need to preserve that files
I think I added to avoid errors after creating this resource but maybe ClamAV (now skipped) was the only error. It's working fine after removing the `delete` param

Changed [here](https://github.com/okfn/ckanext-unhcr/pull/672/commits/af7b4d862aea63f12056b3a624592ce1bf027e8c)

amercader (approved):

PR #673, opened 2021-08-24
Ensure temp file have content and is not deleted

avdata99:
Related to #668.
Simpler than #672 approach to stop getting ClamAV error notifications

2 possible solutions for ClamAV error:
 - Stop scanning temporary files: #672 (maybe could be better)
 - Ensure file have content and is not deleted: This PR


avdata99:
Closed in favor of #672

PR #688, opened 2021-09-07
Skip big files from ClamAV

avdata99:
Related to #676 
Related to Docker PR [docker#64](https://github.com/okfn/docker-ckan-unhcr-aws/pull/64) 

This PR skip analyzing big files with ClamAV


ckanext/unhcr/actions.py
```diff
@@ -621,6 +621,19 @@ def scan_submit(context, data_dict):
     except toolkit.ObjectNotFound:
         return False
 
+    file_size = resource_dict.get('size', 0)
```
amercader, line 624: Can we rely on this value being present? IIRC it's up to the storage backend to provide the `size` value. I know that the local storage backend provides it but on the live environments we use ckanext-s3filestore, which I'm not sure does.

avdata99: No 100% sure if the value will be present.
I checked at CKAN core for [create](https://github.com/ckan/ckan/blob/ckan-2.9.3/ckan/logic/action/create.py#L312) and [update](https://github.com/ckan/ckan/blob/ckan-2.9.3/ckan/logic/action/update.py#L298) and I see we try to set this value before upload.
Do you this won't work on UAT or production @amercader?

avdata99: @amercader Keitaro added `filesize` support to s3filestore [here](https://github.com/keitaroinc/ckanext-s3filestore/commit/9187eec40bdecd31ea6e83f7de3859dfd5483840)

avdata99:
Moved PR to Release-3.1.4

amercader (approved):

Issue #698, opened 2021-09-16
Fix data preview for KoBo resources


avdata99:
After a KoBo dataset is created the preview for CSV and XLS files are not working (JSON is working fine)

PR #701, opened 2021-09-21
Fix preview for KoBo resources

avdata99:
Related to #698 

This PR:
 - Fixes the conflict with underscore prefixed field names for datastore
 - Fixes an unrelated JS error

Selection_407

Selection_406

Selection_401


amercader:
@avdata99 This is a good stopgap measure but I really thing CKAN core behaviour here is wrong, and underscore fields should be allowed and rendered. We should patch our local CKAN install (and eventually upstream CKAN) to allow this.
The `_id` field is a special case because that might affect other things. I'll write an issue in the main CKAN repo

In any case, let's merge this for now so at least there is something to preview, but let's display a notice:

Screenshot 2021-09-29 at 11-39-56 Data in IDP situations - Survey csv data - RIDL

This is done in `ckanext-unhcr/ckanext/unhcr/templates/package/snippets/resource_view.html`:

```
{% ckan_extends %}

{% block resource_view %}

  {% if resource['kobo_type'] and resource_view['view_type'] in ['recline_view', 'datatables_view'] %}
  <div class="alert alert-warning">Note: KoBo fields starting with underscore are prefixed with a <i>kobo</i>string.</div>
  {% endif %}

  {{ super() }}

{% endblock %}
```

amercader (approved):

avdata99:
@amercader I created #706 to follow up this and remove the notice when CKAN core fix is ready.

PR #722, opened 2021-10-12
Release 3.1.4d

avdata99:
- Move questionnaire download as job and 
 - update test
 - Include questionnaire in status detection


avdata99:
Manually merged

Issue #732, opened 2021-10-28
Keep track of geography import runs


avdata99:
Client request
Record the dates and result for each `ckan unhcr import-geographies` runs

Maybe we can also keep track of expired users and other internal tasks and show all in the UI to sysadmins

PR #734, opened 2021-11-01
Record system activities

avdata99:
Related to #732 

This PR create the concept of _system activities_ to track activities at the application level.
This allows to record logs for commands (this PR includes `import-geographies` and `expire-users`)
Perhaps useful as a CKAN core contribution.

System activities list

image

Activity details

image


ckanext/unhcr/templates/admin/system_activities.html
```diff
@@ -0,0 +1,90 @@
...
+      <th>Date</th>
+      <th>System activity</th>
+      <th>Description</th>
+      <th>More info</th>
+    </tr>
+  </thead>
+  <tbody>
+    {% for activity in activities %}
+    <tr>
+      <td>{{ activity.timestamp }}</td>
```
amercader, line 20:
```suggestion
      <td>{{ h.render_datetime(activity.timestamp, with_hours=True) }}</td> 
```

ckanext/unhcr/helpers.py
```diff
@@ -1176,3 +1176,15 @@ def get_resource_value_label(field_name, resource, dataset_type='dataset'):
...
+
+
+def get_system_activities():
+    activities = model.Session.query(
+        model.Activity
+    ).filter_by(
+        activity_type="system activity",
+    ).order_by(
+        model.Activity.timestamp.desc()
+    ).all()
```
amercader, line 1188: Let's add a safety limit here, 500, 1000 whatever you consider.

amercader (approved):
This looks great @avdata99 !

amercader:
@avdata99 I wonder if it's also worth adding a `last_modified` timestamp field in the database model now before importing the geographies in the live environments. This would be set whenever the geography is created or updated (if that is possible). This can be done separately of course.

avdata99:
> @avdata99 I wonder if it's also worth adding a `last_modified` timestamp field in the database model now before importing the geographies in the live environments. This would be set whenever the geography is created or updated (if that is possible). This can be done separately of course.

@amercader I discovered our `geography` table is not ready for migrations.
The migration with `Geography.__table__.create()` will be marked as done even if we add more fields.
So I propose [this](https://github.com/okfn/ckanext-unhcr/commit/d6ba2838ec8c2d0ce2d58c84bb4aed0995f9ba55) so we will be able to add more functionality.

ckanext/unhcr/helpers.py
avdata99: Done [here](https://github.com/okfn/ckanext-unhcr/commit/bd6047db3a034e8d07089b5592d7abf427880435)

avdata99:
> > @avdata99 I wonder if it's also worth adding a `last_modified` timestamp field in the database model now before importing the geographies in the live environments. This would be set whenever the geography is created or updated (if that is possible). This can be done separately of course.
> 
> @amercader I discovered our `geography` table is not ready for migrations. The migration with `Geography.__table__.create()` will be marked as done even if we add more fields. So I propose [this](https://github.com/okfn/ckanext-unhcr/commit/6033a1939406b2c3e335ccd2013252861bc09415) so we will be able to add more functionality.

I sent this PR to re-review because of this migration change. I'm pretty confident it's ok, but it's a _retroactive_ migration @amercader

amercader (approved):
This looks good @avdata99 . I updated the code and restarted the container, triggering the new migration and the table was modified correctly.

Issue #740, opened 2021-11-11
Private label even if you have permission


avdata99:
Show the `Private` label for resources even if the user have permission.
In that case, we can show the label in a different way.

PR #747, opened 2021-11-12
Show private label even if user have permissions

avdata99:
Related to #740 

Stop hiding the `private` label when user have permission
In those cases, mark as private but identify the resource as available for the user

Proposal:

Selection_505


amercader (commented):
I'm OK-ish with this approach but let's deploy it and check with the client. In any case I think that the label should have a `title` attribute so on hover you get a little help text:
* "You are not allowed to download this file"
* "You can download this file, but is private to the rest of non-authorized users"

Issue #738, opened 2021-11-05
Show users that geographies are being searched


amercader:
Right now when you are typing some characters to search for a geography you don't have any indication a search is being performed. In fact it says "No results found" for a while

![Screenshot from 2021-11-05 11-25-24](https://user-images.githubusercontent.com/200230/140496620-3cc1e39c-3c78-4cb1-a171-13d48514bb9d.png)

We should show a spinner or just change the text to "Searching...", which AFAICT is what default select2 does: https://select2.org/data-sources/ajax

avdata99:
I have a proposal here: [CKAN#6554](https://github.com/ckan/ckan/pull/6554)

If it's OK I can prepare the patch file @amercader

PR #751, opened 2021-11-23
Avoid short query to a heavy API call

avdata99:
Related to #738 and requires [CKAN#6554](https://github.com/ckan/ckan/pull/6554)

Avoid short queries for a heavy API (we have 250k geographies)


amercader:
This looks good @avdata99 . You need to add the core select2 patch though right?

PR #756, opened 2021-12-07
Set fallback admin when user deleted (e.g. expire)

avdata99:
Related to #340 

This feature was proposed in #604 and #620 but (by mistake probably) it was never merged to `master`

This PR:
 - Override `user_delete` to detect possible orphan containers
 - Set up a fallback admin if necessary.
 - Test


ckanext/unhcr/jobs.py
```diff
@@ -57,6 +59,51 @@ def process_dataset_on_update(package_id):
...
+
+
+def process_last_admin_on_delete(org_id):
+    """ The only admin of a data container was deleted a we need to assign
+        a fallback admin and notify sysadmin """
+
+    new_admin = _define_new_admin_member(org_id)
+    context = {'model': model, 'job': True, 'ignore_auth': True, 'user': new_admin.name}
+
+    extra_msg = 'You are being assigned as a fallback admin because previous admin user was deleted'
```
amercader, line 93:
```suggestion
    extra_msg = 'You have been assigned as admin of this Data Container because the previous admin user was deleted'
```

amercader (approved):
Looks good!

ckanext/unhcr/jobs.py
avdata99: fixes [here](https://github.com/okfn/ckanext-unhcr/pull/756/commits/f86ee50a2f565c8ccba83b1dfa52d741568b73ae)

Issue #287, opened 2020-04-27
Improve handling of deleted datasets


chris48s:
Brief for this one is:

> Make it easier to keep track of deleted datasets and associated activities

This one needs fleshing out a bit more before I can tackle it

chris48s:
When a dataset is deleted, CKAN does not delete its record from the database, just sets its status to deleted (unless it is then purged from the admin interface). Activities for deleted datasets are also kept. This is in line with the requirements of keeping track of all changes, but it’s difficult to manage these deleted datasets from the UI.

* Unless you know the actual URL there is no way to go back to a previously deleted dataset from the UI, as it doesn’t appear in searches any more (actually the deletion appears on the global feed in the user’s dashboard, but its not permanent)
* There is no way from the UI to “re-activate” a deleted dataset

chris48s:
Requirements confirmed:

- page where we can discover deleted datasets
- ability to restore/un-delete deleted datasets

PR #757, opened 2021-12-08
Show deleted pakcages in search

avdata99:
Related to #287 

Also related with:
 - [CKAN#6591](https://github.com/ckan/ckan/pull/6591)
 - [DOCKER-UNHCR#77](https://github.com/okfn/docker-ckan-unhcr-aws/pull/77)

Notes:
 - CKAN core don't index deleted datasets ([here](https://github.com/ckan/ckan/blob/83eec85e298bc4b374cd6a80d9f7c050c20da4bc/ckan/lib/search/__init__.py#L169) and [here](https://github.com/ckan/ckan/blob/83eec85e298bc4b374cd6a80d9f7c050c20da4bc/ckan/lib/search/__init__.py#L104-L105))
 - Admin Trash read deleted packages directly from DB [here](https://github.com/ckan/ckan/blob/83eec85e298bc4b374cd6a80d9f7c050c20da4bc/ckan/views/admin.py#L127-L128)

This PR:
 - Start showing deleted datasets in search result
 - Show trash datasets from search index instead DB


avdata99:
@amercader 

I see two path here:

1 - Changing CKAN core (add the `ckan.search.allow_deleted`)
 - Upgrade the `rebuild` command for search index to optionally allow including `deleted` datasets
 - Stop removing packages from the search index after deleting them (probably at `dispatch_by_operation` function). 

2 - Not changing CKAN core
 - Add a `search deleted` button (probably at the dataset list). Run this query directly from database.
 - Probably add a search box at the `ckan-admin/trash` section

image

I plan to start with **2** because I assume is simpler for our use case.

amercader:
@avdata99 I'm happy to go with 2 but we really should test how it would scale over time (say with a couple thousand deleted datasets) as it's easier to shoot ourselves in the foot with DB queries than with Solr ones.

amercader (approved):
Great work @avdata99 , this works well but in terms of user experience I think we should not include deleted datasets by default. This is going to cause confusion with the number of results returned not matching counts on other parts of the application, and potentially, as time progresses we could end up with a lot of deleted datasets in the results.

So I think we need users to select a checkbox below the search input that says "Include deleted datasets" (which potentially could only be shown to sysadmins or container admins).
This should be easy to implement. Any parameter prefixed with `ext_` is included in `search_params['extras']`, eg https://localhost:5000/dataset/?ext_include_deleted=True

This could be used in `before_search()` to apply the filter or not:
```diff
diff --git a/ckanext/unhcr/plugin.py b/ckanext/unhcr/plugin.py
index 5f0407a2..38ba5c53 100644
--- a/ckanext/unhcr/plugin.py
+++ b/ckanext/unhcr/plugin.py
@@ -365,7 +365,7 @@ class UnhcrPlugin(
             return search_params
 
         # include deleted datasets
-        if blueprint in ['dataset', 'package']:
+        if blueprint in ['dataset', 'package'] and toolkit.asbool(search_params.get('extras', {}).get('ext_include_deleted')):
             search_params['include_deleted'] = True
 
         elif blueprint in ['organization', 'data-container'] and view != 'edit':
```


For the UI, on `search_form.html`:

```diff
diff --git a/ckanext/unhcr/templates/snippets/search_form.html b/ckanext/unhcr/templates/snippets/search_form.html
index c8226895..1d9e9c62 100644
--- a/ckanext/unhcr/templates/snippets/search_form.html
+++ b/ckanext/unhcr/templates/snippets/search_form.html
@@ -3,6 +3,15 @@
 {% block search_input %}
     <div class="search-input control-group {{ search_class }}">
       <input type="text" class="search" name="q" value="{{ query }}" autocomplete="off" placeholder="{{ placeholder }}">
+
+         {% if form_id == 'dataset-search-form' %}
+           <div class="control-include-deleted">
+               <input type="checkbox" name="ext_include-deleted"{% if h.gdx_get_include-deleted_value() %} checked=checked {% endif %}data-module="checkbox-switcher" />
+               <label for="ext_include-deleted">Inlcude deleted datasets</label>
+           </div>
+       
+           {% asset 'gdx/modules/checkbox-switcher' %}
+  {% endif %}
       {% block search_input_button %}
       <button type="submit" value="search" aria-label="{{_('Submit')}}">
         <i class="fa fa-search"></i>

```

The `checkbox-switcher.js` module is just:

```js
this.ckan.module('checkbox-switcher', {                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                     
  initialize: function () {                                                                                                                                                                                                                                                          
    var _this = this;                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                     
    var form = this.el.closest('form');                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                     
    this.el.on('change', function () {                                                                                                                                                                                                                                               
      form.submit();                                                                                                                                                                                                                                                                 
    });                                                                                                                                                                                                                                                                              
  }                                                                                                                                                                                                                                                                                  
});
```
Does this makes sense? I've implemented this in GDX so can provide more code if needed

Issue #758, opened 2021-12-10
Check KoBo permission before import/update surveys


avdata99:
From client:

 - For newly imported projects: Only users who own the project or have Manage project permission in KoBo should be able to import into RIDL. 
 - In the table of KoBo surveys, we should show a message about insufficient permissions instead of the Import button for those that do not have these.
 - For update of already imported projects: Only users who own the project or have Manage project permission in KoBo and have at least Editor access to the dataset on RIDL should be able to update (re-import) the data. Same as above for the 'Update' button in the table.

PR #759, opened 2021-12-10
Handle KoBo survey permissions

avdata99:
Related to #758 

This PR:
 - Hide import/update buttons from surveys list if user does not have manage permissions
 - Hide Update from KoBo from Dataset package if user does not have manage permissions
 - Updated code as KoBo changed API responses

New survey list

image

Package read page: Only show the update button on dataset where current user is KoBo manager

image


ckanext/unhcr/kobo/api.py
```diff
@@ -63,11 +64,37 @@ def get_surveys(self, limit=100, force=False, all_pages=True):
                     return surveys
                 next_url = response.get('next')
             self.surveys = surveys
+
+        # add permission information
+        # detect if user has permission to manage the survey
+        for survey in self.surveys:
```
amercader, line 70: Maybe it's safer to default `self.surveys` to `[]` instead of `None` otherwise this code will fail if there are none

avdata99: `surveys = None` means _you never get the surveys list_
`surveys = []` means _you've already searched for surveys and there aren't any_

This tries to avoid requesting assets every time the function is called. It will only be called once (unless `force` is used)
At this point, is not possible for `self.surveys` to be `None`. If I'm wrong we can add a new PR to fix @amercader

Issue #770, opened 2022-01-27
Make the ‘Geographies field’ mandatory


avdata99:
Fix the schema

PR #777, opened 2022-01-27
Make the Geographies field required

avdata99:
Related to #770 

This PR:
 - Fixes the schema to make `geographies` field required (not for deposited datasets)
   - Fill empty Geographies with a new default called `UNSPECIFIED`
 - Fixes UI _required_ **\*** mark.
 - Update tests

Also required
 - [x] [DDI PR](https://github.com/okfn/ckanext-ddi/pull/10) is also required
 
Note: Users will be invited to fill `UNSPECIFED` geographies

image


ckanext/unhcr/schemas/presets.json
```diff
@@ -14,13 +14,12 @@
     {
       "preset_name": "geographies",
       "values": {
-        "validators": "ignore_missing geographies_validator",
+        "validators": "scheming_required geographies_validator",
```
avdata99, line 17: This will converted to `ignore_missing` or `not_empty` depending on the value of `required`

amercader (commented):
This looks good @avdata99 but I wonder if it's not more appropriate to use `UNSPECIFIED` as a fallback geography code.

Every schema change that adds a mandatory field should involve a migration on the current datasets to fill the existing fields otherwise edits will fail. Do we need to do this with an API script on UAT and Prod?

avdata99:
> This looks good @avdata99 but I wonder if it's not more appropriate to use `UNSPECIFIED` as a fallback geography code.

Changes [here](https://github.com/okfn/ckanext-unhcr/commit/8a5bbe803e5842515b8a1c0b42af60f4ce7333b5)

> Every schema change that adds a mandatory field should involve a migration on the current datasets to fill the existing fields otherwise edits will fail. Do we need to do this with an API script on UAT and Prod?

We need first to create this fallback Geography before using it via API. Otherwise the `geographies_validator` will raise an error. A migration to update all datasets will be enough.
 
When a user edits an old dataset without Geography, they will be forced to define the Geography. If we add a migration to set all datasets to 'UNSPECIFIED', the edits will not fail, but the user will not realize that he needs to define a Geography.

Do you prefer to skip this error @amercader?

avdata99:
@amercader Update after client meeting ([here](https://github.com/okfn/ckanext-unhcr/pull/777/commits/c4e08be2926c11429b7a6a1743f2f908e85ac0b1))

 - Fill empty Geographies to `UNSPECIFIED`
 - Show this to users and try to get them to fix it.

image

ckanext/unhcr/templates/scheming/form_snippets/geographies.html
```diff
@@ -14,7 +28,7 @@
       data-module-label="name"
       data-module-minimum-input-length="2"
       value="{{ data[field.field_name]|join(',') }}"
-      placeholder="Select geographies"
+      placeholder="{% if unspecified %}PLEASE FILL THIS FIELD, AT LEAST THE COUNTRY{% endif %}"
```
amercader, line 31:
```suggestion
      placeholder="{% if unspecified %}Please provide at least country-level geographic coverage{% endif %}"
```

amercader (approved):
@avdata99 some minor style checks and we are good to go

Issue #775, opened 2022-01-27
Duplicated Geographies


avdata99:
Found duplicated values (two admin2 Kampala). 
Client checked and it is not duplicated in the GIS database.

![image](https://user-images.githubusercontent.com/3237309/151374998-ac2aa7d8-039a-48bb-a65d-d935ac049eb8.png)

Check why this duplicate item appeared

PR #782, opened 2022-02-01
Inactivate ALL geographies before importing them

avdata99:
Related to #775 
We are seeing duplicate geographies because they change the pcode (which we assume as a unique PK)

Previous _Kampala_ pcode was `23UGA001008` and now it's `24UGA001008`

Are those 2 characters at the beginning the version code?

This PR inactivates ALL geographies before re-importing them because if not, we will end up having multiple activated geographies (relative to the same place with a different `pcode`).
It was probably a mistake to only inactivate countries in the previous version

A message were sent to the client to define if preserve or drop the version number for geographies.


amercader:
I'm not familiar with all the import logic so I'm not sure what the implications on inactivating all geographies has, but in theory it seems like it would be safe.

According to Chris' notes [here](https://github.com/okfn/ckanext-unhcr/issues/618#issuecomment-867666653) the pcode that we are using should be unversioned. Has this changed at some point?

avdata99:
I tested locally the _Kampala_ case.
By making this change, all datasets using old geography version, will continue working and showing the geography as we expect.
After getting the new geography version, we inactivate old version are only the new version is available for new datasets. Old version will no be available to use but remains visible in datasets metadata.

I think this is the expect behavior.
I asked to the client and they will ask to the GIS team to confirm @amercader.

avdata99:
> According to Chris' notes [here](https://github.com/okfn/ckanext-unhcr/issues/618#issuecomment-867666653) the pcode that we are using should be unversioned. Has this changed at some point?

`pcode` seems versioned for `admin1` and `admin2` geographies in Chris' notes @amercader
They use 2 digits. I assume those are for the version number.

amercader (approved):
This looks good to me @avdata99 . Just to be sure, have you tried that running the import job after these changes only adds new records for geographies that were not present before? ie existing geographies records are re-activated, they don't get a new one

So if the initial state was

| pcode | active |
| --- | --- |
| A1 | true |
| B1 | true |
| C1 | true |

And there is a new version of the A geography (A2), after the import job we will end up with:

| pcode | active |
| --- | --- |
| A1 | false |
| A2 | true |
| B1 | true |
| C1 | true |

avdata99:
> This looks good to me @avdata99 . Just to be sure, have you tried that running the import job after these changes only adds new records for geographies that were not present before? ie existing geographies records are re-activated, they don't get a new one

Yes, I tested this and works fine @amercader

Issue #773, opened 2022-01-27
Improve geography search


avdata99:
Enable better search by allowing users to search within a specific country or Admin areas

Client notes:

![image](https://user-images.githubusercontent.com/3237309/151373104-3621f59e-f9f5-49d4-a0bf-164028d97cf9.png)

_“I don't know what lib they use for the autocomplete (looks like Bootstrap Typeehead) but this is a good one: https://github.com/jeancroy/FuzzySearch_

The other idea was that if the user types Uganda, maybe everything that’s associated to Uganda could show up in the search bar, e.g. Rhino Camp. Possibly confusing and overwhelming.

PR #783, opened 2022-02-01
Allow similarity search for geographies

avdata99:
Related to #773 

Based on [this tutorial](http://www.postgresonline.com/journal/archives/169-Fuzzy-string-matching-with-Trigram-and-Trigraphs.html)

This PR:
 - Uses similarity search for `geography_autocomplete` API endpoint (requires to add the `pg_trgm` PostgreSQL extension): **is this a good idea?**

Pending:
 - Define a way to search inside countries or admin areas

image


amercader:
> is this a good idea?

It sounds like one! but the only way to be sure is to test it with some real examples and ensure that exact matches are not affected (ie we don't get worse results). It's probably a good thing to add a few tests including the suggested `M'bera` example (that will involve setting this pg extension in the test env of course)

One thing that definitely seems off is creating the index during the [database migration](https://github.com/okfn/ckanext-unhcr/pull/783/files#diff-476f737e797589b1806e040ea8af4320473dcedad6b3a0320a34d3f931ea9539R25). This should be done after the scheduled update of the geographies we do once a week/month, otherwise it will be out of sync. +1 on using GIN as we don't care about build speed.

> Define a way to search inside countries or admin areas

I don't think that's a good idea to be honest, or it should be restricted to the lower admin areas, and even so there will likely be too many results returned

.github/workflows/test.yml
```diff
@@ -33,7 +33,7 @@ jobs:
           --health-retries 5
 
     env:
-      CKAN_SQLALCHEMY_URL: postgresql://ckan_default:pass@postgres/ckan_test
+      CKAN_SQLALCHEMY_URL: postgresql://postgres:pass@postgres/ckan_test
```
avdata99, line 36: A psql superuser required to create DB extensions
`ckan:ckan` is not working here.

avdata99:
> > is this a good idea?
> 
> It sounds like one! but the only way to be sure is to test it with some real examples and ensure that exact matches are not affected (ie we don't get worse results). It's probably a good thing to add a few tests including the suggested `M'bera` example (that will involve setting this pg extension in the test env of course)

Using `similarity>0.35`

image
image

Using `similarity>0.45`

image

Not using `similarity`

image

> One thing that definitely seems off is creating the index during the [database migration](https://github.com/okfn/ckanext-unhcr/pull/783/files#diff-476f737e797589b1806e040ea8af4320473dcedad6b3a0320a34d3f931ea9539R25). This should be done after the scheduled update of the geographies we do once a week/month, otherwise it will be out of sync. +1 on using GIN as we don't care about build speed.

Maybe I'm missing something here. The index must be created only one time (if not we get a `ERROR:  relation "idx_geos_trgm_gin_gis_name" already exists`
Are you referring to an index rebuild?
I assume this will run automatically every time we save a Geography

> 
> > Define a way to search inside countries or admin areas
> 
> I don't think that's a good idea to be honest, or it should be restricted to the lower admin areas, and even so there will likely be too many results returned

Maybe when we add a map to search geographies this will be simpler.

avdata99:
@amercader some fixes ([here](https://github.com/okfn/ckanext-unhcr/pull/783/commits/0edf0176edefabe76b648bef2e65c218b3b1ee35)):
 - The results are now sorted by similarity
 - Literal search (`ilike`) was removed, only `similarity`works now.
 - If we have many results, we add a group called _"Best matches"_ and show it at the top. To group by layer type it's a challenge but this way we have best results and group
 - If we have < 15 results, we doesn't show the _"Best matches group"_  

image

(scrolling previous results)

image

avdata99:
@amercader another change: we search using `similarity` > 0.3 but we now send <0.35 values to the bottom to a new section called _More results ..._
I think this is better than previous version

We are now searching at the `gis_name` field. This not include parent geographies.
So `Córdoba, Spain` and `Córdoba, Argentina` are no different for this search. It doesn't matter if we search for _Córdoba Argentina_, those two will have the same `similarity`.

One more change we can do is to add Full text search through Gis name + [all parent gis names] with a `tsvector` field or something similar. Maybe this will be too much at this point.

.github/workflows/test.yml
amercader: I think we should not run the test as the superuser. Instead of creating the extension in [`create_tables()`](https://github.com/okfn/ckanext-unhcr/pull/783/files#diff-9697c4b8b8db1fd87a216880bfa8c8f746aa04252341697d19d9090b765dbb71R221), create it explicitly with the `postgres` user in `init-db.bash`:

```bash
psql -U postgres -c "CREATE EXTENSION IF NOT EXISTS pg_trgm"
```
This way we don't need to modify the DB user that CKAN uses

amercader (approved):
@avdata99 see comments. None of them are blockers so you can do them on a later PR if necessary

.github/workflows/test.yml
avdata99: @amercader I can't create the extension using `init-db.bash` because the DB requires password for this command (`psql: error: fe_sendauth: no password supplied` error). `sudo` is also not useful. It seems we need to change the `pg_hba.conf` file and this is not available in the test environment. 

`ckanext-spatialuses the same approach, it [creates the extension](https://github.com/ckan/ckanext-spatial/blob/master/ckanext/spatial/tests/conftest.py#L16) for testing inside the pytest fixture.

I will merge this and then check how to deal with this in the docker container for UAT

PR #784, opened 2022-02-04
Script to analyze AWS Cloudwatch insight logs

avdata99:
This PR adds a simple (and probably incomplete) script to analyze AWS CloudWatch Insight logs

Private files: [Original logs](https://docs.google.com/spreadsheets/d/1Ue5rSomT8848h8p7W9DT5jyd9n2n1jEYw_4K7dwE0PI/edit#gid=1511295287) + [final results](https://docs.google.com/spreadsheets/d/1Xb6ysZ4FRFsdIMNqumKjVs_xEjl0N3AWaMz5Cglo27Y/edit#gid=415366933)

Final results were sent to the client as a _pending to validate_ results


amercader:
That is actually a quite decent approach if we want something quick and dirty, definitely good enough for a one-off report. In terms of presentation I think that rather than individual searches timestamped it's probably more useful to have aggregates of the same query to see which ones are more popular

Issue #786, opened 2022-02-08
Cache for KoBo requests


avdata99:
After #785 we realize we need to cache KoBo requests results.
The most importan content to cache is the list of surveys at `/kobo/surveys`

PR #787, opened 2022-02-08
Add Redis cache for KoBo responses

avdata99:
Related to #786 

This PR:
 - Use CKAN Redis instance as a Cache for all KoBo responses


ckanext/unhcr/kobo/api.py
```diff
@@ -13,17 +16,42 @@ class KoBoAPI:
     """ About KoBo API
         https://support.kobotoolbox.org/api.html
     """
-    def __init__(self, token, kobo_url):
+    def __init__(
+        self, token, kobo_url, cache_prefix,
+        redis_host='redis', redis_port=6379, redis_db=0, cache_seconds=600
```
amercader, line 21: I don't think the redis conn parameters and the cache duration should be constructor parameters. Instead we can just use `config["ckan.redis.url"]` or if you think we might need another instance specifically for this you can create a dedicated config option (eg `config["ckanext.unhcr.redis.url") (I don't think we need to)

The cache seconds settings should be a config setting too. And both should be used in the `__init__()` method to set whatever internal property you need

amercader (approved):
This looks great @avdata99 just some comment about config settings

ckanext/unhcr/kobo/api.py
avdata99: I was trying to keep this module ready to move as an external python library. 
It's done now, we use CKAN settings

Issue #791, opened 2022-03-10
Add geographies as filterable facets


avdata99:
Add Geographies as a filter at left in the list datasets.
Show only most used.
Include places and parents

PR #792, opened 2022-03-10
Adding "geographies" field as facet

avdata99:
Related to #791 

This PR:
 - Adds the `geographies` field as searchable facet
 - Use the MultiValued Solr field `vocab_geo` to search single geographies in datasets

image

**Search results**

image


ckanext/unhcr/plugin.py
```diff
@@ -355,8 +356,44 @@ def before_index(self, pkg_dict):
...
+        return search_results
+
     # Always include sub-containers to container_read search
     def before_search(self, search_params):
+
+        # MultiValued SOLR field "vocab_geo" contains all pcodes 
+        # and geographies names to search from
+        fq = search_params.get('fq', '')
+        if 'geographies:' in fq:
+            search_params['fq'] = fq.replace('geographies:', 'vocab_geo:')
```
avdata99, line 395: Not sure on this one. Is there a better way to do it?

amercader (commented):
You don't need to change anything in `before_search()` or `after_search()` @avdata99 , the geographies field is already being indexed correctly here:

https://github.com/okfn/ckanext-unhcr/blob/c58285dbdac09777b22af5ae916d1b2861fe5397/ckanext/unhcr/plugin.py#L296-L308

You just need to add the `vocab_geo` field in the `dataset_facets()` method:

```diff
diff --git a/ckanext/unhcr/plugin.py b/ckanext/unhcr/plugin.py
index dbc30dfe..3dcf82ac 100644
--- a/ckanext/unhcr/plugin.py
+++ b/ckanext/unhcr/plugin.py
@@ -150,6 +150,7 @@ class UnhcrPlugin(
 
         facets_dict['vocab_data_collector'] = _('Data Collector')
         facets_dict['vocab_keywords'] = _('Keywords')
+        facets_dict['vocab_geo'] = _('Geographic coverage')
         facets_dict['vocab_sampling_procedure'] = _('Sampling Procedure')
         facets_dict['vocab_operational_purpose_of_data'] = _(
             'Operational purpose of data')

```

CKAN will render the different values correctly

amercader:
@avdata99 sorry, I now see that in `before_index()` we are adding the name and the pcode of the geography to the field. We don't want the pcodes to show up on the facet list, so let's index a specific `vocab_geo_name` field that just contains the names (and does not contain the parent ones) (so just `geog.gis_name`)

Does this make sense?

avdata99:
I applied the changes [here](https://github.com/okfn/ckanext-unhcr/commit/baa2def12d3245713715330e0f71a95ef71a0faa) @amercader.
You were right about the `after_search`, it's not required.
I'm not agree with remove parent names. If I have 50 datasets from Japan cities, I want to see `Japan: 50` in the facet list.
Also, I don't feel comfortable search by names. I prefer to use `pcodes` (and the names just to show them). Is there a way to do that?

image

amercader (approved):

amercader:
@avdata99 Looks good
> I'm not agree with remove parent names. If I have 50 datasets from Japan cities, I want to see Japan: 50 in the facet list.

Fair enough. Let's deploy it like this to UAT and see how it looks like. My fear is that we will have too many options to filter from

> Also, I don't feel comfortable searching by names. I prefer to use pcodes (and the names just to show them). Is there a way to do that

pcodes are still indexed right? You should be able to search by them. Is that not happening?

avdata99:
> pcodes are still indexed right? You should be able to search by them. Is that not happening?

Yes, we can search with `pcodes`

Issue #386, opened 2020-07-15
Audit trail for access requests


chris48s:
As noted in #308

> At the moment, the access requests page only displays requests which are in the 'requested' state. Once they are approved or rejected, they basically just disappear from the front-end (although the DB record still exists). I think its ok to "launch" the feature like this, but maybe we need to think about a history/audit requirement here. Perhaps one approach would be to log an activity when a request is approved/rejected that could be surfaced in an activity stream? I haven't done anything about it either way.

> I agree we can launch without the requests log, we'll make sure to flag it to see if that is something we should work on.

> Perhaps we should store the message as part of the request for keep an audit log, but we can do that in the future (if needed)

TODO: once launched, confirm requirements with client

PR #796, opened 2022-03-16
Access requests audit

avdata99:
Related to #386 

This PR:
 - Adds a new endpoint to show all access requests (not only pending ones)
 - Only shows access request with permission for each user (sysadmins can see all requests)
 - Tests this new feature

image

Pending: 
 - Add some limit to the access request query


ckanext/unhcr/cache.py
```diff
@@ -0,0 +1,33 @@
+import json
+from redis import Redis, ConnectionPool
+from ckan.common import config
+from ckan.lib.redis import connect_to_redis, is_redis_available
+
+
+def cache_set(key, value, ttl=3600):
```
avdata99, line 7: This is a general cache for the entire CKAN instance
Do we have something like in the CKAN core? I did not find it
Would this be a good functionality for CKAN?

ckanext/unhcr/fanstatic/theme.js
```diff
@@ -497,3 +497,15 @@ $( document ).ready(function() {
 	});
 
 });
+
+$( document ).ready(function() {
+
+	$("#requests-history-list").fancyTable({
```
avdata99, line 503: The list of requests is now filterable and paginated

ckanext/unhcr/templates/user/dashboard_requests_history.html
```diff
@@ -0,0 +1,119 @@
...
+      <th>Status</th>
+      <th>More info</th>
+    </tr>
+  </thead>
+  <tbody>
+    {% for req in access_requests %}
+    <tr>
+      <td title="{{ req.timestamp }}">{{ h.render_datetime(req.timestamp, '%Y-%m-%d') }}</td> 
+      <td>
+        {% set user = h.get_cached_ckan_object_name('user', req.user_id) %}
```
amercader, line 24: I'm not convinced that using a redis cache here is not a premature optimization. If I understand correctly we are requesting the user object to get the `display name`. I'd do that in the [blueprint](https://github.com/okfn/ckanext-unhcr/pull/796/files?diff=unified&w=0#diff-e6168b0e9a70556a6b7ea7e32b2cd381cd1a4efc1acd49ab569d19db6c231234R121), modifying the requests objects to add a new `user_display_name` key with the user name that the templates need. Same on line 87

ckanext/unhcr/helpers.py
```diff
@@ -410,6 +411,34 @@ def get_default_geography():
 
 # Access requests
 
+def get_cached_ckan_object_name(object_type, object_id):
```
amercader, line 414: See comments below, I don't think a cache is needed for this particular feature, but maybe I'm missing something.

ckanext/unhcr/helpers.py
```diff
@@ -410,6 +411,34 @@ def get_default_geography():
...
+    except toolkit.ObjectNotFound:
+        result = {'id': object_id, 'title': 'NOT FOUND'}
+    else:
+        result = {'id': object_id, 'name': object['name']}
+        if object_type == 'user':
+            result['title'] = object['display_name']
+        else:
+            result['title'] = object['title']
+
+    cache_set(cache_key, result.copy())
```
amercader, line 437: This will create a shallow copy, ie only the top level keys (no `resources`, `tags`, etc). Which might be fine in this case but not for a general purpose cache.

ckanext/unhcr/helpers.py
amercader: I can't see any invalidation mechanism either, eg if someone updated the dataset/org in the meantime the cache would have a stale version.

ckanext/unhcr/cache.py
amercader: There is no functionality for a general purpose cache in CKAN core, maybe it would be a good candidate for a separate extension. If needed, let's implement a cache module in ckanext-unchr and test it here before creating a separate extension.
But I don't think we need it :slightly_smiling_face:  (see comments below)

ckanext/unhcr/cache.py
```diff
@@ -0,0 +1,33 @@
...
+    """ Get a cache if exists.
+        Return cache_exists: bool, value: Any"""
+    if not is_redis_available():
+        return False, None
+    
+    cache = connect_to_redis()
+    value = cache.get(key)
+    if value is None:
+        return False, None
+    if type(value) != str:
```
amercader, line 31: If the value is not a string you can't use `json.loads()` on it

amercader (commented):
I really like this feature @avdata99 well done. But I think we are optimizing prematurely with the cache so let's revisit this. Also left some other comments.

ckanext/unhcr/blueprints/dashboard.py
```diff
@@ -94,7 +94,44 @@ def requests():
...
 unhcr_dashboard_blueprint.add_url_rule(
     u'/requests',
-    view_func=requests
+    view_func=requests,
+    strict_slashes=False,
+)
+
+
+unhcr_dashboard_blueprint.add_url_rule(
+    u'/requests-history',
```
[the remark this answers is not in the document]
avdata99: > I think this looks nicer

URL updated [here](https://github.com/okfn/ckanext-unhcr/commit/89c6b542d2d007be4b5fdb80271f0750838980d0)

ckanext/unhcr/templates/user/dashboard_requests_history.html
```diff
@@ -0,0 +1,119 @@
...
+        <b title="@{{ user.name }}">{{ user.title }}</b>
```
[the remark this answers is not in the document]
avdata99: Updated [here](https://github.com/okfn/ckanext-unhcr/commit/0a3dd626417263fa118e9474b88034dc4265e68a).

ckanext/unhcr/templates/user/dashboard_requests.html
```diff
@@ -4,6 +4,8 @@
 {% endblock %}
 
 {% block primary_content_inner %}
+  <a href="{% url_for 'unhcr_dashboard.requests_history' %}">{{ _('Requests history') }}</a>
```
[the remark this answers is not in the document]
avdata99: image
Updated [here](https://github.com/okfn/ckanext-unhcr/commit/a217b4ced29edbc7c412b5942f3ba815d0f77f34)

ckanext/unhcr/templates/user/dashboard_requests_history.html
```diff
@@ -0,0 +1,119 @@
...
+        {% elif req.object_type == 'package' %}
+        <i class="fa-sitemap"></i>
+        {% endif %}
+        
+        {# if it's a request for user, we already have the requester data available #}
+        {% if req.object_type == 'user' %}
+          {# We just needed to know if it's for new user or renewal #}
+          {% if req.is_renewal %}Renewal{% else %}New user{% endif %}
+        {% else %}
+          {% set obj = h.get_cached_ckan_object_name(req.object_type, req.object_id) %}
```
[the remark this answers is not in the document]
avdata99: Yes, you are right. Updated [here](https://github.com/okfn/ckanext-unhcr/commit/cb9be2f2bafcbec4bb5c5a28fa157b9c20e46f30)

ckanext/unhcr/templates/user/dashboard_requests_history.html
```diff
@@ -0,0 +1,119 @@
...
+
+    </div>
+
+  </div>
+</div>
+{% endfor %}
+
+{% endblock %}
+
+{% block secondary_content %}
```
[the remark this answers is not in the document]
avdata99: Dropped [here](https://github.com/okfn/ckanext-unhcr/commit/c483ccae39379e231992c2ed699659595e593ff8)

ckanext/unhcr/templates/user/dashboard_requests_history.html
avdata99: This user reference was update with the changes from previous commit.
`<a href="{{ h.url_for('user.read', id=req.user.id) }}">{{ req.user.fullname or req.user.name }}</a>`
We don't really need the cache here.

For the `actioned_by` field it's different (we use it at the bottom to show _+ info_).
The query for `access_requests` is not getting this information.
In this case it make sense to use the cache.

ckanext/unhcr/cache.py
avdata99: My bad. [This change](https://github.com/okfn/ckanext-unhcr/commit/20a9fcb004761614fb7be7c497b2028349b81839) fixes it.

amercader:
I'm sorry @avdata99 but I'm still not convinced :upside_down_face: 

Caching is hard. It is tempting to use it but the reality is that unless there is a significant performance boost, it very often leads to complexity and very hard to debug bugs. I think that you can achieve what you need (querying the display name of a user and avoiding extra hits in the db) with much fewer lines of code in `blueprints/dahsboard.py`:

```diff
diff --git a/ckanext/unhcr/blueprints/dashboard.py b/ckanext/unhcr/blueprints/dashboard.py
index 0f031f27..754f73fa 100644
--- a/ckanext/unhcr/blueprints/dashboard.py
+++ b/ckanext/unhcr/blueprints/dashboard.py
@@ -118,6 +118,13 @@ def requests_history():
             {'user': user_obj.name},
             {'status': 'all'}
         )
+
+    user_names = {}
+    for request in access_requests:
+        if request['actioned_by'] not in user_names:
+            user_names[request['actioned_by']] = model.User.get(request['actioned_by']).display_name
+        request['actioned_by_name'] = user_names[request['actioned_by']]
+
     template_vars['access_requests'] = access_requests
 
     return toolkit.render('user/dashboard_requests_history.html', template_vars)
```

I agree this is not the ideal place to do that, [dictization](https://github.com/okfn/ckanext-unhcr/blob/master/ckanext/unhcr/actions.py#L867) or the actual [db query](https://github.com/okfn/ckanext-unhcr/blob/794b6a72db371c1158c63cde4f8f4d96cb853a54/ckanext/unhcr/actions.py#L924) would be better candidates so when you get the request objects from the action layer you already had all the necessary properties, but it feels like a good compromise.

We have a a good starting point for a caching mechanism with rhe `cache.py` module in case we need it in the future, but I don't think this feature requires so I'd rather remove it.

avdata99:
I removed the cache [here](https://github.com/okfn/ckanext-unhcr/commit/c89cda4d44a3f4b8948ee248da71f41c6781cfa6) @amercader 
We can discus caching in the future.

:)

redis-fail

amercader (approved):

Issue #797, opened 2022-03-21
Failed to submit a new dataset [PROD!]


avdata99:
External user get 403 after create a new dataset and try to start adding resources

![image001](https://user-images.githubusercontent.com/3237309/159348458-2200cd3e-fcf8-4419-a765-6aac41fcc115.png)

avdata99:
@amercader CKAN uses the `by_name` function to [read the user from context](https://github.com/ckan/ckan/blob/master/ckan/logic/__init__.py#L331) and we are (for some reason) using the user `id`
Even if we have a `context['user']` we [ask for](https://github.com/ckan/ckan/blob/master/ckan/authz.py#L234) `auth_user_obj`.
Why don't use just the `get` function? (I mean, instead `by_name`)
It's mandatory to use the user's `name` and not the `id`?
Sent [this PR](https://github.com/ckan/ckan/pull/6762) to CKAN core if it's a good idea

I have found we use the user id [here](https://github.com/okfn/ckanext-unhcr/blob/797_submit_error/ckanext/unhcr/activity.py#L26) and [here](https://github.com/okfn/ckanext-unhcr/blob/797_submit_error/ckanext/unhcr/commands.py#L80) (with `ignore_auth`) and [here](https://github.com/okfn/ckanext-unhcr/blob/797_submit_error/ckanext/unhcr/helpers.py#L562) (and others)

avdata99:
@amercader there is an alternative fix in #799

avdata99:
I can try this manually on prod tomorrow and check which is the best fix

avdata99:
#799 deployed to production and it's working

avdata99:
#799 merged. This issue remains in-progress until we test and check all `context` occurrences.

PR #798, opened 2022-03-21
Force using user name instead ID

avdata99:
Related to #797 

I don't like this but it's working locally
Should we change all internal references from `id` to `name` for users?


amercader (approved):

avdata99:
#799 is a better solution so far

Issue #801, opened 2022-04-20
Required field "Geographic coverage" error in resources form


avdata99:
We expect for the `geography` field to be required in the dataset form, not the resources one.

![image](https://user-images.githubusercontent.com/3237309/164223488-1ba04b2b-d441-41bb-b32b-687421b913ef.png)

avdata99:
The dataset was created in October 2021
The error while trying to add a resource to an existent dataset
Probably the migration to add the `UNSPECIFIED` geography to all datasets failed

avdata99:
Yes, [our migration](https://github.com/okfn/ckanext-unhcr/blob/release-3.1.7.1/ckanext/unhcr/migration/unhcr/versions/009_f671e76f1a39_create_a_general_geography.py#L31-L36) updates the existent `geographies` but didn't create the new required `package_extras`

This query is working fine locally to add the missing extras for `geographies`

```sql
insert into package_extra ("id", "package_id", "key", "value")
   select '', package.id, 'geographies', '{UNSPECIFIED}' from package
      left join package_extra on package_extra.package_id = package.id and package_extra.key = 'geographies'
      where package_extra.key IS NULL;
```

Options:
 - To create a migration
 - To run the query in the production DB manually

@amercader the query looks good to you?
Should we create a migration?

amercader:
@avdata99  This won't work because you are inserting empty ids (`''`) on the primary key of the package_extra table:

```
ERROR:  duplicate key value violates unique constraint "package_extra_pkey"
DETAIL:  Key (id)=() already exists.
```
Try using one of the approaches listed [here](https://stackoverflow.com/questions/12505158/generating-a-uuid-in-postgres-for-insert-statement) to insert a valid uuid. This one seems to be most portable:

```
insert into package_extra ("id", "package_id", "key", "value")
   select md5(random()::text || clock_timestamp()::text)::uuid, package.id, 'geographies', '{UNSPECIFIED}' from package
      left join package_extra on package_extra.package_id = package.id 
         and package_extra.key = 'geographies'
      where package_extra.key IS NULL;
```


Maybe if you can login to the db server we can just run the statement directly (trying UAT first of course). Make sure to check that the numbers make sense (eg count of inserted rows = count of datasets - count of existing `geographies` extras)

avdata99:
UAT

Test count datasets with and without geography

[...]

PR #803, opened 2022-04-21
Missing query for updating geographies

avdata99:
Related to #801 

Proposed SQL to run manually
Leaving inside migration just for reference


amercader (approved):

avdata99:
Production 

<pre>
ckan=> select count(package.id) from package
ckan->       left join package_extra on package_extra.package_id = package.id
ckan->          and package_extra.key = 'geographies'
ckan->       where package_extra.key IS NULL;
 count
-------
   <b>870</b>
(1 row)

ckan=> select count(package.id) from package
      left join package_extra on package_extra.package_id = package.id
         and package_extra.key = 'geographies'
      where package_extra.key IS NOT NULL;
 count
-------
    <b>17</b>
(1 row)

ckan=> insert into package_extra ("id", "package_id", "key", "value", "state")
ckan->     select md5(random()::text || clock_timestamp()::text)::uuid, package.id, 'geographies', '{UNSPECIFIED}', 'active' from package
ckan->         left join package_extra on package_extra.package_id = package.id
ckan->             and package_extra.key = 'geographies'
ckan->         where package_extra.key IS NULL;
<b>INSERT 0 870</b>

</pre>

Issue #805, opened 2022-05-02
Error in Geographic coverage filter/facet


avdata99:
Geographic coverage filter/facet
    1. If the user wants to see more locations and clicks here

![image](https://user-images.githubusercontent.com/3237309/166237426-d8837f82-6fd4-47b9-a5a9-759c2b7f1110.png)

The page reloads and the ‘Geographic coverage’ facet is collapsed again:

![image](https://user-images.githubusercontent.com/3237309/166237436-792f10b3-d931-4efd-a9d4-e600b3f478f1.png)

If I expand it again, I get more values, but it’s not the most user-friendly implementation – when the page seemed to ‘reload’, I was wondering what went wrong.

PR #820, opened 2022-06-01
Expand selected facets

avdata99:
Related to #805 

This PR ensure we expand facet list when we hit the _Show more ..._ at the end of a facet items list.


amercader:
@avdata99 RIDL uses https://github.com/okfn/ckanext-facetcollapse to modify the default facets visualization. That is probably what's causing that behaviour so perhaps we should apply the fix there (it works fine in a standard CKAN instance)

avdata99:
Closed in favor of https://github.com/okfn/ckanext-facetcollapse/pull/6

Issue #814, opened 2022-05-30
Limit KoBo imports


avdata99:
KoBo [has a limit](https://github.com/kobotoolbox/kpi/blob/master/kobo/settings/base.py#L357) for exports on 30K records.
Make sure we don't allow importing KoBo resources larger than that limit.

PR #821, opened 2022-06-02
Limit KoBop imports

avdata99:
Related to #814 

Avoid start importing from KoBo when it's not allowed

image

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,9 @@
 # RIDL Changelog
 
+## v3.1.7.3 - 2022-06-XX
+Bug fixes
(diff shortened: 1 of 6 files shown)


amercader (approved):
Does it make sense to allow imports but only import the first 30k submissions? (warning the user of course)

avdata99:
> Does it make sense to allow imports but only import the first 30k submissions? (warning the user of course)

@amercader we have a custom paginated export. Maybe the next step is to use for special cases.
We can also ask the client about allowing truncated downloads.
I don't think this PR ends this, it's just a first step.

Issue #809, opened 2022-05-02
Data container About page, missing metadata value


avdata99:
![image](https://user-images.githubusercontent.com/3237309/166240414-e2a8ae4a-aa9c-4ef2-9ba3-c13269c44197.png)

PR #826, opened 2022-06-03
Fill the visible_external field

avdata99:
Related top #809 

We probably don't fill the `visible_external` field for all data containers when we created it.
This PR updates and creates the required group extras.

Still pending to test on UAT because a VPN access expired.


avdata99:
@amercader 
I get access to the VPN again and validate a lot of data containers don't have any `visible_external` extra defined
Those containers were probably created before the fields were defined.

My plan is to run manually the query in UAT

```sql
insert into group_extra ("id", "group_id", "key", "value", "state")
            select md5(random()::text || clock_timestamp()::text)::uuid,
                "group".id, 'visible_external', 'false', 'active' from "group"
                left join group_extra on
                    group_extra.group_id = "group".id 
                    and group_extra.key = 'visible_external'
                    and "group".type = 'data-container'
                where group_extra.key IS NULL;
```

amercader (approved):

PR #829, opened 2022-06-07
Request a title review when importing from kobo

avdata99:
Users don't take care of KoBo survey titles so we are not able to automatically create a good title.
However, we can emphasize that we want a revision of the title

image


amercader (commented):
Maybe instead of using JS for this it's easier to just override the scheming form snippet for the title field (by default it uses [this one](https://github.com/ckan/ckanext-scheming/blob/master/ckanext/scheming/templates/scheming/form_snippets/large_text.html)) and add the Kobo special case like we do [here](https://github.com/okfn/ckanext-unhcr/blob/e66266180f799e2eb70eb18d674d7ae56c32e1d6/ckanext/unhcr/templates/package/snippets/package_form.html#L8)

avdata99:
@amercader updated to use to html snippet

image

PR #835, opened 2022-06-14
Add language headers if they exists in kobo resouorce

### LINKED ISSUE #834: Allow filter KoBo using laguages for headers
Related to #828 and #815 

Start showing the _Values and headers format_ filer:
In KoBo, if we have multiple languages, we are able to define which one to use as labels in the export.

image

Now, [we are using](https://github.com/okfn/ckanext-unhcr/blob/ff43d51aa1472c13f3b088c430512785b0950ea9/ckanext/unhcr/kobo/api.py#L248) just the _default_ value

To do:
 - Get the valid options while importing from KoBo
 - Show them to the user
 - Allow using this filter

avdata99:
fixes #834 

This PR:
 - Get the languages list for a KoBo asset.
 - Allows you to select languages as resource headers (for datasets and resources).

image

image


amercader (approved):
I have absolutely no idea what this does but the implementation looks good.

Would it help if the dropdown said "Labels - Spanish (es)", "Labels - Italian (it)"? is that what you are choosing there?

In general I worry that we are over-complicating the Kobo import form but if people familiar with Kobo know what they are seeing then it might be fine

Issue #762, opened 2022-01-04
Filter survey before import [KoBo integration improvements]


avdata99:
Filter survey before import (submission dates, label vs XML, groups).

Expose to RIDL users parameters that the KoBo API alredy exposes when creating an export, but we really don't want to have to parse the KoBo data ourselves in order to filter the results.

Related to #764 and #763

KoBo export documentation: https://kobo.unhcr.org/exports/
KoBo UI export options:

![image](https://user-images.githubusercontent.com/3237309/148098414-7766ab1f-7141-4e6c-aac9-9a501692d2b1.png)

avdata99:
Filters are done by  #765 
We probably can improve this but with new issues

PR #836, opened 2022-06-15
Allow users to search datasets by geography in a map

### LINKED ISSUE #767: Allow users to search datasets by geography in a map
Provide a map widget where users can draw an area of interest. It requires that we index the geometry (or an approximation of it) of all the geographies we are currently storing, which means changes at the DB and Solr level.

This will require to:
 - [ ] Start saving the geometries in the database
 - [ ] Allow bounding box search
 - [ ] Show a map in the UI to search

Analyze:
 - ckanext-spatial: https://github.com/ckan/ckanext-spatial
 - USA/GSA [location search](https://github.com/GSA/ckanext-geodatagov/blob/0853bee402148f7ccabc8f37ca8c10088e6dcb42/ckanext/geodatagov/logic.py#L23)

avdata99:
Fixes #767


avdata99:
Moved to #762

PR #840, opened 2022-06-16
Get sub types from GIS server

### LINKED ISSUE #839: Improve label for geographies
Replace [Pop. of Concern Location] 

image

Out of the two ‘Pop. of Concern Location’s, one is an urban refugee location, the other one is an urban IDP location. This distinction can only be displayed if we **show the type of location**, instead of the generic ‘Pop. of Concern Location’. So for example for the first one, it would say _“Ouagadougou (Kadiogo, Centre, Burkina Faso)[Refugee - Urban location]”_.

We can get this information from two fields: `poc_type` for the type of population and `loc_subtype` for the type of location.

avdata99:
fixes #839 

This PR:
 - Get sub types for locations from the GIS server
 - Add new local fields to keep track of this information.
 - Show proper labels for locations

image

image


amercader (approved):
This looks good. My only complain is that the dropdown results are getting very verbose and hard to read. Maybe we can add some simple styling that separates the different parts hierarchically. Same for the displayed metadata:

Screenshot 2022-06-29 at 14-47-30 Test dataset 1 - RIDL

This can be done in a separate PR

PR #842, opened 2022-06-22
Hack geography names search

### LINKED ISSUE #838: Fix geography search error
Finding Bobo-Dioulasso is only possible if one searches for ‘Dioulasso’.  
 
image

image

avdata99:
Fixes #838 

This PR hack `similarity` search for this case
The idea is to give some extra points for searches that starts the same as the geography name.

image


CHANGES.md
```diff
@@ -2,6 +2,7 @@
 
 ## v3.1.8.1 - 2022-06-XX
 New features:
+ - Hack geography search [#842](https://github.com/okfn/ckanext-unhcr/pull/842)
```
amercader, line 5:
```suggestion
 - Improve geography search results [#842](https://github.com/okfn/ckanext-unhcr/pull/842)
```

amercader (approved):
This is a good workaround. I suspect that we could try to implement this in postgres  itself changing the query and adding a custom index, making it much faster, but for now this should work.

Issue #761, opened 2022-01-04
Maintenance mode


avdata99:
Allow a maintenance page for RIDL

 - Check if AWS include maintenance options
 - Check this [related commit](https://github.com/okfn/docker-ckan-unhcr-aws/commit/886df747372b78c821468f2e708b644185b909b4) (it seems it was not implemented)
 - Check this [CKAN issue/discussion](https://github.com/ckan/ideas/issues/171)

Allow sysadmins to define a general notification message for all users (e.g. provide advance notice of scheduled application maintenance)

avdata99:
AWS Load balances now supports [redirect and fixed responses](https://aws.amazon.com/about-aws/whats-new/2018/07/elastic-load-balancing-announces-support-for-redirects-and-fixed-responses-for-application-load-balancer/)

By adding a new listener (with a custom HTTP response and HTML conten)

![image](https://user-images.githubusercontent.com/3237309/176954855-c70fe4d1-bf99-491d-b0cd-9c6e76555d10.png)

And changing ports for listeners

![image](https://user-images.githubusercontent.com/3237309/176955177-2efb22e4-79b1-4c5a-a973-41c9cca44b4b.png)

We can manually redirect all traffic 

![image](https://user-images.githubusercontent.com/3237309/176954718-43c0b6aa-f380-4d4e-8c1a-10b26c232c78.png)

Pending to check if we can run this with a simple script

What do you think @amercader ?

avdata99:
@amercader [here](https://github.com/okfn/docker-ckan-unhcr-aws/pull/83/files) a PR to add this as a feature at the infrastructure repo. It this is a good idea. Pending to replicate to production if you agree.

avdata99:
All PR merged

PR #861, opened 2022-07-20
Add ckanext-announcements for testing

avdata99:
Related to #761


bin/init-db.bash
```diff
@@ -3,3 +3,13 @@ set -euo pipefail
 
 ckan -c ./test.ini db init
 ckan -c ./test.ini unhcr init-db
+
+# # Apply alembic migrations for announcements
+# echo "Running announcements migrations"
+# ckan -c ./test.ini db upgrade -p announcements
```
amercader, line 9: This should be uncommented right?

amercader (approved):

bin/init-db.bash
avdata99: @amercader Test fails if I uncomment this because each test drop the database and _announcements_ migration are not running 

I can fix with [this](https://github.com/okfn/ckanext-unhcr/commit/66fda09e76c16bc0c177fb57eea08be8701e07ed).

Do you think this is the best solution?

PR #888, opened 2022-09-30
Script to get geometries, command to import them

avdata99:
In progress


avdata99:
Features moved to #888

PR #891, opened 2022-10-13
Show users permission

avdata99:
Users periodically requires a list of users with permission in containers. 
This implements a feature so they can easily generate it themselves
image


ckanext/unhcr/helpers.py
```diff
@@ -1276,3 +1276,33 @@ def get_system_activities():
 def get_bool_arg_value(args, arg_name, default=False):
     """ Booleanize a value from a dict"""
     return toolkit.asbool(args.get(arg_name, default))
+
+
+def get_all_containers_and_members(include_admins=True):
```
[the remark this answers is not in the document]
amercader: I think this should be an action, rather than a helper (with an auth function). This way it can also be accessed via the API

ckanext/unhcr/templates/admin/containers.html
```diff
@@ -0,0 +1,48 @@
...
+      <th>Users</th>
+    </tr>
+  </thead>
+  <tbody>
+    {% for container in containers %}
+    <tr>
+      <td>{{ container.title }}</td>
+      <td>
+        {% for user in container.users %}
+        <a href='{% url_for "user.read", id=user.id %}'>{{ user.name }}</a> ({{ user.capacity }})
```
amercader, line 26:
```suggestion
        <a href='{% url_for "user.read", id=user.id %}'>{{ user.name }}</a> ({{ user.capacity }}){% if not loop.last %}, {% endif%}
        
```

ckanext/unhcr/templates/admin/containers.html
```diff
@@ -0,0 +1,48 @@
+{% extends "admin/base.html" %}
+
+{% set containers = h.get_all_containers_and_members() %}
+
+{% block primary_content_inner %}
+
+<h2>Container permissions</h2>
```
amercader, line 7: There is a ton of sysadmins on Prod, so perhaps is better to not include them in the table, and add a small text that says "Note: Sysadmin users (link to the sysadmins tab) are not included"

amercader (commented):
Great stuff @avdata99 , left some comments

ckanext/unhcr/helpers.py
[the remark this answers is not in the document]
avdata99: Moved to action [here](https://github.com/okfn/ckanext-unhcr/pull/891/commits/7922104e3c647d47628a37a336539f163497f55e) @amercader

ckanext/unhcr/templates/admin/containers.html
avdata99: Done with both options
image

ckanext/unhcr/plugin.py
```diff
@@ -143,6 +144,7 @@ def update_config(self, config_):
         toolkit.add_resource('fanstatic', 'unhcr')
         toolkit.add_ckan_admin_tab(config_, 'unhcr_search_index.index', 'Search Index', icon='hdd-o')
         toolkit.add_ckan_admin_tab(config_, 'unhcr_system_activities.index', 'System activities', icon='history')
+        toolkit.add_ckan_admin_tab(config_, 'unhcr_containers_info.index', 'Container permissions', icon='dropbox')
```
amercader, line 147:
```suggestion
        toolkit.add_ckan_admin_tab(config_, 'unhcr_containers_info.index', 'Data Containers permissions', icon='user')
```

ckanext/unhcr/templates/admin/containers.html
```diff
@@ -0,0 +1,57 @@
...
+  </tbody>
+</table>
+{% endblock %}
+
+
+{% block secondary_content %}
+<div class="module module-narrow module-shallow">
+  <h2 class="module-heading">
+    <i class="fa fa-history"></i>
+    {{ _('Containers permissions') }}
```
amercader, line 49:
```suggestion
    {{ _('Data Containers permissions') }}
```

amercader (approved):
Looks good @avdata99, I just suggested some minor tweaks to make the naming consistent

avdata99:
@mariannurban a proposal was just sent to change the title and add link for all containers.

PR #901, opened 2022-10-31
Hide data resource fields for attachment resources

### LINKED ISSUE #900: Unexpected  metadata visible for file attachments
image

Some microdata metadata fields are displayed for fille attachment resources

avdata99:
Fixes #900 

There's no easy way to hide field that belong to a schema.
This [is defined](https://github.com/ckan/ckanext-scheming/blob/master/ckanext/scheming/templates/scheming/package/resource_read.html#L69) in the ckanext-scheming extension.

Attachment resource fixed:

image

Data resource:

image


ckanext/unhcr/helpers.py
```diff
@@ -1285,3 +1285,18 @@ def data_containers_with_members_list(include_sysadmins=True, user=None):
...
     return containers
+
+
+def get_resource_data_type_fields():
+    """ Get a list of fields for "data" resource not beloging to "attachment" resources """
+    return [
+        'date_range_start', 'date_range_end', 'hxl-ated',
+        'data_collector', 'version', 'process_status',
+        'identifiability'
+    ]
```
amercader, line 1296: These lists always end up going out of sync. Let's keep our schema as the single source of truth and add a `"data_only": true` or similar key to the relevant resource fields, and then we can do:

```suggestion
    schema = scheming_get_dataset_schema("dataset")                                                                                     
    
    return [f["field_name"] for f in schema['resource_fields'] if f.get("data_only")]
```

ckanext/unhcr/templates/package/resource_read.html
```diff
@@ -89,6 +89,14 @@
 
 {% block primary_content %}
 
+{# Attachment resources are not supposed to display fields they don't have.
+   The "exclude_fields" var belongs to the ckanext-scheming extension.
+   We probably need to define custom schemas for "attachment" and "data" resources.
+#}
+{% if res.type == 'attachment' %}
+  {% do h.override_schema_exclude_fields(exclude_fields) %}
```
amercader, line 97:
```suggestion
  {% do h.extend_schema_exclude_fields(exclude_fields) %}
```

ckanext/unhcr/helpers.py
avdata99: Considering #904 we will need to cover multiple resource schemas. I'll propose a change thinking on that @amercader

avdata99:
[This change](https://github.com/okfn/ckanext-unhcr/commit/0b1cd083488b7dd8d6ffe17aa41e927b5de146b6) is even better @amercader. We are ~ready for #904 

We can define in the resource schema what fields and resource types to hide and

**Data type**

image

**Attachment type**

image

amercader (approved):

PR #905, opened 2022-11-01
Add spatial rebased

avdata99:
This is #862 but rebasing `develop` into `add_spatial_extension`

This is a test


avdata99:
Done, no longer required

PR #910, opened 2022-11-07
Create log for metrics snapshot

avdata99:
Related to [Docker#102](https://github.com/okfn/docker-ckan-unhcr-aws/issues/102)

This PR:
 - Start capturing errors for the `snapshot-metrics` command
 - Create a system activity to log command results

image


amercader (approved):
This seems like a good idea @avdata99 but we should delete old activities otherwise we will end up with a huge table. We could delete the activities of this kind that are older than X days (2 months? 4) as part of the same command (at the beginning of running the command)

amercader:
How often is this command run? Maybe a longer retention period is fine, basically I don't want to end up with thousands of activities that say "yup all went well"

avdata99:
> How often is this command run? Maybe a longer retention period is fine, basically I don't want to end up with thousands of activities that say "yup all went well"

@amercader, this runs [every 6 hours](https://github.com/okfn/docker-ckan-unhcr-aws/blob/master/ckan/yacrontab-prod.yaml#L15).

I think it will be a good idea to preserve only lasts ~500 (with a config setting) activity records
We can do it in the `create_system_activity` function. Every time we add one log, we clean oldest records

avdata99:
@amercader deleting old sysm activity logs here

https://github.com/okfn/ckanext-unhcr/commit/83e266dd178b944c05676eb3accc27f0bf60e016

PR #919, opened 2022-11-15
Update scheming

avdata99:
Related to [Docker#104](https://github.com/okfn/docker-ckan-unhcr-aws/pull/104)
To run test with new scheming version
A [new release](https://github.com/ckan/ckanext-scheming/issues/349) will be better to avoid automatic updates without notice


bin/install-extensions.bash
```diff
@@ -5,7 +5,7 @@ pip install -U pip
 pip install setuptools-rust
 
 
-git clone --depth 1 --branch release-2.1.0 https://github.com/ckan/ckanext-scheming
+git clone --depth 1 --branch master https://github.com/ckan/ckanext-scheming
```
amercader, line 8: Let's point to a specific commit instead, until there is not a formal release

avdata99: Updated [here](https://github.com/okfn/ckanext-unhcr/commit/c2373b4f75c4c85281e0c2d0e4018fb7c6f7ba74) and [here](https://github.com/okfn/docker-ckan-unhcr-aws/pull/104)

amercader (approved):

PR #922, opened 2022-11-23
Delay possible attackers

### LINKED ISSUE #866: User register form improvement
Partner account request page displays an error
message that indicates if a user account has already been registered. Someone can
use this information to enumerate valid usernames in preparation for password
guessing attacks.

Remediation
 - To send same response for both valid and invalid input within the given field.
   - Another way is to always say something like *Thanks for attempt to register, you'll receive an email if the account is approved* 
 - To implement a CAPTCHA on the affected form.
   - AWS include a captcha service using [AWS WAF ](https://docs.aws.amazon.com/waf/latest/developerguide/waf-captcha-how-it-works.html)
   - CKAN is ready to implement Google Recaptcha for the `/user/register` URL
 - To enforce accounts security we can implement 2FA.

image

avdata99:
Fixes #866 

Delay possible attackers while trying to guess user names or emails


ckanext/unhcr/blueprints/user.py
```diff
@@ -239,6 +241,24 @@ def post(self):
...
+                    redis_pool = ConnectionPool.from_url(redis_url)
+                    cache = Redis(connection_pool=redis_pool)
+                except ConnectionError:
+                    log.error(f'Could not connect to Redis at {redis_url}')
+                else:
+                    client_ip_address = toolkit.request.environ.get('REMOTE_ADDR','')
+                    cache_key = 'user_register_{}'.format(client_ip_address)
+                    delay = int(cache.get(cache_key) or 1)
+                    log.error(f'Delaying user creation for {delay} seconds')
+                    sleep(delay)
```
amercader, line 259: This approach will eventually cause a 504 Gateway timeout when the Load Balancer decides to not wait for the response any more. A better one is to store here the error time and the number of seconds a user should have to wait, and the at the beginning of the request handling (ie beginning of the `post()` method) we check if current_time - error_time > seconds, and if not we return an error saying that the user needs to wait before trying again (so we don't hold the response)

Does this make sense?

avdata99: Yes, it makes @amercader 
I updated this PR to change the approach. Now we avoid timeouts.

amercader (approved):
Nice one @avdata99 !

PR #926, opened 2022-11-24
Fix related unath datasets

### LINKED ISSUE #925: Error with related datasets
A user have error at dataset.read if the user has permissions on the dataset but not on its related datasets

<pre>
2022-11-23 13:11:24,223 INFO [ckan.config.middleware.flask_app] 404 Not Found: Dataset not found
2022-11-23 13:11:24,236 INFO [ckan.config.middleware.flask_app] 404 /dataset/bgr-protmon-2022-05-09 render time 0.022 seconds
2022-11-23 13:11:26,356 ERROR [ckan.config.middleware.flask_app] User XXXXX not authorized to read package b3f62933-289f-42ac-8c83-d81927701990
File "/srv/app/src/ckanext-unhcr/ckanext/unhcr/templates/scheming/display_snippets/linked_datasets.html", line 2, in <module>
{% for dataset in h.get_linked_datasets_for_display(data[field.field_name]) %}
File "/srv/app/src/ckanext-unhcr/ckanext/unhcr/helpers.py", line 411, in get_linked_datasets_for_display
dataset = toolkit.get_action('package_show')(context, {'id': id})
File "/srv/app/src/ckan/ckan/logic/__init__.py", line 504, in wrapped
File "/srv/app/src/ckan/ckan/logic/action/get.py", line 978, in package_show
_check_access('package_show', context, data_dict)
File "/srv/app/src/ckan/ckan/logic/__init__.py", line 313, in check_access
raise NotAuthorized(msg)
ckan.logic.NotAuthorized: User XXXXX not authorized to read package b3f62933-289f-42ac-8c83-d81927701990
2022-11-23 13:11:26,427 INFO [ckan.config.middleware.flask_app] 500 /dataset/bgr-protmon-2022-05-09 render time 2.081 seconds
</pre>

avdata99:
Fixes #925


amercader:
@avdata99 help me remember how this works. How can a user end up not having access to a dataset? IIRC visibility is handled at the resources level and datasets are always visible (at least the metadata)

avdata99:
> @avdata99 help me remember how this works. How can a user end up not having access to a dataset? IIRC visibility is handled at the resources level and datasets are always visible (at least the metadata)

@amercader there no easy response

Sysadmin users can define a dataset as `private`
https://github.com/okfn/ckanext-unhcr/blob/develop/ckanext/unhcr/tests/pt/controllers/test_dataset.py#L389-L401
```python
    def test_private_is_false_if_not_sysadmin(self):
        dataset = factories.Dataset(
            private=True, user=self.normal_user)

        assert dataset['private'] == False

    def test_private_can_be_true_if_sysadmin(self):
        dataset = factories.Dataset(
            private=True,
            visibility='private',
            user=self.sysadmin)

        assert dataset['private'] == True
```
Visibility is required for resources and not required for datasets.
The "visibility" of a dataset is calculated as the most restricted of its resources.
https://github.com/okfn/ckanext-unhcr/blob/develop/ckanext/unhcr/jobs.py#L162-L165
```python
    # visibility
    # if some of the resources is restricted then the package will be restricted
    weights = {'restricted': 2, 'public': 1}
    package = _modify_weighted_field(package, 'visibility', weights, default_value='public')
```

visibility = 'private' for datasets means
visibility = 'restricted' + private = True
https://github.com/okfn/ckanext-unhcr/blob/develop/ckanext/unhcr/validators.py#L194-L210
```python
def visibility_validator(key, data, error, context):
    ''' Validates visibility has a correct value.

    Visibility only has two values in the schema, 'restricted' and 'public'. The
    value 'private' is used only to setup the actual private field if needed.
    '''
    value = data.get(key)

    if value not in ('private', 'restricted', 'public'):
        raise Invalid('Invalid value for the visibility field')

    # Set the actual private field
    if value == 'private':
        data[('private',)] = True
        data[('visibility',)] = 'restricted'
    else:
        data[('private',)] = False
```

But now we don't display this field `visibility` for datasets anymore
It's tricky. Maybe a sysadmin do this in the past.

I'll inevstigate a little bit more

avdata99:
@amercader The related dataset is deposited.
We have custom dataset and user _labels_ for deposited datasets
https://github.com/okfn/ckanext-unhcr/blob/develop/ckanext/unhcr/plugin.py#L681-L736
This user probably do not have  permission to deposit datasets

In any case, we need to capture possible errors in that list

ckanext/unhcr/helpers.py
```diff
@@ -408,8 +408,14 @@ def get_linked_datasets_for_display(value, context=None):
...
     ids = normalize_list(value)
-    for id in ids:
-        dataset = toolkit.get_action('package_show')(context, {'id': id})
+    for _id in ids:
+        try:
+            dataset = toolkit.get_action('package_show')(context, {'id': _id})
+        except toolkit.ObjectNotFound:
+            continue
+        except toolkit.NotAuthorized:
+            datasets.append({'text': 'An unathorized dataset', 'href': None})
```
amercader, line 417: Just don't show it to the users if they are not allowed to access it, ie just `continue` without adding it

avdata99: Updated [here](https://github.com/okfn/ckanext-unhcr/commit/925ecb40474991e84344a93e24eb1b3209656a72)

avdata99:
I created #927 to investigate more on this

amercader (approved):

PR #931, opened 2022-12-01
Hide empty maps

### LINKED ISSUE #924: Error showing map in dataset
Some datasets are not properly show map with geometries. 
[Example](https://ridl-uat.unhcr.org/dataset/syr-registration-data-2022-06):

image

avdata99:
Fixes #924 

image


ckanext/unhcr/templates/package/read_base.html
```diff
@@ -53,6 +53,8 @@
       {% set dataset_extent = h.get_geojson_geometries(pkg.geographies) %}
       {% if dataset_extent %}
         {% snippet "spatial/snippets/dataset_map_sidebar.html", extent=dataset_extent %}
+      {% else %}
+      <p>No map available for the selected dataset geography</p>
```
amercader, line 57: This will be shown to datasets that don't have any geographies assigned right? If so let's say
```suggestion
      <p>No geographic coverage defined</p>
```

amercader (approved):
@avdata99 While you are at it, can you make sure we display the Font Awesome globe next to Dataset Extent and move the whole section to the bottom of the secondary content column?

ckanext/unhcr/templates/package/read_base.html
avdata99: No @amercader, it also include defined geographies that don't have a geometry defined. 
I'm proposing [this change](https://github.com/okfn/ckanext-unhcr/commit/faaaf830fe1cabc5c5e9f1ea9062f05521bedd0c)

```html
      {% set dataset_extent = h.get_geojson_geometries(pkg.geographies) %}
      {% if dataset_extent %}
        {% snippet "spatial/snippets/dataset_map_sidebar.html", extent=dataset_extent %}
      {% elif pkg.geographies %}
      <p>No map available for the selected dataset geography</p>
      {% else %}
      <p>No geographic coverage defined</p>
      {% endif %}
```

avdata99:
> @avdata99 While you are at it, can you make sure we display the Font Awesome globe next to Dataset Extent and move the whole section to the bottom of the secondary content column?

@amercader I move the map to the botton and add the globe here

PR #952, opened 2023-01-05
Lint code

avdata99:
Sorry
I need to do this from a long time


amercader:
This is great, but:

* Let's wait until everything else is merged to avoid conflicts
* Separate the setting up and configuration of the linting and the actual linting in separate commits
* If we are going to check linting in CI is better to enforce it before pushing with a pre-commit hook to avoid having to fix the linting after pushing
* Let's use black as we use in other OKF repos

So I suggest:

* Add [pre-commit](https://pre-commit.com/) to the dev-requirements and add a configuration file (`pre-commit-config.yaml`):
```yaml
# See https://pre-commit.com for more information
# See https://pre-commit.com/hooks.html for more hooks
repos:
-   repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v4.3.0
    hooks:
    -   id: trailing-whitespace
-   repo: https://github.com/psf/black
    rev: 22.12.0
    hooks:
    -   id: black
```
* Add a [Github Action job](https://black.readthedocs.io/en/stable/integrations/github_actions.html#usage) to our existing workflow 
* Lint the files
* Add a note in the README about this

avdata99:
@amercader 
 - pre-commit added 
 - GH actions updated

amercader (approved):

Issue #953, opened 2023-01-09
Duplicated email 


avdata99:
This week's weekly summary came duplicate

![image](https://user-images.githubusercontent.com/3237309/211333039-9164ff1e-b787-4d25-ae58-be10155c7b78.png)

PR #954, opened 2023-01-09
Avoid sending duplicate summaries

avdata99:
Check for #953 
This is probably not needed


ckanext/unhcr/click_commands/misc.py
```diff
@@ -77,8 +77,15 @@ def send_summary_emails(verbose):
...
     subject = '[UNHCR RIDL] Weekly Summary'
 
+    emails = []
     for recipient in recipients:
         if recipient['email']:
+            # Avoid dulicate emails
+            if recipient['email'] in emails:
+                click.echo('SKIPPING duplicate summary email to: {}'.format(recipient['email']))
+                continue
+            emails.append(recipient['email'])
```
amercader, line 87: Why is the `emails` var used for?

avdata99: @amercader, `emails` is used to know what email was processed to skip if duplicated

here
```python
# Avoid dulicate emails
 if recipient['email'] in emails:
    continue
```

amercader (approved):

PR #970, opened 2023-02-02
Prepare plugin for CKAN 2.10

### LINKED ISSUE #966: CKAN 2.10 IPackageController and IResourceController hooks renamed
All methods from IPackagController and IResourceController were renamed to include the entity name in order to avoid clashes between the two interfaces, eg IPackageController.before_create() -> IPackageController.before_dataset_create() and IResourceController.before_create() -> IResourceController.before_resource_create() (full list of renamed hooks in [#6501](https://github.com/ckan/ckan/pull/6501)).

If your CKAN extension just targets one CKAN version just rename the hooks on your plugins so they get call again, eg change this:
```python
class MyPlugin():

    p.implements(p.IPackageController)

        # Do whatever
```
to this:

```python
class MyPlugin():

    p.implements(p.IPackageController)

    def after_dataset_create(self, context, data_dict):
        # Do whatever
```

If your extension targets multiple CKAN versions you can just add both hooks to the plugin class and the relevant one will get called on each CKAN version, eg:

```python
class MyPlugin():

    p.implements(p.IResourceController)

    # CKAN < 2.10
        return self.after_resource_create(context, data_dict)

    # CKAN >= 2.10
    def after_resource_create(self, context, data_dict):
        # Do whatever
```

avdata99:
Fixes #966


ckanext/unhcr/plugin.py
```diff
@@ -513,44 +510,53 @@ def _package_after_create(self, context, pkg_dict):
...
                     jobs.process_dataset_on_update,
                     [res_dict['package_id']],
-                    title="_resource_after_update {}".format(res_dict['id'])
+                    title="after_resource_update {}".format(res_dict['id'])
                 )
 
+    # TODO remove this function onve CKAN 2.10 were released
     def after_delete(self, context, data_dict):
+        if 'owner_org' in data_dict and 'package_id' not in data_dict:
+            self.after_dataset_delete(context, data_dict)
```
avdata99, line 557: If a resource has the same ID than a package, we will delete it

amercader (approved):

Issue #969, opened 2023-02-02
CKAN 2.10 Frontend and templates


avdata99:
# CSRF input field
Starting from CKAN 2.10 all forms submitted from the UI need to include the CSRF token. To include them, just add the following helper call to your form templates:
```html
<form class="dataset-form form-horizontal" method="post" enctype="multipart/form-data">
     {{ h.csrf_input() }}
```
If your extension needs to support older CKAN versions, use the following:
```html
<form class="dataset-form form-horizontal" method="post" enctype="multipart/form-data">
     {{ h.csrf_input() if 'csrf_input' in h }}
```

# CSRF and JavaScript
Forms that are submitted via JavaScript modules also need to submit the CSRF token, here’s an example of how to append it to an existing form:
```js
  // Get the csrf value from the page meta tag
  var csrf_value = $('meta[name=_csrf_token]').attr('content')
  // Create the hidden input
  var hidden_csrf_input = $('<input name="_csrf_token" type="hidden" value="'+csrf_value+'">')
  // Insert the hidden input at the beginning of the form
  hidden_csrf_input.prependTo(form)
```

API calls performed from JavaScript modules from the UI (which use cookie-based authentication) should also include the token, in this case in the X-CSRFToken header. CKAN Modules using the builtin [client](https://docs.ckan.org/en/2.9/contributing/frontend/index.html?#client) to perform API calls will have the header added automatically. If you are performing API calls directly from a UI module you will need to add the header yourself.

# Bootstrap 5 data attributes

CKAN 2.10 ships with Bootstrap 5 by default. The custom HTML data attributes used by the Bootstrap JS components (eg modal dialogs, popups, etc) are now prefixed with data-bs- instead of just data-, so you need to update those to trigger the actions:
```
diff --git a/ckanext/harvest/templates/source/new_source_form.html b/ckanext/harvest/templates/source/new_source_form.html
index 6ea495a4..eda3e201 100644
--- a/ckanext/harvest/templates/source/new_source_form.html
+++ b/ckanext/harvest/templates/source/new_source_form.html
@@ -37,7 +37,7 @@
         <label class="radio">
           <input type="radio" name="source_type" value="{{ harvester['name'] }}" {{ "checked " if checked }} data-module="harvest-type-change">
           {{ harvester['title'] }}
-          <i class="fa fa-question-circle icon-question-sign muted" title="{{ harvester['description'] }}" data-toggle="tooltip"></i>
[...]

PR #974, opened 2023-02-03
CSRF input for all HTML forms

avdata99:
Related to #969


amercader:
This looks good @avdata99 but should we create a ckan-2.10 feature branch where we merged all the 2.10 stuff? This way if we need to deploy another unrelated change to UAT/prod we don't need to upgrade to 2.10 yet

avdata99:
@amercader all PRs for CKAN 2.10 preserves compatibility with CKAN 2.9. Let me know if they aren't.

The plan is to make this code valid for both version an once we update and check everything works, clean the code.
This will allow to merge this code and avoid a future complex merge for a big _ckan_2.10_ branch.
I'm not sure which is better. I try to avoid big merges. 

Let me know which path do you think it's better.

amercader (approved):

PR #975, opened 2023-02-03
Add CSRF for JS requests and forms

avdata99:
Related to #969


ckanext/unhcr/src/js/checkbox-switcher.js
```diff
@@ -1,13 +1,10 @@
...
-        form.submit();                                                                                                                                                                                                                                                                 
-      });                                                                                                                                                                                                                                                                              
-    }                                                                                                                                                                                                                                                                                  
+this.ckan.module('checkbox-switcher', {
+
+    initialize: function () {
+      var _this = this;
+      var form = this.el.closest('form');
+      this.el.on('change', function () {
+        form.submit();
```
amercader, line 7: Will submitting forms include the default headers that you added on `ajaxSend()`? I'm not sure it uses the same mechanism. It's probably safer to include a hidden field:

```suggestion
        // Get the csrf value from the page meta tag
        var csrf_field_name = $('meta[name=csrf_field_name]').attr('content');
        var csrf_value = $('meta[name=' + csrf_field_name + ']').attr('content')
        // Create the hidden input
        var hidden_csrf_input = $('<input name="_csrf_token" type="hidden" value="'+csrf_value+'">')
        // Insert the hidden input at the beginning of the form
        hidden_csrf_input.prependTo(form)
        form.submit();
```

amercader (approved):

Issue #955, opened 2023-01-11
Migrate `setup` fn tests to the recommended `pytest` way of using fixtures


avdata99:
CKAN 2.10 will require this change in tests:

https://github.com/ckan/ckan/wiki/CKAN-2.9-to-2.10-migration-tips#test-classes-setup-methods-not-firing-or-firing-in-wrong-order

Some already updated tests: [here](https://github.com/okfn/ckanext-unhcr/commit/b6864f61f4d29ad62c59213fbd386682d8bd0db3) and [here](https://github.com/okfn/ckanext-unhcr/commit/e4842bbe965d1936e0b41a18eda6928b518a5866)

PR #978, opened 2023-02-08
Proposal for working tests for 2.9 & 2.10

avdata99:
Related to #955 

This is a proposal.
A new fixture to get the right `extra_environ` for `app` calls was created and applied to two test files.
Local `factories` were updated to use Users with tokens if available.


amercader (commented):
This approach looks great to me

avdata99:
> This approach looks great to me

Assume this as an approval.

PR #979, opened 2023-02-08
Replace test `setup` functions

avdata99:
Related to #955 

Start replacing test `setup` functions


amercader (approved):
Nice approach, didn't know about `SimpleNamespace`

PR #997, opened 2023-03-01
Start RIDL with CKAN 2.10

avdata99:
In progress


ckanext/unhcr/auth.py
```diff
@@ -536,10 +536,8 @@ def _get_user(username_or_email):
     if userobj:
         return userobj
     user = model.User.by_email(username_or_email)
-    # TODO remove list type check on CKAN 2.10
-    if user and isinstance(user, list):
-        if len(user) == 1:
-            return user[0]
+    if user:
```
pdelboca, line 539: `User.by_email()` calls `first()` that returns Object or `None`. So it is safe to just return user.

avdata99: Thanks  @pdelboca, updated [here](https://github.com/okfn/ckanext-unhcr/commit/a882cf7e647cbd47a39722045f7e9171e82ef000)

PR #1012, opened 2023-04-10
Ignore existing members when adding

### LINKED ISSUE #1009: Improve adding member to data containers
Adding members to data containers could be improved:
 - If a user is already member yo can add it again (this will not generate a duplicate but it could be better to hide it)
 - You can not see the real name of a user while search for them

image

avdata99:
fixes #1009 

This PR:
 - Add the user _fullname_ to the user list
 - Hide users who are already members of the group

image

I didn't managed to override `ckan.views.api.user_autocomplete` to add the new parameter I need: `ignore_existent_group_members` so I take the parameter from the `request`
Is there a better way to do this?


ckanext/unhcr/actions.py
```diff
@@ -1587,6 +1588,7 @@ def unhcr_user_autocomplete_schema(
     return {
         'q': [not_missing, unicode_safe],
         'ignore_self': [ignore_missing],
+        'ignore_existent_group_members': [ignore_missing],
```
amercader, line 1591:
```suggestion
        'ignore_existing_group_members': [ignore_missing],
```

ckanext/unhcr/actions.py
```diff
@@ -1617,6 +1621,31 @@ def user_autocomplete(context, data_dict):
...
     ignore_self = data_dict.get('ignore_self', False)
 
+    # You maybe will want to ignore some users
+    ignore_users = []
+
+    # There is no way to override ckan.views.api.user_autocomplete
+    # to add a new parameter
+    # If we have a request context, we can get the parameter from there
+    if current_request:
+        ignore_existent_group_members = current_request.params.get('ignore_existent_group_members')
```
amercader, line 1631: Actions should not use the request object.

What do you mean by not being able to override `ckan.views.api.user_autocomplete`?
Can we not register this in our own ridl `api.py` blueprint?

```python
def unhcr_user_autocomplete():
   ...

version_rule = '/<int(min=1, max=2):ver>'
unhcr_api_blueprint.add_url_rule(
    rule='/util/user/autocomplete',
    view_func=unhcr_user_autocomplete,
    methods=['GET'],
)
unhcr_api_blueprint.add_url_rule(
    rule=version_rule + '/util/user/autocomplete',
    view_func=unhcr_user__autocomplete,
    methods=['GET'],
)


```

avdata99: Ok, I'll try that way. I try to monkeypatch the function (like we did for others) and didn't work.

avdata99: Done @amercader, ready for re-review

amercader (approved):

PR #1025, opened 2023-05-11
Deprecate API keys

### LINKED ISSUE #1023: Deprecate API Keys
Force deprecate API keys for CKAN 2.9
API keys do not have usage logs
Force users to define an API token

avdata99:
Fixes #1023 

Related to [Docker#138](https://github.com/okfn/docker-ckan-unhcr-aws/pull/138)

This PR:
 - Deprecates API Keys


amercader:
I think a cleaner approach is to delete all api keys from the database (`update "user" set apikey = null`, maybe except the site user one). In CKAN 2.10 API keys are no longer supported by default so this won't be an issue

avdata99:
@amercader But maybe we'll need to block the `user_generate_apikey` action to prevent users to regenerate API Keys.

amercader (approved):

PR #1032, opened 2023-05-19
Search error: empty filters

avdata99:
Fixes #1018 

This PR is not the actual fix but ([this is](https://github.com/okfn/ckanext-facetcollapse/pull/7)):
 - Fix the name for the `before_search` plugin function
 - Add the missing style for facet list


amercader:
@avdata99 ok then apply the change to the more specific class (`hidden separator`) or we'll hide something somewhere else without intending it

amercader (approved):

avdata99:
@amercader the `hidden` class is supposed to hide things in CKAN 2.9.
https://github.com/ckan/ckan/blob/dev-v2.9/ckan/public/base/css/main.css#L6586
It look like CKAN 2.10 changes this.

Updated [here](https://github.com/okfn/ckanext-unhcr/commit/e7e719274378e28a2c264e634ad1e50c1ac326b0)

image

PR #1037, opened 2023-05-22
Add org action redirections

avdata99:
I started having an error related with the hierarchy extension: 
`"Action 'data-container_show' not found"`
Hierarchy calls `group_type + 'show'` action and this action it's not defined.

This solves the issue but I want to be sure is the right approach.
This is the only CKAN action hierarchy calls.


amercader:
That feels like a bug in ckanext-hierarchy to be honest. I guess the original intent was to call either `group_show` or `organization_show`, but it should not use custom types

avdata99:
@amercader. This was changed [last year](https://github.com/ckan/ckanext-hierarchy/commit/5880c364e5a183523864e700485ec656b4b00ffe) and probably other users are already using _custom_group_type_SHOW_ actions.
Do you think something like https://github.com/ckan/ckanext-hierarchy/pull/63 will help?

amercader:
@avdata99 see https://github.com/ckan/ckanext-hierarchy/pull/63#issuecomment-1562762112 for the best approach

avdata99:
Closed in favor of #1046

